PR #22, opened 2015-01-03
Not init warnings

aaustin:
@qinweigong 

Coffee Meets Bagel noticed that the URL callback was never made if they forgot to init Branch. We should return a helpful error message in this case.


aaustin:
@qinweigong can you check this one as well when you get a chance

qinweigong:
:+1:

PR #34, opened 2015-01-21
Unchecked iterator type

SleeplessByte:
You probably want to use the internal names() mapping provided by `JSONObject`. 
If you use a `StringBuilder`, probably want to append the strings to the buffer as opposed to creating 5 strings in the concatenation process and then adding that to the buffer.

I didn't actually test it.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/BranchRemoteInterface.java
+++ b/Branch-SDK/src/io/branch/referral/BranchRemoteInterface.java
@@ -277,12 +277,12 @@ public class BranchRemoteInterface extends RemoteInterface {
 		
 		if (json != null) {
             JSONArray names = json.names();
-	        if ( names != null ) {
+	        if (names != null) {
                 boolean first = true;
                 int size = names.length();
-                for( int i = 0; i < size; i++ ) {
+                for(int i = 0; i < size; i++) {
 
-                    String key = names.getString( i );
+                    String key = names.getString(i);
 
 	        		if (first) {
 		        		result.append("?");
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/BranchRemoteInterface.java
```diff
@@ -270,32 +272,35 @@ public void sendLog(String log) {
...
+	private String convertJSONtoString(JSONObject json) throws JSONException {
 		StringBuilder result = new StringBuilder();
 		
 		if (json != null) {
-	        Iterator<String> iter = json.keys();
-	        if (iter.hasNext()) {
-	        	boolean first = true;
-	        	do {
+            JSONArray names = json.names();
+	        if ( names != null ) {
```
aaustin: Can you stick with the Branch style convention we have and not insert spaces before and after the ( internals )?

SleeplessByte: Yeah, no problem!

SleeplessByte: 1424806

Branch-SDK/src/io/branch/referral/BranchRemoteInterface.java
```diff
@@ -270,23 +272,26 @@ public void sendLog(String log) {
 		}
 	}
 	
-	private String convertJSONtoString(JSONObject json) {
+	private String convertJSONtoString(JSONObject json) throws JSONException {
```
qinweigong: We are still catching JSONException down there, do we still need to throw it here?

SleeplessByte: Myself I really don't like to return null on an exception; surely you could return null instead.

Branch-SDK/src/io/branch/referral/BranchRemoteInterface.java
```diff
@@ -270,23 +272,26 @@ public void sendLog(String log) {
...
 		        		result.append("&");
 		        	}
-	        		
-	        		String key = iter.next();
-	        		String value;
+
+                    String value;
 	        		try {
 						value = json.getString(key);
 					} catch (JSONException e) {
```
qinweigong: With the method signature having "throws JSONException", we can remove the try/catch here?

PR #39, opened 2015-01-31
switch for smart session; bnc_no_value prevention

qinweigong:
@aaustin


aaustin:
Looks great - have you tested calling getShortUrl or something before init was called? If you've played around with it, :+1:

PR #43, opened 2015-02-06
Applisting!

qinweigong:
@aaustin @SleeplessByte


Branch-SDK/src/io/branch/referral/SystemObserver.java
```diff
@@ -51,26 +57,86 @@ public boolean hasRealHardwareId() {
...
-	        	  
-	          }
-	    } catch (NameNotFoundException ignored ) {
+	            scheme = new ApkParser().decompressXML(xml);
+	        } catch (Exception ignored) {
+	        } finally {
+	        	xml = null;
+	        	try {
+	        		jf.close();
+	        		is.close();
```
aaustin: do we want to set jf and is to null as well to ensure there aren't any dangling pointers?

qinweigong: Yes, good points.

qinweigong:
Problem:
The memory issue is not about the ApkParser, but mainly caused by the instantiation of a new inputstream for the manifest file. No matter how we parse the file, we can't walk around using the inputstream.

Solution:
1 - Instead of checking/running processListOfApps on install/open after certain interval, just schedule to run the process at mid-night every Sunday (or a different date/time?). This scheduling is done after install.

2 - If memory usage is still a concern, we can apply a limit on manifest file size, i.e. if the file is larger than 20KB (I've seen some as big as 120-130KB), then skip it.

3 - We provide the public method to disable app listing at all.

aaustin:
Seems like a great idea - scheduler ready for review and test?

qinweigong:
Scheduled to run at 2am every Saturday.

SleeplessByte:
You probably want to switch using the `AlarmManager` + `BroadcastReceiver` so the task holds a CPU wake lock and is also executed if the app is not running.

qinweigong:
@SleeplessByte Thanks for the suggestion. Will try it out. The memory footprint on some test phones could reach 26MB, but what's more annoying is it keeps sending out messages about memory usage expansion, which may well be a concern for our partners when they see them.

SleeplessByte:
Android 4.4+ supports ActivityManager.isLowRamDevice(). That might solve
part of the equation.

qinweigong:
@SleeplessByte Did some research on AlarmManager & BroadcastReceiver today, and actually got it setup and working. 
However, seems I have to add a <receiver> to AndroidManifest, which is not desired (we don't want our partners to add too much stuff to their manifest).
I did find there's a way to add it programmatically by context.registerReceiver(). But seems this way it depends on application life time?

SleeplessByte:
Manifests should be merged. So you could add it to the library's manifest and the partner doesn't need to so anything.

qinweigong:
Hmm, I added the receiver to Branch-SDK/AndroidManifest.xml like below:

<application android:allowBackup="true">
        <receiver android:name="io.branch.referral.BranchBroadcastReceiver" >
        </receiver>
    </application>

But it doesn't work.

Adding  <receiver android:name="io.branch.referral.BranchBroadcastReceiver" /> to  Branch-SDK-TestBed/AndroidManifest.xml works fine!

Or, do I need to do anything to manually merge the two manifests?

SleeplessByte:
It should work. I thought [this](https://www.youtube.com/watch?feature=player_embedded&v=Erd2k6EKxCQ#t=2939s) was legacy, but watch it nonetheless.

qinweigong:
Yah, I found it too. But when playing the "manifestmerger.enabled=true" trick, seems it only works when added to partner project's project.properties, but not our only project.properties file! Then it's the same thing - still need partners to do extra work. :-(

SleeplessByte:
Myeah, it will be part of the default options in the future. But not right
now. I suppose for the time being you could go with
http://developer.android.com/reference/java/util/concurrent/ScheduledExecutorService.html
and update it when the manifest merger is default. The 'problem' with the
non-manifest options is that they don't run if the app isn't loaded in
background. On these low end devices, the app get's killed --

aaustin:
Nice work here Qinwei! Looks awesome. Let's get it out there

aaustin:
:+1:

PR #44, opened 2015-02-09
added getAdversitingId

qinweigong:
@aaustin  Tested


aaustin:
Whoa nice! Works in both instances?

qinweigong:
Yes. I added Google Play services SDK (the package required for advertising id) to the Branchster project, ran it on your Samsung, got the advertising id.
When running the TestBed which doesn't have the google play services SDK, it just returns null.

aaustin:
:+1:

aaustin:
Just looked at the backend PR @qinweigong, which looks for key 'google_advertising_id'. This PR submits 'advertising_id'

aaustin:
seems like there's a disconnect here?

qinweigong:
This PR wasn't coupled with the backend PR. Please see today's Android PR https://github.com/BranchMetrics/Branch-Android-SDK/pull/45, which is in line with the backend PR, both using "google_advertising_id".

aaustin:
I guess I don't understand why we needed anything else done in Android after this PR

qinweigong:
The change in this PR was about retrieving the google advertising id (if Google Play service SDK is imported to the project) by ourselves.
Today's new PR is about accepting the google advertising id the partners retrieve by themselves and pass it to us in initSession call. This is per what you told me in Slack yesterday. Am I having misunderstanding here?

qinweigong:
Maybe I should've just add today's change to this PR instead of creating another PR.

aaustin:
Ahhh - no - this was miscommunication. Sorry! Partners would never pass us Google Advertising ID via init since we can just read it directly from the Play Services framework. I think we've got it all covered in the last PR - can you just make sure the keys match server/SDK?

qinweigong:
Oh, sure!

PR #45, opened 2015-02-11
Pass advertising id to initSession

qinweigong:
@aaustin


qinweigong:
Backend PR https://github.com/BranchMetrics/branch-backend/pull/631

PR #51, opened 2015-03-02
Lock on queue itself!

qinweigong:
@aaustin


aaustin:
didn't we add this lock before, for a reason?

qinweigong:
We had a lock from the beginning, and it was a lock on the queue itself.
However, it got changed to a lock on an independent lock object by DerkJan later.
We are using a Collections.synchronizedList as data structure. As its documentation says, no manual sync is required for enqueue/dequeue. However, when traversing this list, it still needs synchronization. If we are locking on a separate object during traversing, it has no sync effect on enqueue/dequeue on the other threads. That's why I am changing the lock back.

aaustin:
Why did DerkJan change it though? I thought he had a good reason

qinweigong:
@SleeplessByte 
Sorry that it's been a while back and you may not be able to recall, but just want to get your clarification here.
Seems you changed the lock in ServerRequestQueue in
https://github.com/BranchMetrics/Branch-Android-SDK/commit/98d244bafb10338318f1ce305994efec21c0cb9b

From synchronized(queue) -> synchronized(lock)

Could you please clarify what the reason was? Would there be any concern/issue if I change it back?

SleeplessByte:
If you lock on the queue, which is not final, you lose the lock when you create a queue.

````Java
Object someObject = null;
lock( someObject ) //locks a null pointer

someObject = new Object();
lock( someObject ) //locks the first reference

someObject = new Object();
lock( someObject ) //locks the second reference
````

Any code in the three locks above would be able to run in parallel. You should not be locking on the queue itself, unless it's final (e.g. will only ever have one reference that's not null). Creating a lockobject (final) and locking on that will ensure only one thread accessing the code inside the block.

In the current code you could move back to locking the queue, but you'll have to make it final, and thus initialize it in the constructor.

qinweigong:
Ahh, I see. It should be fine, as the queue is instantiated in a singleton manner, i.e. there's always only one instance created. But guess it doesn't hurt to make it final. Thanks!

SleeplessByte:
@qinweigong If you want to lock on it, you should make it final as it will ensure no one can lose the lock later on (code changes?). I just changed it to an object lock because I did that everywhere in the codebase -- consistency.

qinweigong:
Makes sense! Will make it final. Thanks for your advice!

PR #52, opened 2015-03-04
Initial Tests

qinweigong:


qinweigong:
@aaustin

qinweigong:
CircleCI test success is false. Currently all tests will pass their Android tests, and they are working on it.

<quote>
Hi,
We're currently working on revamping our android support to make it more robust and one thing we've noticed is that adb shell doesn't exit properly.
You might want to install fb-adb in your container and use that. It will be included in an upcoming container release on our side, but you can custom install it for now.
Hope that helps,
Rob.
</quote>
Their ETA is 3/12. Will add CircleCI integration to asana todo list.

aaustin:
This is awesome!! :+1: Super excited. Can you document the outstanding tests you think we should run as a todo for the eventual Android hire?

qinweigong:
I already added the todo list to Android Asana tasks.

PR #55, opened 2015-03-17
Not sure how so many closes are added..

aaustin:
@qinweigong Thoughts?
https://gist.github.com/samshahidi/5a9f3741872de0c30a9c
https://gist.github.com/samshahidi/286e38e548d9d53dceb9


qinweigong:
:+1: Examining our code for the recursive call close, but this fix looks great!

PR #58, opened 2015-04-10
Android retry

Sarkar:
Refactor retry to network requests instead of at the Branch level. Similar logic to @shortstuffsushi PR. 

@qinweigong - since you have a big PR request coming in, let's figure out the best strategy to avoid conflicts yeah?

### DIFF
--- a/Branch-SDK/src/io/branch/referral/BranchConfig.java
+++ b/Branch-SDK/src/io/branch/referral/BranchConfig.java
@@ -1,0 +1,8 @@
+package io.branch.referral;
+
+/*
+ *
+ */
+public class BranchConfig {
+    public static final String TAG = "BranchSDK";
+}
--- a/Branch-SDK/src/io/branch/referral/RemoteInterface.java
+++ b/Branch-SDK/src/io/branch/referral/RemoteInterface.java
@@ -7,6 +7,7 @@ import java.io.InputStreamReader;
 import java.net.SocketException;
 import java.net.UnknownHostException;
 
+import android.content.Context;
 import android.os.NetworkOnMainThreadException;
 import android.util.Log;
 
@@ -33,7 +34,13 @@ public class RemoteInterface {
     private static final String SDK_VERSION = "1.4.4";
     private static final int DEFAULT_TIMEOUT = 3000;
 
+    protected PrefHelper prefHelper_;
+
     public RemoteInterface() { }
+
+    public RemoteInterface(Context context) {
+        prefHelper_ = PrefHelper.getInstance(context);
+    }
 
     private HttpClient getGenericHttpClient(int timeout) {
         if (timeout <= 0) {
@@ -88,7 +95,7 @@ public class RemoteInterface {
         return make_restful_get(url, tag, timeout, 0, true);
     }
 
-    public ServerResponse make_restful_get(String url, String tag, int timeout, int retryCount, boolean log) {
+    public ServerResponse make_restful_get(String url, String tag, int timeout, int retryNumber, boolean log) {
         try {
             if (url.indexOf('?') == -1) {
                 url += "?sdk=android" + SDK_VERSION;
@@ -100,7 +107,18 @@ public class RemoteInterface {
             }
             HttpGet request = new HttpGet(url);
             HttpResponse response = getGenericHttpClient(timeout).execute(request);
-            return processEntityForJSON(response.getEntity(), response.getStatusLine().getStatusCode(), tag, log, null);
+
+            int statusCode = response.getStatusLine().getStatusCode();
+            boolean isRetryableStatusCode = statusCode >= 500;
+
+            // Retry request before processing
+            if ( (retryNumber < prefHelper_.getRetryCount()) && isRetryableStatusCode) {
+                retryNumber +=1;
+                make_restful_get(url, tag, timeout, retryNumber, log);
+            } else {
+                Log.d(BranchConfig.TAG, "k");
+                return processEntityForJSON(response.getEntity(), response.getStatusLine().getStatusCode(), tag, log, null);
+            }
 
         } catch (ClientProtocolException ex) {
             if (log) {
--- a/Branch-SDK/src/io/branch/referral/RemoteInterface.java
+++ b/Branch-SDK/src/io/branch/referral/RemoteInterface.java
@@ -26,128 +26,156 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 public class RemoteInterface {
-	public static final String NO_TAG_VALUE = "no_tag";
-	public static final int NO_CONNECTIVITY_STATUS = -1009;
-	public static final int NO_API_KEY_STATUS = -1234;
+    public static final String NO_TAG_VALUE = "no_tag";
+    public static final int NO_CONNECTIVITY_STATUS = -1009;
+    public static final int NO_API_KEY_STATUS = -1234;
 
-	private static final String SDK_VERSION = "1.4.4";
-	private static final int DEFAULT_TIMEOUT = 3000;
-	
-	public RemoteInterface() { }
-	
-	private HttpClient getGenericHttpClient(int timeout) {
-		if (timeout <= 0)
-			timeout = DEFAULT_TIMEOUT;
-		HttpParams httpParams = new BasicHttpParams();
-		HttpConnectionParams.setConnectionTimeout(httpParams, timeout);
-		HttpConnectionParams.setSoTimeout(httpParams, timeout);
-		return new DefaultHttpClient(httpParams);
-	}
-	
-	private ServerResponse processEntityForJSON (HttpEntity entity, int statusCode, String tag, boolean log, BranchLinkData linkData) {
-		ServerResponse result = new ServerResponse(tag, statusCode, linkData);
-		try {
-			if (entity != null) {
-		    	InputStream instream = entity.getContent();
-		    	BufferedReader rd = new BufferedReader(new InputStreamReader(instream));
-		    	
-		    	String line = rd.readLine();
-		    	if (log) PrefHelper.Debug("BranchSDK", "returned " + line);
+    private static final String SDK_VERSION = "1.4.4";
+    private static final int DEFAULT_TIMEOUT = 3000;
 
-		    	if (line != null) {
-		    		try {
-		    			JSONObject jsonObj = new JSONObject(line);
-		    			result.setPost(jsonObj);
-		    		} catch (JSONException ex) {
-		    			try {
-		    				JSONArray jsonArray = new JSONArray(line);
-		    				result.setPost(jsonArray);
-		    			} catch (JSONException ex2) {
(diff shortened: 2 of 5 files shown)


Branch-SDK/src/io/branch/referral/BranchConfig.java
```diff
@@ -0,0 +1,8 @@
+package io.branch.referral;
+
+/*
+ *
+ */
+public class BranchConfig {
```
qinweigong: Do you have plan to move more stuff into this class? If only for this TAG, maybe we can just put it into PrefHelper.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -26,128 +27,183 @@
...
-		}
-	}
+    public static final String NO_TAG_VALUE = "no_tag";
+    public static final int NO_CONNECTIVITY_STATUS = -1009;
+    public static final int NO_API_KEY_STATUS = -1234;
+
+    private static final String SDK_VERSION = "1.4.4";
+    private static final int DEFAULT_TIMEOUT = 3000;
+
+    protected PrefHelper prefHelper_;
```
qinweigong: Besides moving prefHelper_ here, can I assume all the other lines are for tab -> spaces changes? It's hard to compare so many lines one by one :-)

shortstuffsushi:
Adding to @qinweigong's statement, it would be good in the future to separate formatting from pull request content, or at least make a separate commit. Definitely makes diffing harder.

Sarkar:
Sorry about that guys. I should've done that. I can say that the biggest
diff is inside the base GET and POST functions adding logic to retry while
having a retry count. Essentially what Graham wrote for iOS  perhaps
looking at individual commits will help

On Thursday, April 9, 2015, Graham Mueller <notifications@github.com> wrote:

> Adding to @qinweigong <https://github.com/qinweigong>'s statement, it
> would be good in the future to separate formatting from pull request
> content, or at least make a separate commit. Definitely makes diffing
> harder.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/BranchMetrics/Branch-Android-SDK/pull/58#issuecomment-91418246>
> .
>


-- 

*Sahil Verma*

*Branch Metrics *

w branch.io | c 510-386-1750 | s sahilroyverma

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -26,128 +27,183 @@
...
+            return new ServerResponse(tag, NO_CONNECTIVITY_STATUS);
+        } catch (Exception ex) {
+            if (log) {
+                PrefHelper.Debug(getClass().getSimpleName(), "Exception: " + ex.getMessage());
+            }
+            if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.HONEYCOMB) {
+                if (ex instanceof NetworkOnMainThreadException)
+                    Log.e(BranchConfig.TAG, "Branch Error: Don't call our synchronous methods on the main thread!");
+            }
+            return new ServerResponse(tag, 500);
```
shortstuffsushi: 500 doesn't seem like the right code here -- indicates server failure.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -26,128 +27,183 @@
...
+            request.setHeader("Content-type", "application/json");
+            HttpResponse response = getGenericHttpClient(timeout).execute(request);
+            int statusCode = response.getStatusLine().getStatusCode();
+            boolean isRetryableStatusCode = statusCode >= 500;
+
+            if ((retryNumber < prefHelper_.getRetryCount()) && isRetryableStatusCode) {
+                retryNumber +=1;
+                make_restful_post(body, url, tag, timeout, retryNumber, log, linkData);
+            } else {
+                return processEntityForJSON(response.getEntity(), response.getStatusLine().getStatusCode(), tag, log, null);
```
shortstuffsushi: Seems this else and the next line could be combined.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -26,128 +27,183 @@
...
+            }
+
+            body.put("sdk", "android" + SDK_VERSION);
+            if (log) {
+                PrefHelper.Debug(BranchConfig.TAG, "posting to " + url);
+                PrefHelper.Debug(BranchConfig.TAG, "Post value = " + body.toString(4));
+            }
+            HttpPost request = new HttpPost(url);
+            request.setEntity(new ByteArrayEntity(body.toString().getBytes("UTF8")));
+            request.setHeader("Content-type", "application/json");
```
shortstuffsushi: Content-Type should be PascalCased, though according to the RFC, the server should accept it either way.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -26,128 +27,183 @@
...
+    }
+
+    public ServerResponse make_restful_post(JSONObject body, String url, String tag, int timeout, boolean log) {
+        return make_restful_post(body, url, tag, timeout, 0, log, null);
+    }
+
+    public ServerResponse make_restful_post(JSONObject body, String url, String tag, int timeout,
+                                            int retryNumber, boolean log, BranchLinkData linkData) {
+        try {
+            if (body.has("app_id") && body.getString("app_id").equals(PrefHelper.NO_STRING_VALUE)) {
```
shortstuffsushi: On iOS, I'm adding the app_id myself in this method, rather than having it passed into this method. That way, rather than having to remember this for every request, things that we *always* want are added to each request. I also add things like sdk=iOS0.x.x, etc.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -26,128 +27,183 @@
...
+        } catch (IOException ex) {
+            if (log) {
+                PrefHelper.Debug(getClass().getSimpleName(), "IO exception: "
+                        + ex.getMessage());
+            }
+        }
+        return result;
+    }
+
+    //region GET Requests
```
shortstuffsushi: Unrelated to anything, but does IntelliJ respect regions?

Branch-SDK/src/io/branch/referral/RemoteInterface.java
qinweigong: @shortstuffsushi I am make that change in my PR, FYI.

Branch-SDK/src/io/branch/referral/BranchConfig.java
shortstuffsushi: It seems like this would be an appropriate place for things like the release number -- it's used for that in iOS. It also sets the base server url depending on whether you're in debug.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
Sarkar: This is all super valuable feedback but it actually doesn't pertain to what
I changed. Let's sync up about this tomorrow? So I can go over with you two
what I changed then we finalize?

On Thursday, April 9, 2015, qinweigong <notifications@github.com> wrote:

> In Branch-SDK/src/io/branch/referral/RemoteInterface.java
> <https://github.com/BranchMetrics/Branch-Android-SDK/pull/58#discussion_r28121046>
> :
>
> > +    public ServerResponse make_restful_post(JSONObject body, String url, String tag, int timeout) {
> > +        return make_restful_post(body, url, tag, timeout, 0, true, null);
> > +    }
> > +
> > +    public ServerResponse make_restful_post(JSONObject body, String url, String tag, int timeout, BranchLinkData linkData) {
> > +        return make_restful_post(body, url, tag, timeout, 0, true, linkData);
> > +    }
> > +
> > +    public ServerResponse make_restful_post(JSONObject body, String url, String tag, int timeout, boolean log) {
> > +        return make_restful_post(body, url, tag, timeout, 0, log, null);
> > +    }
> > +
> > +    public ServerResponse make_restful_post(JSONObject body, String url, String tag, int timeout,
> > +                                            int retryNumber, boolean log, BranchLinkData linkData) {
> > +        try {
> > +            if (body.has("app_id") && body.getString("app_id").equals(PrefHelper.NO_STRING_VALUE)) {
>
> @shortstuffsushi <https://github.com/shortstuffsushi> I am make that
> change in my PR, FYI.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/BranchMetrics/Branch-Android-SDK/pull/58/files#r28121046>
> .
>


-- 

*Sahil Verma*

*Branch Metrics *

w branch.io | c 510-386-1750 | s sahilroyverma

shortstuffsushi: Ah, see that now -- lost that cuz of the big block diff :wink: Sounds good though.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -26,128 +27,183 @@
...
+    public ServerResponse make_restful_get(String url, String tag, int timeout) {
+        return make_restful_get(url, tag, timeout, 0, true);
+    }
+
+    public ServerResponse make_restful_get(String url, String tag, int timeout, int retryNumber, boolean log) {
```
shortstuffsushi: My only other feedback about this change is the fact that the method w/ retry number is exposed here. On iOS, the actual retry method is private -- not sure if we care to do the same here.

Branch-SDK/src/io/branch/referral/BranchConfig.java
qinweigong: My bad. I created BNCConfig on iOS side, but too lazy to do the same on Android side at that time.
Anyway, I'd prefer moving all the stuff to here in one shot. People getting this release will find this class with so little content. That doesn't look good :-)

Sarkar: @shortstuffsushi this is exactly what I intend for thsi class. I'd rather move final statics like the end point, URL, version number, etc all in one place so it's never manually specified else where. 

@qinweigong in that case that will be a separate PR then.

Branch-SDK/src/io/branch/referral/RemoteInterface.java
Sarkar: yep! + the retry count

Branch-SDK/src/io/branch/referral/RemoteInterface.java
Sarkar: This is a good catch. I'll add this

Branch-SDK/src/io/branch/referral/RemoteInterface.java
Sarkar: heck yes! code folding all the way

Branch-SDK/src/io/branch/referral/RemoteInterface.java
Sarkar: i thought about this too. the base method should be private for both GETs and POSTs. I'll make this change as well.

PR #66, opened 2015-04-26
specify branch key in manifest directly

qinweigong:
@aaustin 
See detailed explanation in https://github.com/BranchMetrics/Branch-Android-SDK/issues/65


aaustin:
:+1:

PR #71, opened 2015-05-19
add new size

aaustin:
@qinweigong some people still reporting stackoverflow errors where the queue is processing > 25 items.


qinweigong:
:+1:

PR #73, opened 2015-05-26
change uri scheme

aaustin:
@sojanpr can you review

this is to update the URI scheme


Branch-SDK-TestBed/AndroidManifest.xml
```diff
@@ -24,7 +24,7 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
             <intent-filter>
-                <data android:scheme="branchtest" android:host="open" />
+                <data android:scheme="branchtestversion2" android:host="open" />
```
aaustin: why are we changing this?

PR #75, opened 2015-06-03
New Session Management Changes

sojanpr:
@qinweigong @aaustin 

Changes to do Session management with out overriding activity life
cycle methods.This implementation requires applications developers to
extend application class by BranchLinkApplication.
This implementation requires minimum API level 14+

BranchTestBed Application is also modified to test the session
changes.Added a switch in main activity(SESSION_MANAGEMENT_MODE) to
switch between legacy session management and new session management.


shortstuffsushi:
@sojanpr You should add the `.metadata` folder to gitignore and not check all those files in.

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
```diff
@@ -24,6 +25,17 @@
 
 public class MainActivity extends Activity {
 	Branch branch;
+
+    public enum SESSION_MANAGEMENT_MODE {
```
qinweigong: Shall we move this into the SDK instead?

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
```diff
@@ -24,6 +25,17 @@
...
+        AUTO     /*Branch SDK Manages the session for you.For this mode minimum API level should
+                   be 14 or above.Make sure to instantiate BranchLinkApp class to use this mode.*/
+        ,MANUAL  /* You are responsible for managing the session.Need to initialise the session on
```
qinweigong: For style sake, could you move this leading comma to after `AUTO` instead?

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
qinweigong: In our iOS SDK, we have a separate BNCConfig.h file which hosts come config settings. Due to my laziness, I didn't do the same thing on the Android side. Now I think it makes sense to add it there too, and it'd be a perfect place to host this session management mode thing. We don't use BNC file prefix on the android side, so could name it BranchConfig.java.

sojanpr: Added this just for switching between Session Management mode while we are testing.SDK already know about the mode by checking whether application is an instance of extending BranchLinkApp class.The following code in SDK is checking for it .
''' public static void createInstance(Application application) {
		isBranchAppInstantiated = application instanceof  BranchLinkApp; '''

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
sojanpr: Sure.Changed.

qinweigong:
Just a thought: what if a partner/developer happens to need to inherit their application from another super class?

sojanpr:
All Android Application class must be  inherited form "android.os.Application" class.If developer want to Inherit some other class,that should be a subclass of "android.os.Application" class.

Branch-SDK/src/io/branch/referral/BranchException.java
```diff
@@ -0,0 +1,49 @@
+package io.branch.referral;
+
+/**
+ * <p>
+ * Class for representing any Branch SDK exceptions
+ * Exception details are provided in exception message.
+ * </p>
+ */
+public class BranchException extends Exception {
```
qinweigong: Maybe we should consolidate this with BranchError.java

Branch-SDK/src/io/branch/referral/BranchLinkApp.java
```diff
@@ -0,0 +1,29 @@
...
+ * <p>
+ * Add this entry to the manifest if you don't have an Application class :
+ * </p>
+ * <p/>
+ * <pre style="background:#fff;padding:10px;border:2px solid silver;">
+ * <application
+ * -----
+ * android:name="io.branch.referal.BranchLinkApp">
+ */
+public class BranchLinkApp extends Application {
```
qinweigong: How about just call it BranchApp?

Branch-SDK/src/io/branch/referral/BranchException.java
sojanpr: You mean , moving the message string to BranchError.Java or Remove the exceptions completely

qinweigong: We should keep either BranchException.java or BranchError.java, but not both I think.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -334,6 +347,97 @@ private Branch(Context context) {
...
+	 *                 Adding your app key to your project</a>
+	 *
+	 * @param application  Your {@link Application} instance.
+	 *
+	 */
+
+	public static void createInstance(Application application) {
+		isBranchAppInstantiated = application instanceof  BranchLinkApp;
+      /* Create and initialise a new instance of Branch. */
+		Branch createdInstance = getBranchInstance(application,true);
```
qinweigong: So, in android Application and Context are the same thing? Current getBranchInstance accepts Context as its first param, and I don't see it change here.

sojanpr: AndroidApplication Class is already extending ContextWrapper.So we can use it as the Context

qinweigong: Neat!

Branch-SDK/src/io/branch/referral/BranchLinkApp.java
sojanpr: Cool.I like the name BranchApp

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -334,6 +347,97 @@ private Branch(Context context) {
...
+
+			//Check if Activity life cycle callbacks are set
+		else {
+			if (branchReferral_.isActivityObserverInitialised() == false) {
+				throw BranchException.getAPILevelException();
+			}
+
+			else {
+				//Set branch Key with  the Live key
+				String branchKey = branchReferral_.prefHelper_.getBranchKey(true);
```
qinweigong: this local variable is not used?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -871,7 +990,7 @@ public boolean onTouch(View v, MotionEvent ev) {
 	 */
 	public void closeSession() {
 		if (prefHelper_.getSmartSession()) {
-			if (keepAlive_) {
+			if (keepAlive_ && !isBranchAppInstantiated) { //No need to check for keepAlive_ for auto session management
```
qinweigong: I'm confused. Should this be `if (isBranchAppInstantiated || keepAlive)` ?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -334,6 +347,97 @@ private Branch(Context context) {
...
+	 *
+	 * @see          <a href="https://github.com/BranchMetrics/Branch-Android-SDK/blob/05e234855f983ae022633eb01989adb05775532e/README.md#add-your-app-key-to-your-project">
+	 *                 Adding your app key to your project</a>
+	 *
+	 * @param application  Your {@link Application} instance.
+	 *
+	 */
+
+	public static void createInstance(Application application) {
+		isBranchAppInstantiated = application instanceof  BranchLinkApp;
```
qinweigong: The variable name isn't reflecting what it means, maybe something like `isBranchApp`?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3026,6 +3145,82 @@ public void run() {
...
+			activityLifeCycleObserver_ = this;
+		}
+
+		@Override
+		public void onActivityCreated(Activity activity, Bundle bundle) {}
+
+		@Override
+		public void onActivityStarted(Activity activity) {
+			if(m_ActivityCnt < 1){ //Check if this is the first Activity.If so start a session
+				onSessionStarted();
```
qinweigong: onSessionStarted isn't called anywhere else, can we just call `initSession()` here instead and remove `private void onSessionStarted()` (same for onSessionEnded())?

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
qinweigong: Ahh, got it. So it'll be removed when we release, right?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Correct. Will remove the variable.
getKey() is used just to set the branch keys. prefHelper_.getBranchKey() method is actually setting the key internally. Is it a good idea to create a new method "setKey(isLive)" for setting the keys?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: closeSession() should not return if "isBranchAppInstantiated = true" .isBranchAppInstantiated is true when developer uses the new session management by extending BranchLinkApp class

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -334,6 +347,97 @@ private Branch(Context context) {
...
+
+		try {
+         /*Set an observer for ativity life cycle events*/
+			createdInstance.setActivityLifeCycleObserver(application);
+
+		}catch (NoSuchMethodError Ex){
+			//LifeCycleEvents are  available only from API level 14.
+			Log.w(TAG,BranchException.BRANCH_API_LVL_ERR_MSG);
+		}
+	}
```
qinweigong: Can we merge this createInstance method into the existing `getBranchInstance()` method? something like
```
private static Branch getBranchInstance(Application application, boolean isLive) {
		if (branchReferral_ == null) {
			branchReferral_ = Branch.initInstance(application);

			String branchKey = branchReferral_.prefHelper_.getBranchKey(isLive);
	        if (branchKey == null || branchKey.equalsIgnoreCase(PrefHelper.NO_STRING_VALUE)) {
	        	Log.i("BranchSDK", "Branch Warning: Please enter your branch_key in your project's res/values/strings.xml!");
	        }
		}
		branchReferral_.context_ = application;
		
		isBranchAppInstantiated = application instanceof BranchLinkApp;
		
		try {
			/*Set an observer for ativity life cycle events*/
			branchReferral_.setActivityLifeCycleObserver(application);
			
		}catch (NoSuchMethodError Ex){
			//LifeCycleEvents are  available only from API level 14.
			Log.w(TAG,BranchException.BRANCH_API_LVL_ERR_MSG);
		}
		
		return branchReferral_;
	}
```

Branch-SDK/src/io/branch/referral/Branch.java
qinweigong: If the new session management is in use, closeSession should have no effect, right? So we should return here immediately, otherwise, it will go on and call `executeClose()` down there, no?

Branch-SDK/src/io/branch/referral/Branch.java
qinweigong: We do have `public void setBranchKey` in PrefHelper.java

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: New Session Management is also using closeSession().Only difference is that close session is called only when the application goes to the background not with each activity stop.
So executeClose() is still called when the last  activity leaves the screen

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: how about the name "isAutoSessionMode" instead of "isBranchAppInstantiated"

qinweigong: Sounds much better :-)

Branch-SDK/src/io/branch/referral/Branch.java
qinweigong: As we talked, we'd probably better support the scenario where developer adopts the new session management, but forget to remove those closeSession calls in (some) activities. It'll be a nightmare for our support.

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
sojanpr: yes

Branch-SDK/src/io/branch/referral/BranchLinkApp.java
sojanpr: Done

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Done

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Done

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Done.

Branch-SDK/src/io/branch/referral/BranchException.java
sojanpr: @qinweigong  It seems like BranchError.Java specified String are used only as parameters in callback methods and seems to be an ideal design.
In the case of getInstance() we can't return error messages and also we may have different type of errors.So Exception forces Developers to handle it in their own  way.What is your suggestion?

qinweigong: Mmm, after a second thought, I agree that it makes sense to have both error and exception classes.

shortstuffsushi:
@sojanpr poke to remove `.metadata` :wink:

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -333,6 +346,33 @@ private Branch(Context context) {
...
+	 *
+	 * @return An initialised singleton {@link Branch} object
+	 *
+	 * @throws BranchException Exception</br>
+	 *          1)If your {@link Application}  is not instance of {@link BranchApp} </br>
+	 *          2)If the minimum API level is below 14
+	 */
+	public static Branch getInstance() throws BranchException {
+		// Check if BranchApp is instantiated.
+		if(branchReferral_ == null || isAutoSessionMode == false ) {
```
aaustin: can you add a space between *if* and *(*? Just to stay consistent

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -301,6 +307,13 @@
 
 	private ScheduledFuture<?> appListingSchedule_;
 
+	/* BranchActivityLifeCycleObserver instance.Should be initialised on creating Instance with Application object. */
```
aaustin: Can you fix the spacing around the period here and everywhere else as we talked about?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -333,6 +346,33 @@ private Branch(Context context) {
 		linkCache_ = new HashMap<BranchLinkData, String>();
 	}
 
+
+	/**
+	 * <p>Singleton method to return the pre-initialised object of the type {@link Branch}.
+	 * Make sure your app is  instantiating {@link BranchApp} before calling this method</p>
```
aaustin: Sorry to be a stickler, but can you clean up the double spacing here and everywhere else? There's 2 spaces between *is* and *instantiating*

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -333,6 +346,33 @@ private Branch(Context context) {
...
+	 */
+	public static Branch getInstance() throws BranchException {
+		// Check if BranchApp is instantiated.
+		if(branchReferral_ == null || isAutoSessionMode == false ) {
+			throw BranchException.getInstatiationException();
+		}
+
+			// Check if Activity life cycle callbacks are set.
+		else if (branchReferral_.isActivityObserverInitialised() == false) {
+				throw BranchException.getAPILevelException();
```
aaustin: not sure if it's Github or not, but there seems to be an extra tab here?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -333,6 +346,33 @@ private Branch(Context context) {
...
+	 * <p>Singleton method to return the pre-initialised object of the type {@link Branch}.
+	 * Make sure your app is  instantiating {@link BranchApp} before calling this method</p>
+	 *
+	 * @return An initialised singleton {@link Branch} object
+	 *
+	 * @throws BranchException Exception</br>
+	 *          1)If your {@link Application}  is not instance of {@link BranchApp} </br>
+	 *          2)If the minimum API level is below 14
+	 */
+	public static Branch getInstance() throws BranchException {
```
aaustin: I'm not a big fan of having to make the developer wrap all our method calls in try/catch personally. It always annoyed me that I had to do it for JSON. Can we just silently fail and throw an error log? Thoughts?

sojanpr: So in case of an error inside  getInstace() what should we return ,a Null?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3026,6 +3095,68 @@ public void run() {
...
+		application.registerActivityLifecycleCallbacks(new BranchActivityLifeCycleObserver());
+	}
+
+	/**
+	 * <p>Class that observes activity life cycle events and determines when to start and stop
+	 * session.</p>
+	 */
+	@TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+	private class BranchActivityLifeCycleObserver implements Application.ActivityLifecycleCallbacks{
+		int m_ActivityCnt = 0; //Keep the count of live  activities
```
aaustin: Can we stick with the private variable naming convention that we use elsewhere. Eg:
private boolean initNotStarted_;

Branch-SDK/src/io/branch/referral/Branch.java
qinweigong: and remove the space between false and )

Branch-SDK/src/io/branch/referral/BranchApp.java
```diff
@@ -0,0 +1,65 @@
...
+ *      <meta-data android:name="io.branch.sdk.BranchKey" android:value="Your_Branch_Live_Key" />
+ *      <meta-data android:name="io.branch.sdk.BranchKey.test" android:value="Your_Branch_Test_Key" />
+ */
+public class BranchApp extends Application {
+
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        if(isTestModeEnabled() ==false) {
```
aaustin: some spacing fixes here please

Branch-SDK/src/io/branch/referral/SystemObserver.java
```diff
@@ -638,4 +638,32 @@ public String getAdvertisingId() {
...
+			Class<?> AdvertisingIdClientClass = Class.forName("com.google.android.gms.ads.identifier.AdvertisingIdClient");
+			Method getAdvertisingIdInfoMethod = AdvertisingIdClientClass.getMethod("getAdvertisingIdInfo", Context.class);
+			Object adInfoObj = getAdvertisingIdInfoMethod.invoke(null, context_);
+			Method getLatMethod = adInfoObj.getClass().getMethod("isLimitAdTrackingEnabled");
+
+			latVal = (Boolean) getLatMethod.invoke(adInfoObj) ? 1 :0;
+		} catch(IllegalStateException ex) {
+			ex.printStackTrace();
+		} catch(Exception ignore) {
+		}
```
aaustin: another stickler thing, sorry.. Can you put empty exception handling brackets on the same line?

Branch-SDK/src/io/branch/referral/BranchApp.java
```diff
@@ -0,0 +1,65 @@
...
+    private boolean isTestModeEnabled() {
+        boolean isTestMode = false;
+        String testModeKey = "io.branch.sdk.TestMode";
+        try {
+            final ApplicationInfo ai = getPackageManager().getApplicationInfo(getPackageName(), PackageManager.GET_META_DATA);
+            if (ai.metaData != null) {
+                isTestMode = ai.metaData.getBoolean(testModeKey, false);
+            }
+        } catch (final PackageManager.NameNotFoundException e) {
+        }
```
aaustin: align brackets as I commented below

Branch-SDK/src/io/branch/referral/Branch.java
aaustin: Hmmmmmmm - If we leave it as is, will Eclipse/Android studio enforce try/catch handling for all getInstance calls?

sojanpr: Yes all IDEs will enforce user to handle the exception.That way developers will be aware of errors /mistakes happening and they can handle it in their own way

aaustin:
In Eclipse, there's a warning in the SDK.
screen shot 2015-06-05 at 5 26 05 pm

aaustin:
Also:
1. Ran test bed app
2. clicked getShortUrl
screen shot 2015-06-05 at 5 27 59 pm

sojanpr:
@aaustin  I am unable to reproduce the crash.What version of Android is in your phone?

sojanpr:
Just simulated API < 14 and could reproduce the issue. SDK will provide Brranch Exception before this Happens.Currently we are ignoring it.Is it good to show a dialog for the error in this case.

'''io.branch.referral.BranchException: BranchApp class can be used only with API level 14 or above. Please make sure your minimum API level supported is 14. If you wish to use API level below 14 consider calling getInstance(Context) instead.
W/System.err(27076): 	at io.branch.branchandroiddemo.MainActivity.onStart(MainActivity.java:218)
W/System.err(27076): 	at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1239)
W/System.err(27076): 	at android.app.Activity.performStart(Activity.java:5322)'''

aaustin:
Ahh - I think it's 5.0. Let me try again. Maybe the project sync failed

sojanpr:
How about defaulting back to Legacy way on getting API level error in BranchTest app?

aaustin:
worked this time

sojanpr:
Added Handling for API level error with fall back to legacy session management

aaustin:
One other request: can you make it so that in the old method, keepAlive is only called for open or install? I think we're launching the keep alive for all calls and we only need it for open/close for smart session tracking across activities

sojanpr:
Done

aaustin:
Let's do it :+1:

PR #76, opened 2015-06-10
Android-SDK-Revamp

sojanpr:
Android SDK Revamping for the backlog items


aaustin:
:+1: can we roll all the small tweaks into one giant release?

aaustin:
or is that what this PR is for?

sojanpr:
Yes.This PR is for all individual items in the Android backlog.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1096,6 +1097,18 @@ public void run() {
...
 	/**
+	 * Indicates whether or not this user has a custom identity specified for them. Note that this is independent of installs.
+	 * If you call setIdentity, this device will have that identity associated with this user until logout is called.
+	 * This includes persisting through uninstalls, as we track device id.
+	 *
+	 * @return 				A {@link Boolean} value that will return <i>true</i> only if user already has an identity.
+	 *
+	 */
+	public boolean isUserIdentified() {
+		return !prefHelper_.getIdentity().equals(PrefHelper.NO_STRING_VALUE);
```
shortstuffsushi: This could be null, couldn't it?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3112,6 +3168,39 @@ public void run() {
...
+		JSONObject requestObject = resp.getRequestObject();
+		if (requestObject != null) {
+			if (requestObject.has("bucket") && requestObject.has("amount")) {
+				try {
+					int redeemedCredits = requestObject.getInt("amount");
+					String creditBucket = requestObject.getString("bucket");
+					updateListener = redeemedCredits > 0;
+
+					int updatedCreditCount = prefHelper_.getCreditCount(creditBucket) - redeemedCredits;
+					updatedCreditCount = updatedCreditCount > 0 ? updatedCreditCount : 0;
```
shortstuffsushi: Is it possible for this to be less than zero? Also, why not just like `Math.max(count, 0)`?

Branch-SDK/src/io/branch/referral/ServerResponse.java
```diff
@@ -135,4 +140,28 @@ public JSONArray getArray() {
...
+	}
+
+	/**
+	 * <p>Checks whether the request params cause this response is an instance of a
+	 * {@link JSONObject} object, and if so type-casts it to the corresponding class type.</p>
+	 *
+	 * @return A {@link JSONObject} containing the post params sent with corresponding  server request, or a null
+	 * value if the post params is not of the {@link JSONObject} type.
+	 */
+	public JSONObject getRequestObject() {
```
shortstuffsushi: It seems to me that this, and all other instances being named this way, should be named `getResponseObject`, not request.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3112,6 +3168,39 @@ public void run() {
...
+				}
+			}
+		}
+		final boolean finUpdateListener = updateListener;
+		Handler mainHandler = new Handler(context_.getMainLooper());
+		mainHandler.post(new Runnable() {
+			@Override
+			public void run() {
+				if (redeemStateChangedCallback_ != null) {
+					if (finUpdateListener)
```
shortstuffsushi: Avoid duplicating the actual callback here; use a local variable and assign as null for success.

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: This couldn't be null.Default return value for  `prefHelper_.getIdentity()` is `PrefHelper.NO_STRING_VALUE`

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: This cannot be less than 0 with current scenario.Added a proactive check there.
Math.max(cnt,0) seems more readable will add this

Branch-SDK/src/io/branch/referral/ServerResponse.java
sojanpr: It is actually the request send to the server in order get the `ServerResponse' object.
It seems like naming to `getResponseObject` may make the impression that this is the response data we got form the post. Does it make sense?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Not sure what you mean here by "duplicating Callback".Can you please explain a little bit more.

Branch-SDK/src/io/branch/referral/ServerResponse.java
shortstuffsushi: I'm not really sure I understand what you're saying here, but based on [the way you use it here](https://github.com/BranchMetrics/Branch-Android-SDK/pull/76/files#diff-11fc8f6026de99b60216e9aca37d6d26R3173) , it looks to me like you're treating it as a response body, right?

Or, are you adding the request object onto the response, and then inspecting fields on the request? If so, I'd really prefer we avoid that. I specifically removed that same behavior from the iOS SDK already -- we want to avoid making every class dependent on every other class; there should be a better way to keep track of the amount redeemed.

If this is all you can do for this pull request, that's ok, but I really want to avoid this in the future.

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: I mean, it is possible for the credit count to become less than zero *before* this check? You're adding a check here like you think it's possible for the credit count to become negative from the server... shouldn't it have been declined by the server in that case?

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: Ah right, you haven't changed that yet. We should try to stop passing around `NO_STRING_VALUE` (that constant should be removed). That can be part of a later version, though.

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: You have

```
if (x)
  doY(a, b)
else 
  doY(a, c)
```

You could consolidate that and avoid the duplication of that code block.

```
BranchError error = finUpdateListener ? null : BranchError...;
callback(finUpdateListener, error);
```

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3112,6 +3168,39 @@ public void run() {
 		});
 	}
 
+	private void processRedeemRewardsResponse(final ServerResponse resp) {
+		boolean updateListener = false;
```
shortstuffsushi: Also, `updateListener` and `finUpdateListener` aren't good variable names, try using something more descriptive like `redemptionSucceeded`.

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: hmm,For preference  I guess it may be a good idea   to have a default value like this.It will help to differentiate between the condition
1) "Key" is actually not present in the preference
2)Value for the key is intentionally set to NULL.
  Any way the API for getting a String from preference actually need a default value (can be specified as null too)

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Credit cannot be less than zero.Will remove this extra check

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: We don't want to have that value sent to the server, we should just have proper handling around null checks. `NO_STRING_VALUE` is just kind of a lazy hack around that. It needs to be removed throughout the SDK at some point and properly handled -- doesn't have to be this PR though.

Branch-SDK/src/io/branch/referral/ServerResponse.java
sojanpr: Yes I am adding the request Json to the response. The reason for doing that is server is providing enough info on some response. For e.g.   `redeemCredit` succes  response doesn't  have the amount of credit redeemed.It may not be a good idea to keep the parameters in the request locally since it can be overwritten with a successive call of same request.      
  This change actually make the response Object having the  to request parameters with them.  
What you think.Is there a suggestion for a better approach to this issue?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: make sense .will change

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Agree with you.Unfortunately all other places these variables are named like this.if we change may need to change all of them

shortstuffsushi: If you're adding something new, you might as well make it clean. No use continuing bad habits.

Branch-SDK/src/io/branch/referral/ServerResponse.java
shortstuffsushi: My suggestion: look at the latest version of the iOS SDK. I've had to work all of these same problems :smiley:

shortstuffsushi: Specifically, if you notice the way that I do callbacks in there, they keep their own context clean; they don't keep global variables around, they just keep them within the scope of the function.

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Fortunately for now I can see `NO_STRING_VALUE` is checked everywhere .So we are not sending it server in any case.
So change you are suggesting is equivalent to setting `NO_STRING_VALUE = null` .correct?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: :)

Branch-SDK/src/io/branch/referral/ServerResponse.java
sojanpr: I have  a ticket for creating many thread.I am looking forward to create  an AsyncTasks for requests so that we can control the thread pool size. That way I can confine the callbacks and the requests params a single object

shortstuffsushi: Yeah, that's exactly how it should be. That should make it a whole lot easier to deal with, without having to worry about any sort of thread management or trying to drag items around through different scopes. Everything just stays where it belongs.

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: Nope -- `NO_STRING_VALUE` and all occurrences should just be entirely removed.

Branch-SDK/src/io/branch/referral/PrefHelper.java
```diff
@@ -332,10 +333,22 @@ public String getAppKey() {
     }
 
 	public void setBranchKey(String key) {
-	    Branch_Key = key;
-    }
+		Branch_Key = key;
+		String currentBranchKey = getString(KEY_BRANCH_KEY);
+		if (key == null || currentBranchKey == null || !currentBranchKey.equals(key)) {
```
shortstuffsushi: This should be checking against the *previous* Branch key as well. The key isn't likely going to change during the actual app life time, it will more likely be changing between runs. Note in the iOS SDK, it keeps track of the last used Branch key, and compares against that.

sojanpr: `currentBranchKey` here is the last saved branch key.Any time there is a  new  branch key it is saved to the preference. Is there any thing missing?

aaustin:
Found in Alex junit testing:
- [ ] calling setIdentity correctly changes the identity_id in settings but does not clear the link cache. Same link is returned for different identity. (just discovered in iOS)
- [ ] Turn connection off. Open app, /v1/open called automatically. initSession does not return callback. Should callback with error
- [ ] Turn connection off. Open app, switch to credit history, click back so that initSession is called. Returns with error = null (should callback with error)
- [ ] App running in background. Turn off wifi. Open app. /v1/open fails. Queue up a bunch of requests like buy. Home screen. Re-enable wifi. Open up app. /v1/ open is called twice in a row, creating 2 sessions. Then the remainder of the queue executes. /v1/open should only be called a single time when the app opens.

There's some fundamental mishandling of requests in poor connectivity so I'll stop testing until we address them. Good luck.

aaustin:
Also make sure to sync because I fixed some stuff

sojanpr:
@SleeplessByte With the new changes  the request processing is sequential. only a single request is processed at a time. Also we are updating the pending request in the queue if there is any change in session state. Do you think still there is a chance for race condition?

SleeplessByte:
@sojanpr jup, I saw the single request process, which is great! I just advice to move the logic to processing time, instead of modifying queue items. I have to see the whole code to determine if threading might be an issue.

Do you want me to?

sojanpr:
@SleeplessByte I see your point how it can help in confirming multi threading issues, if happens. I will check that out

aaustin:
@sojanpr found these:
- [ ] Load TestBed app with connection, press set identity. Turn wifi off. open app. Calling setIdentity with callback does not call the callback.
- [ ] With connection out. Redeem rewards error is just 'Trouble redeeming rewards. 'make sure you have credits available'. This should be a connection error.
- [ ] With connection out. loadActionCounts and loadRewards do not pass in an error when there are connection problems. It just returns false
- [ ] With connection out. initSession does not return an error.

Stopping testing for now.

aaustin:
Ok - let's do this! God help us all.

:+1:

SleeplessByte:
EXCITING!

sojanpr:
Already merged

PR #78, opened 2015-06-17
Enhance debugger

qinweigong:
@sojanpr @aaustin


qinweigong:
Yah, I actually thought of having only a singleton BranchWindowCallback. The thing is it needs the current activity to work, which means when switching activities, developers need to set the target activity each time?

qinweigong:
Good catch!

Taking notes of our discussion:
Whenever a dialog is loaded, its onStart() is called, hence initSession is called, so debugger listener is attached to the dialog.
However, when the dialog goes away, the activity under it won't have its onStart() called, so the listener is still with the (now gone) dialog! This might well be the reason for those support requests complaining the debugger is not working!

With the new session management, we can attach the listener back to the activity after the dialog above it goes away by overriding `onActivityResumed`. Even better, we discovered a better (more sensitive) way - WindowCallback.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -935,10 +944,9 @@ public boolean onTouch(View v, MotionEvent ev) {
 					return true;
```
qinweigong: @sojanpr are you talking about this line? so we should return false here?

sojanpr:
cool
:+1:

PR #79, opened 2015-06-22
Deprecation annotation and tag

SleeplessByte:
A deprecated token such as this function should be annotated with the `java.lang.Deprecated` annotation so that compiler warnings may arise and IDE's such as Android Studio may highlight it as such.

The oracle and google spec dictate that additionally a token should be tagged in the javadoc with a reference to the replacement, if any.


SleeplessByte:
IDE example

shortstuffsushi:
You are correct, but this method should *not* be deprecated. We decided too reverse that decision, as several SDKs utilizing the Android SDK still need that. Instead, that comment should be removed from the file.

SleeplessByte:
Please update the documentation, and the settings page on the website accordingly.

shortstuffsushi:
@SleeplessByte thanks for the notice on this, it will be done :+1: closing this since it will be addressed elsewhere.

SleeplessByte:
:+1:

PR #86, opened 2015-06-23
Actually teach the user something

SleeplessByte:
This is why one must call the `Activity#setIntent( Intent )` function. Not `for some reason`.


aaustin:
:+1:

PR #89, opened 2015-06-29
Fix #88: don't leak the activity when setting the BranchWindowCallback

### LINKED ISSUE #88: Memory Leak in BranchWindowCallback
I'm using [leakcanary] (https://github.com/square/leakcanary) to detect memory leaks in my app.  I just added the Branch SDK to my app and it looks like it's leaking the activity:
```
* GC ROOT static io.branch.referral.Branch.branchReferral_
* references io.branch.referral.Branch.longPressed_
* references io.branch.referral.Branch$BranchWindowCallback$1.this$1 (anonymous class implements java.lang.Runnable)
* references io.branch.referral.Branch$BranchWindowCallback.callback_
* leaks com.foo.myapp.MainActivity instance
```

edenman:


shortstuffsushi:
Hey @edenman, is there a good way for us to test this? Doesn't the callback need to be re-attached in `onResume`?

Also, before pulling this in, can you make sure that it conforms the the current coding standards (tabs).

shortstuffsushi:
Also, can you see if this is already addressed by #76?

edenman:
@shortstuffsushi it doesn't seem to be addressed by #76

edenman:
I'll look into adding a test.  And will convert to tabs (lol).

shortstuffsushi:
Yeah, we're going to be moving away from tabs, but we want to make it all in one big push (and try to be consistent until then). It's non-ideal, but it's the way the SDK is currently. Thanks for doing that for us.

I'm guessing you're right about it not being fixed, just wanted to throw it out there as a quick check. Hopefully we're not introducing any additional leaks with that.

edenman:
@shortstuffsushi there don't seem to be any tests for the `getAutoInstance` cases

shortstuffsushi:
Oh, sorry, I didn't mean unit tests (I'm not really sure of the state of unit tests in this sdk), I meant is there a good way for us to test out this branch locally?

edenman:
Ah.  I could add LeakCanary to the TestBed app?  Not sure how much the TestBed app is actually used.

shortstuffsushi:
Hmm, it seems like that may be useful (thought @qinweigong or @sojanpr could probably speak better to that). Is there any way to use that in an automated fashion, so that CI could catch it?

edenman:
I don't think leakcanary is currently able to detect leaks in a CI-parceable fashion.

shortstuffsushi:
Ah, too bad. Well, I think adding it to the TestBed is probably appropriate, as we'll need to know about these issues as they come up. Could you add that?

Also, can you make sure the listener is re-added on resume?

edenman:
Added a second commit that adds LeakCanary.  

The listener will be re-added on resume (that's what the `debugListenerInitHistory_.remove(System.identityHashCode(activity));` line does)

Branch-SDK-TestBed/AndroidManifest.xml
```diff
@@ -14,7 +14,7 @@
         android:allowBackup="true"
         android:icon="@drawable/ic_launcher"
         android:label="@string/app_name"
-        android:name="io.branch.referral.BranchApp" >
+        android:name="io.branch.branchandroiddemo.CustomBranchApp" >
```
shortstuffsushi: Why did this change?

Branch-SDK-TestBed/build.gradle
```diff
@@ -11,7 +12,7 @@ android {
     buildToolsVersion project.ANDROID_BUILD_TOOLS_VERSION
     
     defaultConfig {
-        minSdkVersion 7
+        minSdkVersion 8
```
shortstuffsushi: Is it possible to only change this for the app, and not for the SDK (it could be, but I'm not sure from here).

Branch-SDK-TestBed/AndroidManifest.xml
shortstuffsushi: Never mind, I see what it's doing now.

Branch-SDK-TestBed/build.gradle
edenman: Yup, this minSdkVersion only affects the TestBed app, not the SDK

Branch-SDK-TestBed/AndroidManifest.xml
edenman: LeakCanary's default implementation requires an app-level hook (see the code for `CustomBranchApp`, it's pretty straightforward)

shortstuffsushi:
I'm confused by the re-adding logic; `onPause` you're removing the Branch Listener and replacing with the original listener, but how is the Branch Listener re-added when the app resumes?

edenman:
The `BranchActivityLifeCycleObserver` is still attached, so `onActivityResumed` fires when the activity resumes.  That calls `setTouchDebugInternal` which checks the `debugListenerInitHistory_`.  In `clearTouchDebugInternal` I'm removing the key from `debugListenerInitHistory_` which means the if statement condition evaluates to true (`debugListenerInitHistory_.get(System.identityHashCode(activity)) == null`) and the `BranchWindowCallback` gets installed around the original window `Callback`

shortstuffsushi:
Ahhh right, my bad. I wasn't aware of the existing implementation, but that makes sense. OK, I think this all looks good then :+1:

edenman:
Thanks!  Any idea when the next release will be?

shortstuffsushi:
As soon as @sojanpr gets it out. He's in the process now, I believe.

sojanpr:
We are testing the merged code.Will be releasing once it soon

edenman:
Awesome, thanks guys!

sojanpr:
Already deployed the new version with this fix (v1.5.10). Will be available shortly as maven central updates.

PR #96, opened 2015-07-03
Installs should override all opens

### LINKED ISSUE #92: Branch stopped working "User session has not been initialized!"
````
[behavior] CREATED: BranchMetricsLoggedInBehavior
I/BranchSDK﹕ Branch Warning: You are using your test app's Branch Key. Remember to change it to live Branch Key during deployment.
I/BranchSDK﹕ Branch Error: User session has not been initialized!
[branch-behavior] user is @derkasaurus/978
[branch-behavior] initializing for null intent data
I/BranchSDK﹕ Branch Error: User session has not been initialized!
````

- In `MyCustomApplicationClass` call `Branch.getAutoInstance( this );`.
- In `MyActivity` call `Branch.getInstance( Activity#getApplicationContext(), activity#getString( 0 R.string.api_branch_key ) );` or `Branch.getInstance( Activity#getApplicationContext()`
- Set identity using `Branch#setIdentity`
- Call `Branch#initSession( ( referringParams, error ) -> { ... } )`

The biggest problem with all this is that initSession does not return. The callback is **never** called. No error messages except for the logs. Short url generation also doesn't work.

### PR DESCRIPTION
**Pull request by SleeplessByte**
When creating a new `Branch` instance, regardless of auto sessions or not, it will create a new `ServerRequestQueue` which in turn calls `ServerRequestQueue#retrieve`.

If there, for any reason, is an `open` request in the retrieved storage, it will bump any later `install` request (because of missing user information). Since the install will always be trumped by the earlier open, no request will ever return, callbacks not called and users not initialized or installed.

This fixed #92 and #94 and should be incorperated in #76 .

Further more, from this analysis we can see that `setIdentity` requests should also be overwritten. We only care about the newest call, not earlier ones.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
+++ b/Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
@@ -286,8 +286,6 @@ public class ServerRequestQueue {
                         || req.getTag().equals(BranchRemoteInterface.REQ_TAG_REGISTER_OPEN))) {
                     if (req.getTag().equals(BranchRemoteInterface.REQ_TAG_REGISTER_INSTALL)) {
 						tag = BranchRemoteInterface.REQ_TAG_REGISTER_INSTALL;
-					} else {
-						tag = BranchRemoteInterface.REQ_TAG_REGISTER_OPEN;
 					}
 					iter.remove();
 					break;

### CODE REVIEW

**Comment by qinweigong** [comment]
:+1:

**Comment by aaustin** [comment]
Increment version and release @sojanpr

File: Branch-SDK/src/io/branch/referral/ServerRequestQueue.java [hunk 1]
```diff
@@ -286,8 +286,6 @@ public void moveInstallOrOpenToFront(String tag, int networkCount) {
                         || req.getTag().equals(BranchRemoteInterface.REQ_TAG_REGISTER_OPEN))) {
                     if (req.getTag().equals(BranchRemoteInterface.REQ_TAG_REGISTER_INSTALL)) {
```
qinweigong: We should honor the incoming tag (method argument) and remove the entire if/else block.

SleeplessByte:
Well, then an open after an install (can that happen??) would override the install. However if that can't happen, we should. I can change it?

SleeplessByte:
Should be fixed in 1.6.0

sojanpr:
Yes.This is fixed in v1.6.0

PR #98, opened 2015-07-08
Fixing sdks crashing when used below API level 14

sojanpr:
Hot fix for SDK crash on Api level <14

@aaustin @qinweigong @shortstuffsushi


aaustin:
:+1:

qinweigong:
:+1:

PR #99, opened 2015-07-12
First Cut Sharing SDK

sojanpr:
First version of Android  sharing  SDK.
@aaustin @shortstuffsushi

### DIFF
--- a/Branch-SDK-TestBed/res/layout/activity_main.xml
+++ b/Branch-SDK-TestBed/res/layout/activity_main.xml
@@ -151,6 +151,16 @@
         android:layout_marginTop="5dp"
         android:text="Referral Code"
         android:textSize="14sp" />
+
+    <TextView
+        android:id="@+id/share_btn"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_below="@+id/cmdReferralCode"
+        android:textSize="21sp"
+        android:text="Share Link"
+        android:layout_centerHorizontal="true"
+        android:layout_marginTop="20dp"/>
     
 </RelativeLayout>
     
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -262,6 +262,20 @@ public class MainActivity extends Activity {
 						.setFeature("feature1")
 						.addTag("Tag1")
 						.addTag("Tag2")
+						.setCallback(new Branch.BranchLinkShareListener() {
+							@Override
+							public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchError error) {
+								if(error != null) {
+									Log.i("BranchTestBed", "onLinkShareResponse... " + sharedLink + " " + sharedChannel + " " + error.getMessage());
+								}else{
+									Log.i("BranchTestBed", "onLinkShareResponse... " + sharedLink + " " + sharedChannel);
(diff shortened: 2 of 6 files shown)


aaustin:
Really, really nice. A couple minor things.

- [ ] Can we add the negative slide out animation on minimize
- [ ] In the no-connection case, I believe in iOS, we use a default URL if the Branch link can't be created. This allows the user to initiate a share message/copy whether or not it's a fresh Branch link or not. Can you add a default URL option instead of just failing the share?
- [ ] In the no-connection case, can we just fill the message without a link if there's backup link?
- [ ] Should we return SHARE_WITH type in the onLinkShareResponse? I believe it's currently returning the package name of the app shared to?
- [ ] Can we show a short Toast that 'copied link to clipboard!' after copy is selected
- [ ] I'm not sure how to simulate slow connections, but if /v1/url takes a second or two, or there are retries, do we have some sort of network delay handing UI? I can see folks in India running into this problem quite a bit.
- [ ] Really nitpicky, but can we move the More options.. below the Copy link option?
- [ ] Is that the standard icon for Copy? Can we use a different one? Not sure where you got it, but if there's one in this style (http://fortawesome.github.io/Font-Awesome/icon/clipboard/) that'd be idea.

sojanpr:
- [ ] Can we add the negative slide out animation on minimize
    -  Since we deploy SDK as jar we cannot add resource xml files. At this point window animations can 
         be set only with animation defined in resource xml.

- [ ] In the no-connection case, I believe in iOS, we use a default URL if the Branch link can't be created. 
This allows the user to initiate a share message/copy whether or not it's a fresh Branch link or not. Can you add a default URL option instead of just failing the share?
    -  Added an option for setting default url in builder. 

- [ ] In the no-connection case, can we just fill the message without a link if there's backup link?
     -  In no-connection case or any error the default url will be shared (if a default url is specified). 

- [ ] Should we return SHARE_WITH type in the onLinkShareResponse? I believe it's currently returning the package name of the app shared to?
     -  We are returning the Application Name(Facebook ,Twitter etc). Since we have  "more" option which
        shows all different application avaible on the device, not necessarily the selected app is defined in 
         SHARED_WITH always. 
 
- [ ] Can we show a short Toast that 'copied link to clipboard!' after copy is selected
     -  Changed toast text to 'copied link to clipboard!'

- [ ] I'm not sure how to simulate slow connections, but if /v1/url takes a second or two, or there are retries, do we have some sort of network delay handing UI? I can see folks in India running into this problem quite a bit.
     -  Slow connections case is tested on emulator. Emulator provides setting to specify connection speed
        and performance. The share dialog window will be up in disabled state until  the v1/url is respond in 
        order to prevent the activity context change by user. Devs can override this feature by calling
         `cancelSharingDialog()` 

- [ ] Really nitpicky, but can we move the More options.. below the Copy link option?
      -  More options moved to the bottom.

- [ ] Is that the standard icon for Copy? Can we use a different one? Not sure where you got it, but if there's one in this style (http://fortawesome.github.io/Font-Awesome/icon/clipboard/) that'd be idea.
      - The icons currently showing for "More" and "Copy link" are standard android icons for  "More" and 
         "Save". We cannot add resources to the SDK since we are deploying  Jar file. There is a "Copy" 
         Icon similar to the one shown in the link above but thats not available on all API levels.

sojanpr:
Added a slide down animation when user selected an application on the closer dialog

aaustin:
:+1: beautiful

sojanpr:
Merged with master and conflicts resolved

PR #100, opened 2015-07-15
Removing Branch Exception with getInstance() Method

sojanpr:
Removing Branch exception thrown from getInstance()
@aaustin @qinweigong


sojanpr:
After removing the exception, it seems like getInstance() method is really confusing. Is it a good idea to remove the getInstance() method?

aaustin:
The problem I had with the exception is that the developer could go through all the work and publish an app, only to realize that it was excepting on v14 or below. At least now, they will know in development time if they need to use getInstance or getInstance(ctx). Is there anyway to add a note in Javadoc to send them to the latter version if they need to support < v14?

aaustin:
:+1: pending notice for devs

PR #104, opened 2015-07-18
Checking for Activity null case while stripping the URI prams

sojanpr:
@shortstuffsushi  @aaustin 

Adding a null check while stripping install URI params.


aaustin:
:+1:

PR #105, opened 2015-07-19
Hotfix: Auto sessions are ignoring intents

aaustin:
The auto session init code was ignoring the intents and intent string,
so that link identifiers were not being used. This fixes that issue

@sojanpr


sojanpr:
:+1:

PR #106, opened 2015-07-21
First cut changes for Android auto deep linking

sojanpr:
First cut changes for Android Auto Deep link

@aaustin @shortstuffsushi 

 Note : The configurations for auto deep linking is done in manifest since devs will prefer less code changes.  Do we need to anticipate a situation where dev want to change auto deep linking params during run time?

### DIFF
--- a/Branch-SDK-TestBed/AndroidManifest.xml
+++ b/Branch-SDK-TestBed/AndroidManifest.xml
@@ -31,16 +31,24 @@
             	<category android:name="android.intent.category.BROWSABLE" />
         	</intent-filter>
         </activity>
-        <activity
-            android:name="io.branch.branchandroiddemo.CreditHistoryActivity">
-        </activity>
-        <activity
-            android:name="io.branch.branchandroiddemo.ReferralCodeActivity">
+
+        <activity android:name="io.branch.branchandroiddemo.CreditHistoryActivity"/>
+
+        <activity android:name="io.branch.branchandroiddemo.ReferralCodeActivity"/>
+
+        <activity android:name=".AutoDeepLinkTestActivity">
+            <!-- Keys for auto deep linking this activity -->
+            <meta-data android:name="io.branch.sdk.auto_link_keys" android:value="auto_deeplink_key_1,auto_deeplink_key_2" />
+            <!-- Optional request ID for launching this activity on auto deep link key matches -->
+            <meta-data android:name="io.branch.sdk.auto_link_request_code" android:value="201" />
         </activity>
 
         <meta-data android:name="io.branch.sdk.TestMode" android:value="false" /> <!-- Set to true to use Branch_Test_Key -->
         <meta-data android:name="io.branch.sdk.BranchKey" android:value="key_live_jbgnjxvlhSb6PGH23BhO4hiflcp3y8kx" />
         <meta-data android:name="io.branch.sdk.BranchKey.test" android:value="key_test_jkptOCZtmtxhOMZ11ynbXecdDCd93cbr" />
+
+        <!-- Optional. Set to true to disable auto deep link feature-->
+        <meta-data android:name="io.branch.sdk.auto_link_disable" android:value="false"/>
 
         <uses-library android:name="android.test.runner" />
     </application>
--- a/Branch-SDK-TestBed/AndroidManifest.xml
+++ b/Branch-SDK-TestBed/AndroidManifest.xml
@@ -53,12 +53,23 @@
         <uses-library android:name="android.test.runner" />
 
         <!-- Add this for branch for install referrer tracking-->
-        <receiver
-            android:name="io.branch.referral.InstallListener" android:exported="true">
+        <receiver android:name="io.branch.referral.InstallListener" android:exported="true">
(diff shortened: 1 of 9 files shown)


Branch-SDK-TestBed/AndroidManifest.xml
```diff
@@ -31,18 +31,35 @@
...
+
+        <activity android:name="io.branch.branchandroiddemo.CreditHistoryActivity"/>
+
+        <activity android:name="io.branch.branchandroiddemo.ReferralCodeActivity"/>
+
+        <activity android:name=".AutoDeepLinkTestActivity">
+            <!-- Keys for auto deep linking this activity -->
+            <meta-data android:name="io.branch.sdk.auto_link_keys" android:value="auto_deeplink_key_1,auto_deeplink_key_2" />
+            <!-- Optional request ID for launching this activity on auto deep link key matches -->
+            <meta-data android:name="io.branch.sdk.auto_link_request_code" android:value="201" />
```
aaustin: What's the difference between the request code and the deep link keys? Do we need both/

sojanpr: Request code is optional . In most of the case Devs would need an activity result on finishing activity for handling flow differently for auto deep link flow and regular app flow launch. So we need  to `startActivityForResult`. The same  `requestCode` is returned on activity result. It is very common for application to have a base activity which handle `onActivityResult()` and take action depends on the request code. 
   If they don't care about this they don't need  to provide the request code in meta.We are starting the Activity with a default request code.

  Deep link keys are  the actuals keys in the params which this activity need deep linked to

aaustin: Ah I see - so their main activity would call onActivityResult and pass in that code?

aaustin:
Question: Have you been able to test the duration from install -> broadcast received? Does it happen *before* the user opens the app or after? Can you explain the lifecycle for me?

Branch-SDK-TestBed/AndroidManifest.xml
sojanpr: Yes when the auto linked activity finishes it will call `onActivityResult()` of parent (Most cases the main activity) .So Devs can pass any parameter denoting that this is launched by auto linking and handle the flow accordingly.

Branch-SDK-TestBed/AndroidManifest.xml
```diff
@@ -31,18 +31,35 @@
...
         <meta-data android:name="io.branch.sdk.BranchKey.test" android:value="key_test_jkptOCZtmtxhOMZ11ynbXecdDCd93cbr" />
 
+        <!-- Optional. Set to true to disable auto deep link feature-->
+        <meta-data android:name="io.branch.sdk.auto_link_disable" android:value="false"/>
+
         <uses-library android:name="android.test.runner" />
+
+        <!-- Add this for branch for install referrer tracking-->
+        <receiver
+            android:name="io.branch.referral.InstallListener" android:exported="true">
```
aaustin: Great. Now to add a little complexity to your life. Here are two additional possibilities:
1. The app already has their own INSTALL_REFERRER broadcast receiver
2. The app already has their own INSTALL_REFERRER broadcast receiver and it's subclassed by another service

(similar to Application)

aaustin:
Last thought: Do we want to expose some mechanism for them to close the Activity in an easy way? Do they just call finish() or how does that work? Do we need to tell them to register their main activity for onActivityResult? I'm just asking as we should provide some best practices - you know?

sojanpr:
I tried only with adb command. According to the documentation this broad cast can happen any time after install and before application `onCreate()`. So we should not depend on any lifecycle methods. Thats the reason why we added the `IntialisationListener` out side Branch and storing the linkClickID for future use. 
When the first activity starts we will have the linkClickID from the broadcast. With ADB we cannot simulate the situation. We need  to try this with uploading play store app. Will test it with a real scenario

Branch-SDK-TestBed/AndroidManifest.xml
sojanpr: I will double check this and confirm. 
AFIK multiple receivers  can be registered for an ordered broadcast and receive the broadcast. Provided one of the receiver is not consuming it by aborting the broadcast. So even if they have different broadcast receivers Branch installListener should get the broadcast.

sojanpr:
For Auto link , I think we should give the complete freedom to the devs to write their activity and app flows. They should be able to know whether this activity is launched because of Branch Auto init feature.
`isAutoDeepLinkLaunch()` api helps for that. There may case where  they want to navigate to another activity on user interaction  from the auto linked activity. I think with the `isAutoDeepLinkLaunch()` api an d the optional request code we are providing a better practice

aaustin:
Can you add this result monitoring to the testbed app so that I know how to document it properly?

sojanpr:
Added example for monitoring activity finish for auto deep linked activity

sojanpr:
Added an example to show how to do m multiple  attribution services along with branch to check Instal referrence.

aaustin:
Really awesome man! :+1:

PR #107, opened 2015-07-22
Clearing LinkCache on key change

sojanpr:
Clearing the branch link cache on a  new key Branch key

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -394,9 +394,10 @@ public class Branch {
 		branchReferral_.context_ = context.getApplicationContext();
 		if (branchKey.startsWith("key_")) {
 			boolean isNewBranchKeySet = branchReferral_.prefHelper_.setBranchKey(branchKey);
-			//on setting a new key clear link cache
+			//on setting a new key clear link cache and pending requests
 			if (isNewBranchKeySet) {
 				branchReferral_.linkCache_.clear();
+				branchReferral_.requestQueue_.clear();
 			}
 		} else {
 			branchReferral_.prefHelper_.setAppKey(branchKey);
@@ -416,9 +417,10 @@ public class Branch {
 			} else {
 				isNewBranchKeySet = branchReferral_.prefHelper_.setBranchKey(branchKey);
 			}
-			//on setting a new key clear link cache
+			//on setting a new key clear link cache and pending requests
 			if (isNewBranchKeySet) {
 				branchReferral_.linkCache_.clear();
+				branchReferral_.requestQueue_.clear();
 			}
 		}
 		branchReferral_.context_ = context;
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -393,7 +393,11 @@ public static Branch getInstance(Context context, String branchKey) {
 		}
 		branchReferral_.context_ = context.getApplicationContext();
 		if (branchKey.startsWith("key_")) {
-			branchReferral_.prefHelper_.setBranchKey(branchKey);
+			boolean isNewBranchKeySet = branchReferral_.prefHelper_.setBranchKey(branchKey);
+			//on setting a new key clear link cache
+			if (isNewBranchKeySet) {
+				branchReferral_.linkCache_.clear();
```
shortstuffsushi: Link cache doesn't get persisted between launches, does it?

shortstuffsushi: Instead, the goal of "clearing when key changes" is supposed to remove identity, install/session params, etc.

shortstuffsushi: See the [iOS SDK implementation](https://github.com/BranchMetrics/Branch-iOS-Deep-Linking-SDK/blob/aa0fa04364ddbc5abb19461e01e1f93b5e48e84e/Branch-SDK/Branch-SDK/Branch.m#L829)

sojanpr: Link Cache is not persisted across launches how ever there is option for user to set a new key in the same session by calling the API `Branch getInstance(Context context, String branchKey)`

shortstuffsushi: In that case, you should just do both :smiley:

shortstuffsushi:
Hmm, the iOS SDK is not clearing the Link Cache (probably should be). However, I noticed it is clearing the queue, which is not being done here. Does that need to occur as well?

Branch-SDK/src/io/branch/referral/PrefHelper.java
```diff
@@ -340,13 +340,20 @@ public String getAppKey() {
...
+	 * @param key A {@link String} representing Branch Key.
+	 * @return A {@link Boolean} which is true if the key set is a new key. On Setting a new key need to clear all preference items.
+	 */
+	public boolean setBranchKey(String key) {
 		Branch_Key = key;
 		String currentBranchKey = getString(KEY_BRANCH_KEY);
 		if (key == null || currentBranchKey == null || !currentBranchKey.equals(key)) {
 			clearPrefOnBranchKeyChange();
 			setString(KEY_BRANCH_KEY, key);
+			return true;
```
aaustin: nice and clean

Branch-SDK/src/io/branch/referral/Branch.java
aaustin: Agreed - need to clear more than link cache. all Branch ids.

aaustin:
:+1:

sojanpr:
Added changes to clear pending server request on key change.

shortstuffsushi:
:+1:

PR #108, opened 2015-07-25
Removing hard tabs to 4 spaces

sojanpr:
Removing hard to soft 4 space tabs
@shortstuffsushi


shortstuffsushi:
This misses a ton of files, and doesn't actually address all the tabs even in some of the files you did touch. Try doing a tab search across the project.

shortstuffsushi:
I believe what I jus pushed should be all inclusive.

shortstuffsushi:
This also enforces newline at end of file and no trailing spaces (on files I've opened, since my Sublime settings see to that for me).

sojanpr:
Removed the tabs in doc and comments in the pending files

shortstuffsushi:
This looks good to me :+1: 

@sojanpr @aaustin @qinweigong Make sure you change your editors to use SPACES instead of TABS as well, so we don't need to do this again in the future.

PR #109, opened 2015-07-29
Restricting URI scheme  reading only to debug mode

sojanpr:
Restricting URI scheme  reading only to debug mode


sojanpr:
@aaustin Adding Changes  to restrict Reading URI scheme only while debugging

aaustin:
This is a hard one to read, but it looks like the the few lines related to URI scheme reading are good :+1:

PR #111, opened 2015-08-01
Adding Builders for Short links

sojanpr:
Adding a builder to create  short url in Asynchronous or synchronous
method. Deprecating previous overloaded method for creating short url

@aaustin @shortstuffsushi


Branch-SDK/src/io/branch/referral/builders/BranchShortUrlBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     * <p>Configures and requests a short URL to be generated by the Branch servers, via a asynchronous
+     * call; The {@link Branch.BranchLinkCreateListener} is called back with the url when the url is generated.</p>
+     *
+     * @param callback A {@link Branch.BranchLinkCreateListener} callback instance that will trigger
+     */
+
+    public void getShortUrlAsync(Branch.BranchLinkCreateListener callback) {
```
Kirkkt: since it returns void, can we rename it to generate... instead of get...?

sojanpr: Nice changes the build methods name 
getShoertLinkSync() -> getShortLink()

getShortLinkAsync() ->  generateShortLink()

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1985,12 +2027,78 @@ public void getShortUrl(int type, Collection<String> tags, String channel, Strin
...
+     *                 link.
+     * @param channel  A {@link String} denoting the channel that the link belongs to. Should not
+     *                 exceed 128 characters.
+     * @param feature  A {@link String} value identifying the feature that the link makes use of.
+     *                 Should not exceed 128 characters.
+     * @param stage    A {@link String} value identifying the stage in an application or user flow
+     *                 process. Should not exceed 128 characters.
+     * @param params   A {@link JSONObject} value containing the deep linked params associated with
+     *                 the link that will be passed into a new app session when clicked
+     * @param duration An {@link int} the time that Branch allows a click to remain outstanding
```
Kirkkt: nit: I understand this is deprecated, but \@param duration appeared twice

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1652,38 +1677,42 @@ public String getShortUrlSync(int type, Collection<String> tags, String channel,
...
-    /**
-     * <p>Configures and requests a short URL to be generated by the Branch servers.</p>
-     *
-     * @param callback A {@link BranchLinkCreateListener} callback instance that will trigger
-     *                 actions defined therein upon receipt of a response to a create link request.
-     */
-    public void getShortUrl(BranchLinkCreateListener callback) {
-        generateShortLink(null, LINK_TYPE_UNLIMITED_USE, 0, null, null, null, null, stringifyParams(null), callback, true);
+
+    public void getReferralUrl(String channel, JSONObject params, BranchLinkCreateListener callback) {
```
shortstuffsushi: Where are the Javadocs for this method?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1985,12 +2027,78 @@ public void getShortUrl(int type, Collection<String> tags, String channel, Strin
...
+     * @see BranchLinkData#putFeature(String)
+     * @see BranchLinkData#putStage(String)
+     * @see BranchLinkData#putParams(String)
+     * @see BranchLinkData#putDuration(int)
+     * @see BranchLinkCreateListener
+     * @deprecated use {@link io.branch.referral.builders.BranchShortUrlBuilder} instead.
+     */
+    @Deprecated
+    public String generateShortLink(final String alias, final int type, final int duration, final Collection<String> tags, final String channel, final String feature, final String stage, final String params, BranchLinkCreateListener callback, boolean async) {
+        ServerRequestCreateUrl req = new ServerRequestCreateUrl(context_, alias, type, duration, tags,
```
shortstuffsushi: Why not just run this through the builder?

shortstuffsushi:
Short of the couple of things I mentioned, looks good to me :+1:

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Builder is internally calling is this method to create the link. Builder has  no logic for link creation or link cache.

sojanpr:
@shortstuffsushi What are the missing things to improve the experience for devs?

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: Wait, the builder is internally calling a deprecated method? How is that going to work when you removed that method? You should extract that logic out to the builder and then update the deprecated methods within Branch to call the builder. That will allow you to easily remove these methods in the future, and make it so you don't need logic in more than one place.

Branch-SDK/src/io/branch/referral/builders/BranchShortUrlBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    //------------------------- Utility methods -----------------------//
+
+    protected static String stringifyParams(JSONObject params) {
+        if (params == null) {
+            params = new JSONObject();
+        }
+        try {
+            params.put("source", "android");
+        } catch (JSONException e) {
+            e.printStackTrace();
```
shortstuffsushi: We should probably do something nicer than print stack trace (or log a message with it, as well as the stack trace, via `Log.e`).

Branch-SDK/src/io/branch/referral/builders/BranchShortUrlBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    public void generateShortUrl(Branch.BranchLinkCreateListener callback) {
+        String shortUrl = null;
+        if (branchReferral_ != null) {
+            branchReferral_.generateShortLink(alias_, type_, duration_, tags_, channel_, feature_, stage_, stringifyParams(params_), callback, true);
+        }
+    }
+
+    //------------------------- Utility methods -----------------------//
+
+    protected static String stringifyParams(JSONObject params) {
```
shortstuffsushi: Why is this protected and not just private? Also, `stringifyParams` is somewhat of a misnomer since you're modifying it and then just calling `toString` on it. You're not really doing any logic here other than adding an extra param.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1985,12 +2040,78 @@ public void getShortUrl(int type, Collection<String> tags, String channel, Strin
...
+     *                 exceed 128 characters.
+     * @param feature  A {@link String} value identifying the feature that the link makes use of.
+     *                 Should not exceed 128 characters.
+     * @param stage    A {@link String} value identifying the stage in an application or user flow
+     *                 process. Should not exceed 128 characters.
+     * @param params   A {@link JSONObject} value containing the deep linked params associated with
+     *                 the link that will be passed into a new app session when clicked
+     * @param callback A {@link BranchLinkCreateListener} callback instance that will trigger
+     *                 actions defined therein upon receipt of a response to a create link request.
+     * @param  async   A {@link Boolean} whose value is true if the link should be created asynchronously.
```
shortstuffsushi: What is this? This method can be either sync or async? That's a little disturbing. Is this a new method, or has this always existed? Shouldn't the presence of a callback be enough to determine whether or not to treat it as async, if you really need to support both?

shortstuffsushi:
Nothing is missing re: dev experience, I just meant the couple of cleanup items that myself and @Kirkkt had mentioned, most of which you got in that last commit. Reading through further (on my computer vs my phone), I came up with some additional questions, however.

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: This method is originally there.It is supporting both .It is made back to private and is deprecated.

shortstuffsushi: I'm not a fan of this method. It doesn't have to be removed now, but the fact that it has different return types *and* different control flows depending on a flag is offputting to say the least. Deprecated, so that's good at least..

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
```diff
@@ -119,6 +119,17 @@ public void onRequestSucceeded(ServerResponse resp, Branch branch) {
         }
     }
 
+    /**
+     * Calls the callback with the URL. This should be called on finding an existing url
+     * up on trying to create a URL asynchronously
+     * @param url
+     */
+    public void onUrlAvailable(String url){
```
shortstuffsushi: Where are this an `isAsync` used?

sojanpr: it `ServerRequestCreateUrl.isAsync()` used to determine whether this is a synchronous request

sojanpr:
Updated with builders for ReferralUrl and ContentUrl creation

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
shortstuffsushi: Right, but where are either of these used?

Branch-SDK/src/io/branch/referral/BranchUtil.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Replaces illegal characters in the given JSONObject.
+     *
+     * @param inputObj JSONObject to remove illegal characters.
+     * @return A {@link JSONObject} with illegal characters replaced.
+     */
+    public static JSONObject filterOutBadCharacters(JSONObject inputObj) {
+        JSONObject filteredObj = new JSONObject();
+        if (inputObj != null) {
+            Iterator<?> keys = inputObj.keys();
```
shortstuffsushi: Don't these have to be `String`s?

Branch-SDK/src/io/branch/referral/BranchUtil.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Convert the given JSONObject to string and adds source value as
+     *
+     * @param params JSONObject to convert to string
+     * @return A {@link String} value representing the JSONObject
+     */
+    public static String stringifyParams(JSONObject params) {
+        if (params == null) {
```
shortstuffsushi: It's not possible for this to be null since `filterOutBadCharacters` always returns a `JSONObject`.

Branch-SDK/src/io/branch/referral/BranchUtil.java
```diff
@@ -0,0 +1,66 @@
...
+        return stringifyParams(filterOutBadCharacters(params));
+    }
+
+    /**
+     * Convert the given JSONObject to string and adds source value as
+     *
+     * @param params JSONObject to convert to string
+     * @return A {@link String} value representing the JSONObject
+     */
+    public static String stringifyParams(JSONObject params) {
```
shortstuffsushi: I still argue this isn't an appropriate method name.

Branch-SDK/src/io/branch/referral/BranchUrlBuilder.java
```diff
@@ -0,0 +1,114 @@
...
+                    getJsonParamString(params_), callback, true);
+            branchReferral_.generateShortLinkInternal(req);
+        }
+    }
+
+    /**
+     * Get the the JSON Parameters specified for the link in string format
+     * @return A{@link String} representation for the JSON parameters specified for creating this link.
+     */
+    protected String getJsonParamString(JSONObject params){
```
shortstuffsushi: Why not remove this and make the other methods call the Util method?

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+     *              process. Should not exceed 128 characters.
+     * @return A {@link BranchShortLinkBuilder} instance.
+     */
+    public BranchShortLinkBuilder setStage(String stage) {
+        this.stage_ = stage;
+        return this;
+    }
+
+    /**
+     * <p>Set the number of times the link should perform deep link.</p>
```
shortstuffsushi: This is a confusing description. The user passes an `int` into this function, so saying "the number of times it deep links" (which is wrong, it's how many times reward rules will be triggered) indicates that I can set that to an arbitrary number and should see that as the result.

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * <p>Adds the the given key value pair to the parameters associated with this link.</p>
+     *
+     * @param key   A {@link String} with value of key for the parameter
+     * @param value A {@link String} with value of value for the parameter
+     * @returnA {@link BranchShortLinkBuilder} instance.
+     */
+    public BranchShortLinkBuilder addParameters(String key, String value) {
```
shortstuffsushi: Why is this not just part of the base class?

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+        super(context);
+    }
+
+    /**
+     * <p>Adds a tag to the iterable collection of name associated with a deep link.</p>
+     *
+     * @param tag {@link String} tags associated with a deep link.
+     * @return A {@link BranchShortLinkBuilder} instance.
+     */
+    public BranchShortLinkBuilder addTag(String tag) {
```
shortstuffsushi: Same, could be base class.

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * <p> Set a name that identifies the feature that the link makes use of.</p>
+     *
+     * @param feature A {@link String} value identifying the feature that the link makes use of.
+     *                Should not exceed 128 characters.
+     * @return A {@link BranchShortLinkBuilder} instance.
+     */
+    public BranchShortLinkBuilder setFeature(String feature) {
```
shortstuffsushi: Really, this is the only function that isn't available to the other two.

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
sojanpr: Used  in `Branch.generateShortLinkInternal()`

Branch-SDK/src/io/branch/referral/BranchUtil.java
sojanpr: There are methods in branch which uses this method without calling `filterOutBadCharacters`.
For eg. `generateShortLink(null, LINK_TYPE_UNLIMITED_USE, 0, null, null, null, null,BranchUtil.stringifyParams(null), callback, true)`

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
sojanpr: Cannot. Since this is a builder ,it has to return the child class instance ignored to call other builder methods or the build method in the class

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
sojanpr: Same as above

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
sojanpr: For Referral url and Content url user has to `Channel` `Params` and `Tags` are the only optional item that can be set by dev

Branch-SDK/src/io/branch/referral/BranchUtil.java
shortstuffsushi: Why would it be calling stringify on null?

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
shortstuffsushi: You could make the Base class take generics for the builder, that would eliminate that need.

```
public class BaseBuilder<T extends BaseBuilder> {
    public T addParameter(String key, String value) {
        _params.put(key, value);
        return this;
    }
}
```

That sort of thing.

shortstuffsushi: Heck, could make this even easier: drop the subclasses altogether. Just do `getShortUrl`, `getReferralUrl`, and `getContentUrl` off the base class, where those two methods set the feature internally.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1885,111 +1939,228 @@ public void getShortUrl(String channel, String feature, String stage, JSONObject
      *                 the link that will be passed into a new app session when clicked
      * @param callback A {@link BranchLinkCreateListener} callback instance that will trigger
      *                 actions defined therein upon receipt of a response to a create link request.
+     * @param  async   A {@link Boolean} whose value is true if the link should be created asynchronously.
```
shortstuffsushi: Don't need this; the presence of a callback is sufficient to determine that it's async.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1885,111 +1939,228 @@ public void getShortUrl(String channel, String feature, String stage, JSONObject
...
+    @Deprecated
+    private String generateShortLink(final String alias, final int type, final int duration, final Collection<String> tags, final String channel, final String feature, final String stage, final String params, BranchLinkCreateListener callback, boolean async) {
+        ServerRequestCreateUrl req = new ServerRequestCreateUrl(context_, alias, type, duration, tags,
+                channel, feature, stage,
+                params, callback, async);
+
+        if (!req.constructError_ && !req.handleErrors(context_)) {
+            if (linkCache_.containsKey(req.getLinkPost())) {
+                String url = linkCache_.get(req.getLinkPost());
+                if (callback != null) {
```
shortstuffsushi: Keep in mind that they're expecting this to be called asynchronously, and you're calling it immediately in the same context here, which could lead to some weird behavior. Maybe throw it in the Looper.next function (don't remember what that's called), just to push it off this run loop.

shortstuffsushi: You're also returning here, should probably pick one or the other and not both.

shortstuffsushi: Really this should just be broken into two methods. You shouldn't have a sync and async flow through the same method.

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
shortstuffsushi: I don't understand the flow there. You're manually setting it on the request so that it can call the callback, but it's sync, so it's just getting returned anyway?

shortstuffsushi:
This is getting messy in terms of comments. My primary recommendation is to stop merging sync/async into the same methods. It's going to lead to confusion and tears down the road.

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Please  note that these are  private methods that are not exposed outside.

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
sojanpr: There are issue wi the both of the above approach.
The reason for making them into separate class is  that `getReferralURL`,`GetContentURl`, and `getShortURL` has different methods that can be set form the builder. If they are  in same builder all the builder methods are applicable to all of them. For e.g. `getReferralURl`should not have builder method `setType()` where as `getShortUrl` need that

shortstuffsushi: Leaving the `setType` exposed would be fine, it would just be overridden when you called `getContentUrl` for example.

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
```diff
@@ -0,0 +1,153 @@
...
+        super.addTag(tag);
+        return this;
+    }
+
+    /**
+     * <p>Adds the the given key value pair to the parameters associated with this link.</p>
+     *
+     * @param key   A {@link String} with value of key for the parameter
+     * @param value A {@link String} with value of value for the parameter
+     * @returnA {@link BranchShortLinkBuilder} instance.
```
Kirkkt: super nit: the convention seems to be:

` * @return this {@link BSLB}` (no period)

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: I'm not sure what it being private has to do with it, I think this is a bad choice for us to be using internally as well :stuck_out_tongue: I think this is going to lead to bugs within the SDK, since we're mixing code flow control patterns. If you'd rather, this can be refactored later, but I think this is introducing technical debt.

Branch-SDK/src/io/branch/referral/BranchShortLinkBuilder.java
sojanpr: What is the advantage in making `getContentURl`, `GetReferralURL` in to a single class ?
 Don't you think this may create  confusion to users that some parameters they set with link is not working. For e.g. for a referral URL the link is created with `LINK_TYPE_UNLIMITED_USE ` . so if users see an option to setLink type to `LINK_TYPE_ONE_TIME_USE` the will expect the same for the link behavior.

shortstuffsushi: The advantage is strictly in terms of removing redundancy. From here, it appears you have three separate classes that contain ~90% of the same code. Is there no way to make them all inherit from a single class? Even if ShortUrl has it's own builder, maybe all three extend from an AbstractUrlBuilder class or something.

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: I agree that it is not the right code path, Since this PR is already flooded with lot of builder changes , I would like to do it an spirit PR.

shortstuffsushi:
OK, I think this will work for now :+1:

aaustin:
Just scanning, all seems fine :+1: Hopefully you tested extensively.

sojanpr:
Merged with master

PR #112, opened 2015-08-06
Adding Auto deep linked parameters to Intent

sojanpr:
The data associated with the link are added to the Intent extras. The
values  can be retried from the intent extra with the same key used
for creating the  link

@aaustin


aaustin:
:+1: looks great

PR #114, opened 2015-08-07
Fix: HttpClient leak in Strict mode

sojanpr:
Preventing an SMV issue with HTTP Post and GET which can delay the
request response.
@aaustin @qinweigong @shortstuffsushi

### DIFF
--- a/Branch-SDK/src/io/branch/referral/RemoteInterface.java
+++ b/Branch-SDK/src/io/branch/referral/RemoteInterface.java
@@ -185,6 +185,7 @@ class RemoteInterface {
     private ServerResponse make_restful_get(String baseUrl, String tag, int timeout, int retryNumber, boolean log) {
         String modifiedUrl = baseUrl;
         JSONObject getParameters = new JSONObject();
+        HttpClient httpClient = null;
         if (addCommonParams(getParameters, retryNumber)) {
             modifiedUrl += this.convertJSONtoString(getParameters);
         } else {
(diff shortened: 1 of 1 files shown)


shortstuffsushi:
Isn't this shutting down between every single request? If you're calling that on the same instance, does it "restart" on the next request, or will subsequent requests fail due to the connection manager being shut down?

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -220,6 +222,10 @@ private ServerResponse make_restful_get(String baseUrl, String tag, int timeout,
         } catch (IOException ex) {
             if (log) PrefHelper.Debug(getClass().getSimpleName(), "IO exception: " + ex.getMessage());
             return new ServerResponse(tag, 500);
+        } finally {
+            if (httpClient != null && httpClient.getConnectionManager() != null) {
```
shortstuffsushi: You could encapsulate this inside a separate function:

```
private HttpResponse executeRequest(HttpRequest request) {
     HttpResponse resp = null;
     HttpClient client = null;
     try {
         client = getGenericHttpClient(timeout);
         resp = client.execute(request);
     }
     finally {
         if (client != null && client.getConnectionManager() != null) {
             client.getConnectionManager().shutdown();
         }
     }

     return resp;
}
```

shortstuffsushi: You could do most of this for the two functions, really all you need to do is construct the Request outside and pass it in.

sojanpr:
The `getGenericHttpClient` is creating a new instance of HttpClient for each request
It's not safe to reuse same HttpClient, since it is not thread safe. Alternative approach is  to  use `'AndroidHttpClient` which is thread safe.

   Creating a new method for executing for get and post request looks nice .I really wish to keep the changes minimum for this Hot fix. Will add this change with next  release if thats good.

shortstuffsushi:
Yeah, that's fine. Looks OK to me then :+1:

aaustin:
:+1:

PR #115, opened 2015-08-08
Fix setDebug and change isReferrable

aaustin:
@sojanpr

Set debug is currently not working, so we’ll migrate to the manifest
line. Also, we’re going to change the default of isReferrable to be
true for all instances.


aaustin:
I'd rather safely fallback. Going to merge for now and we can deploy the 1.8.4 in a bit

PR #116, opened 2015-08-09
Activity to deeplink path mapping

aaustin:
@sojanpr This allows for the developer to specify paths to match to an Activity, like so:
```xml
<!-- Deep link path for auto deep linking -->
<meta-data android:name="io.branch.sdk.auto_link_path" android:value="custom/path/{ variable },another/path/" />
```

Then, when $deeplink_path or $android_deeplink_path match, we'll auto deep link. This is very similar to DLD.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,9 @@ build/
 *.gen
 bin/
 gen/
+
+Branch-SDK-TestBed/gradle/wrapper/gradle-wrapper.jar
+
+Branch-SDK-TestBed/gradlew
+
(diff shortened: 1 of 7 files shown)


shortstuffsushi:
This is adds hard tabs back into the project, can you please switch them to spaces (now especially, so we don't have a mix of spaces and tabs)?

aaustin:
think I got the tabs - hard to see them in Github

sojanpr:
Looks cool :+1:

PR #117, opened 2015-08-12
Fix: Crash for Install Listner 

sojanpr:
Fix for install listener crashing when the the param is not in key-value pairs

@aaustin

### DIFF
--- a/Branch-SDK/src/io/branch/referral/InstallListener.java
+++ b/Branch-SDK/src/io/branch/referral/InstallListener.java
@@ -36,7 +36,7 @@ public class InstallListener extends BroadcastReceiver {
 
                 for (String referrerParam : referralParams) {
                     String[] keyValue  =  referrerParam.split("=");
-                    if(keyValue.length > 1) { // To make sure that there is one key value pair in referrer
+                    if (keyValue.length > 1) { // To make sure that there is one key value pair in referrer
                         referrerMap.put(URLDecoder.decode(keyValue[0], "UTF-8"), URLDecoder.decode(keyValue[1], "UTF-8"));
                         Log.d("installTest", keyValue[0] + " = " + keyValue[1]);
                     }
--- a/Branch-SDK/src/io/branch/referral/InstallListener.java
+++ b/Branch-SDK/src/io/branch/referral/InstallListener.java
@@ -26,7 +26,7 @@ public class InstallListener extends BroadcastReceiver {
 
     @Override
     public void onReceive(Context context, Intent intent) {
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/InstallListener.java
```diff
@@ -36,8 +36,10 @@ public void onReceive(Context context, Intent intent) {
 
                 for (String referrerParam : referralParams) {
                     String[] keyValue  =  referrerParam.split("=");
-                    referrerMap.put(URLDecoder.decode(keyValue[0],"UTF-8"),URLDecoder.decode(keyValue[1],"UTF-8"));
-                    Log.d("installTest", keyValue[0] +" = " + keyValue[1]);
+                    if(keyValue.length > 1) { // To make sure that there is one key value pair in referrer
+                        referrerMap.put(URLDecoder.decode(keyValue[0], "UTF-8"), URLDecoder.decode(keyValue[1], "UTF-8"));
+                        Log.d("installTest", keyValue[0] + " = " + keyValue[1]);
```
aaustin: do we want to always log in their production app? just noticed this now

Branch-SDK/src/io/branch/referral/InstallListener.java
```diff
@@ -36,8 +36,10 @@ public void onReceive(Context context, Intent intent) {
 
                 for (String referrerParam : referralParams) {
                     String[] keyValue  =  referrerParam.split("=");
-                    referrerMap.put(URLDecoder.decode(keyValue[0],"UTF-8"),URLDecoder.decode(keyValue[1],"UTF-8"));
-                    Log.d("installTest", keyValue[0] +" = " + keyValue[1]);
+                    if (keyValue.length > 1) { // To make sure that there is one key value pair in referrer
+                        referrerMap.put(URLDecoder.decode(keyValue[0], "UTF-8"), URLDecoder.decode(keyValue[1], "UTF-8"));
+                        Log.d("installTest", keyValue[0] + " = " + keyValue[1]);
```
aaustin: but why do we want to print logs in their app?

Branch-SDK/src/io/branch/referral/InstallListener.java
sojanpr: remove the Logs

aaustin:
are the logs for debugging purposes? Maybe I'm just a stickler but I'd rather not have them commented out if we're not going to use them again. It'd be better to just delete them.

:+1: your call

sojanpr:
Will remove the comments. If need for testing i think we can add them np

PR #119, opened 2015-08-14
Hot Fix for missing `setSubject()` method in Sharing builder

sojanpr:
Adding missing builder method to set subject for email/sms share

@aaustin @shortstuffsushi


Sarkar:
:+1:

PR #121, opened 2015-08-20
Internationalization Support Sharing SDK

sojanpr:
@aaustin @shortstuffsushi 
 Adding builder methods to Sharing sdks to support customization of
labels and icons

### DIFF
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -272,6 +272,12 @@ public class MainActivity extends Activity {
                             public void onShareLinkDialogLaunched() {
                                 Log.i("BranchTestBed", "onShareLinkDialogLaunched()");
                             }
+
+                            @Override
+                            public void onShareLinkDialogDismissed() {
+                                Log.i("BranchTestBed", "onShareLinkDialogDismissed()");
+                            }
+
                             @Override
                             public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchError error) {
                                 if (error != null) {
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -270,7 +270,7 @@ public class MainActivity extends Activity {
                         .setCallback(new Branch.BranchLinkShareListener() {
                             @Override
                             public void onShareLinkDialogLaunched() {
-                                Log.i("BranchTestBed", "onShareLinkDialogLaunched");
+                                Log.i("BranchTestBed", "onShareLinkDialogLaunched()");
                             }
                             @Override
                             public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchError error) {
@@ -286,6 +286,9 @@ public class MainActivity extends Activity {
                                 Log.i("BranchTestBed", "onChannelSelected... " + channelName);
                             }
                         })
+                        // Custom style for Copy url and More options
+                        //.setCopyUrlStyle(getResources().getDrawable(android.R.drawable.ic_menu_send),"Save this URl","Link added to clipboard")
+                        //.setMoreOptionStyle(getResources().getDrawable(android.R.drawable.ic_menu_search), "Show more")
                         .shareLink();
             }
         });
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -18,9 +18,7 @@ import java.util.Iterator;
 import io.branch.referral.Branch;
 import io.branch.referral.Branch.BranchReferralInitListener;
 import io.branch.referral.Branch.BranchReferralStateChangedListener;
-import io.branch.referral.BranchContentUrlBuilder;
 import io.branch.referral.BranchError;
-import io.branch.referral.BranchReferralUrlBuilder;
 import io.branch.referral.BranchShortLinkBuilder;
 import io.branch.referral.SharingHelper;
 
@@ -270,6 +268,10 @@ public class MainActivity extends Activity {
                         .addTag("Tag2")
                         .setDefaultURL("https://play.google.com/store/apps/details?id=com.kindred.android")
                         .setCallback(new Branch.BranchLinkShareListener() {
+                            @Override
+                            public void onShareLinkDialogLaunched() {
+                                Log.i("BranchTestBed", "onShareLinkDialogLaunched");
+                            }
                             @Override
                             public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchError error) {
                                 if (error != null) {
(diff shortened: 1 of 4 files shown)


Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
```diff
@@ -271,6 +269,10 @@ public void onClick(View view) {
                         .setDefaultURL("https://play.google.com/store/apps/details?id=com.kindred.android")
                         .setCallback(new Branch.BranchLinkShareListener() {
                             @Override
+                            public void onShareLinkDialogLaunched() {
```
shortstuffsushi: What is the value of this new function? What are devs going to do here?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2332,14 +2334,15 @@ public void applyReferralCode(final String code, final BranchReferralInitListene
      * user selected clients.</p>
      *
      * @param builder A {@link io.branch.referral.Branch.ShareLinkBuilder} instance to build share link.
+     * @return The instance of the {@link Dialog} holding the share view.
      */
-    private void shareLink(ShareLinkBuilder builder) {
+    private Dialog shareLink(ShareLinkBuilder builder) {
```
shortstuffsushi: Why do these functions now return the Dialog?

sojanpr: We have service request from partners that they want to handle dialog events directly. Also developer are so used with the standard  `AlertDialogBuilder`  which returns the Dialog instance from `create()` call in builder, Providing option for user to  handle dialog especially while context finishing.

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
sojanpr: Partners want to add a progress in the window between calling showShareDialog and the time at which dialog actually appears. This may take a second or 2  in some devices depending on the preferred option setting

Branch-SDK/src/io/branch/referral/Branch.java
shortstuffsushi: If we're returning it to them, we probably shouldn't be calling `show` on it. If we hand this control over to them, they probably want to manage that as well.

sojanpr: Hmm That would be ideal only worry is that existing applications also need  to change when upgrading the SDK.

shortstuffsushi: The way I see it, there are two ways this will go:
  1. We'll get support requests from new users who duplicate calls to `show` if we leave it.
  1. We'll get support requests from old users who don't know that the need to call `show` if we don't.

sojanpr: Unlink the `create` method in `alertBuilder` we are using `shareLink()` method to start sharing.That method already say it will create  chooser dialog for them. So I guess for the newer users we can survive with this. Also if user explicitly call `Show() ` when the dialog is up , there is nothing change

sojanpr: @shortstuffsushi  Another thinking on this is to keep the `shareLink()` method as it is now to return void. Then add a new method `createShareLinkDialog()` method to get the instance of the dialog is they want to Handle it by them self. How do you think about that?

shortstuffsushi: I'd prefer not to add two methods, that just seems likely to cause more problems. Can you explain again why we need to return the Dialog? If the dialog has already been displayed, they can't modify anything. What are they going to do with it?

sojanpr: OK The Idea was to give users full handle to the dialog. For e.g We are closing the dialog with an animation. Incase user want to to launch a new activity they want to dismiss the dialog immediately to prevent some  window leaks. With the  Dialog instance they can completely handle the situation.
    We can remove returning the dialog and provide `onDialogLaunched()` and `onDialogDismissed()` callbacks. Does that make sense?

shortstuffsushi: Yes, I believe that would be the better solution, as we may (now or later) need to have control over those callbacks ourselves, and not want users to be able to override their values. I think exposing those additional options is better. Also, aren't we already providing them with an `onShareLinkDialogLaunched` function?

sojanpr: Ok removed Dialog in return param and added callback for dialog launch and dismiss

sojanpr:
Ohh for got to remove that method. Removed it and replaced comparison with Math.max

Branch-SDK/src/io/branch/referral/AnimatedDialog.java
```diff
@@ -0,0 +1,148 @@
...
+    @Override
+    public void cancel() {
+        slideClose();
+    }
+
+    /**
+     * </p> Dismiss the dialog.</p>
+     */
+    @Override
+    public void dismiss() {
```
shortstuffsushi: If you're just calling the super here, probably don't need to override this method.

Branch-SDK/src/io/branch/referral/AnimatedDialog.java
```diff
@@ -0,0 +1,148 @@
...
+        ((ViewGroup) getWindow().getDecorView()).getChildAt(0).startAnimation(slideUp);
+
+        AnimatedDialog.super.show();
+    }
+
+    /**
+     * </p> Closes the dialog with a translation animation to the content view </p>
+     */
+    private void slideClose() {
+        if (!isClosing_) {
```
shortstuffsushi: I'm assuming you're using this to prevent race conditions, but you should do something more like

```
if (isClosing_) {
    return;
}

isClosing = true;
```

Because you have it in that callback, it's possible for slideClose to be called twice before the animation begins, which would still allow this to occur twice.

Branch-SDK/src/io/branch/referral/AnimatedDialog.java
```diff
@@ -0,0 +1,148 @@
...
+            slideDown.setAnimationListener(new Animation.AnimationListener() {
+                @Override
+                public void onAnimationStart(Animation animation) {
+                    isClosing_ = true;
+                }
+
+                @Override
+                public void onAnimationEnd(Animation animation) {
+                    isClosing_ = false;
+                    AnimatedDialog.super.dismiss();
```
shortstuffsushi: Why don't you just call dismiss?

Branch-SDK/src/io/branch/referral/AnimatedDialog.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * </p> Opens the dialog with a translation animation to the content view </p>
+     */
+    private void slideOpen() {
+        TranslateAnimation slideUp = new TranslateAnimation(Animation.RELATIVE_TO_SELF, 0, Animation.RELATIVE_TO_SELF, 0, Animation.RELATIVE_TO_SELF, 1.0f, Animation.RELATIVE_TO_SELF, 0f);
+        slideUp.setDuration(500);
+        slideUp.setInterpolator(new AccelerateInterpolator());
+        ((ViewGroup) getWindow().getDecorView()).getChildAt(0).startAnimation(slideUp);
+
+        AnimatedDialog.super.show();
```
shortstuffsushi: Why not just call `super.show()`?

Branch-SDK/src/io/branch/referral/AnimatedDialog.java
sojanpr: It was because we  used override the `Dismiss()` .Since we removed , will be changed  to  `dismiss`

shortstuffsushi:
Alright, I think this is good now :+1:

aaustin:
:+1: great

PR #127, opened 2015-09-04
Fixing Callback clearing issue 

sojanpr:
When Called internally form auto session it used to remove the callback
with null (if user already called intisession() from anywhere else )

@Sarkar @aaustin

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ServerRequestRegisterInstall.java
+++ b/Branch-SDK/src/io/branch/referral/ServerRequestRegisterInstall.java
@@ -3,6 +3,7 @@ package io.branch.referral;
 import android.app.Application;
 import android.content.Context;
 import android.util.DisplayMetrics;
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/ServerRequestRegisterInstall.java
```diff
@@ -3,6 +3,7 @@
 import android.app.Application;
 import android.content.Context;
 import android.util.DisplayMetrics;
+import android.util.Log;
```
Sarkar: don't think this import is needed

sojanpr: removed.

Sarkar:
looks good to me - tested as well :+1:

aaustin:
:+1:

PR #131, opened 2015-09-08
Removing request on Conflict

sojanpr:
@aaustin @qinweigong  Removing a request from the request queue on
getting 409( Resource conflict). Originally we never removed the
request from the queue and always retried. Seeing this issue ,feel like
it is better to remove the request after notifying through the callback.
https://github.com/BranchMetrics/Android-Deferred-Deep-Linking-SDK/issue
s/130

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -2879,7 +2879,7 @@ public class Branch {
                         if (thisReq_.isSessionInitRequest()) {
                             initState_ = SESSION_STATE.UNINITIALISED;
                         }
-                        //On a bad request continue processing
+                        // On a bad request notify with call back and remove the request.
                         if (status == 409) {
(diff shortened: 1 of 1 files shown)


Sarkar:
:+1:

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2887,6 +2887,7 @@ protected void onPostExecute(ServerResponse serverResponse) {
                                 Log.i("BranchSDK", "Branch API Error: Conflicting resource error code from API");
                                 handleFailure(0, status);
                             }
+                            requestQueue_.remove(thisReq_);
```
spennington: Would it be better to remove this before notifying the callback? The callback may want to queue a new request and if `thisReq_` is still at the head of the queue it may get repeated.

sojanpr: Thank you. Absolutely correct use case. Added changes to remove the request before calling the callback

PR #134, opened 2015-09-10
SYNC INIT

sojanpr:
Fixed issue with initSession Sync.
1.Enable synchronous init session and Init session time out settings
are added in the manifest. This variables cannot be passed with
initSession() call since in AutoSession session is started with out
user calling initSession. At that point Branch has no idea whether to execute sync/ async way
@aaustin @Sarkar @dmitrig01

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -2867,14 +2867,18 @@ public class Branch {
         protected ServerResponse doInBackground(Void... voids) {
             //Google ADs ID  and LAT value are updated using reflection. These method need background thread
             //So updating them for install and open on background thread.
-            if(thisReq_.isSessionInitRequest()){
-                thisReq_.updateGAdsParams(systemObserver_);
-            }
+//            if(thisReq_.isSessionInitRequest()){
+//                thisReq_.updateGAdsParams(systemObserver_);
+//            }
             if (thisReq_.isGetRequest()) {
                 return kRemoteInterface_.make_restful_get(thisReq_.getRequestUrl(), thisReq_.getRequestPath(), timeOut_);
+
             } else {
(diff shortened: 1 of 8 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -847,7 +875,7 @@ private void initUserSessionInternal(BranchReferralInitListener callback, Activi
         //If already initialised
         if (hasUser() && hasSession() && initState_ == SESSION_STATE.INITIALISED) {
             if (callback != null)
-                callback.onInitFinished(new JSONObject(), null);
+                callback.onInitFinished(getLatestReferringParams(), null);
```
aaustin: Hmm - this is a behavior change. The problem being that if developers call initSession when the state is already initialized, we didn't want to return the parameters over and over again. This is an issue when the developer uses initSession with callback in multiple Activities (which was common before auto session management). I can see how this would happen now with auto session management.

Can you check if auto session management, return getLatest, else, return new.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2865,106 +2919,117 @@ protected ServerResponse doInBackground(Void... voids) {
...
-                            networkCount_ = 0;
-
-                            //Finally call the request callback with the error.
-                            for (ServerRequest req : requestToFail) {
-                                req.handleFailure(status);
-                                //If request need to be replayed, no need for the callbacks
-                                if (req.shouldRetryOnFail())
-                                    req.clearCallbacks();
-                            }
+    private void processSeverResponse(ServerRequest serverRequest, ServerResponse serverResponse) {
```
aaustin: please correct spelling to 'Server'

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Added fix  fror handling it differently for nonAutoSession

PR #135, opened 2015-09-10
Clearing session param on Closing session.

sojanpr:
On closing session clears the session params. (It used to be cleared
only when next open succeeded. So if init fails app wilt last session
params)

@aaustin @Sarkar


aaustin:
Looks great :+1:

Sarkar:
nice catch! :+1:

PR #136, opened 2015-09-10
Android M Support2

sojanpr:
Support for Android M . No permission changes needed since branch is
using only internet permission and is  a Default permission in Android
M
@aaustin


SleeplessByte:
Please install -23 on circleci (through ci script) so the check doesn't fail!

sojanpr:
Changes merged with PR https://github.com/BranchMetrics/Android-Deferred-Deep-Linking-SDK/pull/143

PR #137, opened 2015-09-14
Fixing Prams going empty on Sync Init

sojanpr:
Since Auto session mode initialize the session by itself on starting
the first activity, we need to provide user.

The referring params if they call init session after init is completed.
Note that user wont do InitSession per activity in auto session mode.

@aaustin @Sarkar @dmitrig01


Sarkar:
:+1: just confirmed flow with Sojan.

aaustin:
But Pinterest is using the countdownlatch - where does that boolean get set?

PR #142, opened 2015-09-28
Potential NPE case fix

sojanpr:
Adding Null check  for  possible NPE Reported over crashlytics and
some other unit tests

@aaustin @Sarkar


aaustin:
:+1:

PR #144, opened 2015-10-01
Fixing issue :  Same params transmitted on all init calls

sojanpr:
IN Auto session management we were returning latest params with all
init call if the Branch is already initialized. This is due to the
reason that Init session is caused by Branch SDK in case of Auto
session and by the time user start an init session , session might be
already finished.

Added a fix to check if the init Params are  called back at lease once.
only affect the auto session

@aaustin @Sarkar


aaustin:
Looks great - sorry I didn't notice this on the previous PR :+1:

PR #145, opened 2015-10-02
Fixing issue : creating link before session init

sojanpr:
Fixing Issue from Partner
```There is a scenario in our partner's app in which they end up calling BranchShortLinkBuilder.getShortLink() before they get the BranchReferralInitListener.onInitFinished callback. In this case, the short-link generation fails.```

Please note that only Asynchronous  Link creation call ( generateShortURl()) will  only be served if called before Init session. getShortUrl() is a synchronous call and this will return null immediately if called before initialised.

@aaustin @derrickstaten @Sarkar

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -921,7 +921,7 @@ public class Branch {
         }
         //If uninitialised or initialising
         else {
-            // In case of Auto session init will be called from banch before user. So initialising
+            // In case of Auto session init will be called from Branch before user. So initialising
             // State also need to look for isReferrable value
             if (isReferrable) {
                 this.prefHelper_.setIsReferrable();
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Branch-SDK/src/io/branch/referral/Branch.java [hunk 1]
```diff
@@ -2719,9 +2719,9 @@ private void handleNewRequest(ServerRequest req) {
             if((req instanceof ServerRequestRegisterClose)){
                 Log.i(TAG, "Branch is not initialized, cannot close session");
                 return;
-            }
-            else{
-                initializeSession(null);
+            } else {
+                boolean isReferrable = prefHelper_.getIsReferrable() == 1;
```
aaustin: Will this adhere to the policy of keeping isReferrable true unless overridden?

sojanpr: This case will occur when user calling a Branch API (other than open/install) when Branch is not initialized. At this point we have no idea what user want.
- In case app init session and it failed , this will use the user specified value. 
- in case app is trying to call any API before init  app will use last known value
- On a fresh install and API is called before init , this will use default value which is `non Referrable`

aaustin: I thought we decided that is_referrable is true by default.

sojanpr: Ok  That was a logic error. Changed to make referrable unless explicitly specified by user

aaustin:
:+1:

PR #149, opened 2015-10-14
Branch Universal Object

sojanpr:
First cut implementation for BranchUniversalObject


sojanpr:
First cut for BranchUniversalObject
@aaustin @derrickstaten

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
```diff
@@ -244,60 +237,56 @@ public void onClick(View v) {
...
+                    public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchError error) {
+                    }
+                    @Override
+                    public void onChannelSelected(String channelName) {
+                    }
+                });
             }
         });
+
+        branchUniversalObject = new BranchUniversalObject("canonical/identifier/", "My Content Title")
```
aaustin: wouldn't you want to specify this stuff before calling the share sheet?

aaustin:
First comment: Instead of deleting the old methods (which will cause thousands of voices to cry out in anguish on upgrade) can you add a deprecation notice with indication to review the new mechanism? You can link to 'https://dev.branch.io/recipes/content_sharing/android/' for an example.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2765,6 +2766,14 @@ public void onActivityStarted(Activity activity) {
                 initSessionWithData(intentData, activity); // indicate  starting of session.
             }
             activityCnt_++;
+
+            // Activity stack for finding the structured path to this activity
+            if (activityStack_.size() > 1 && (activityStack_.get(activityStack_.size() - 2)).equals(activity.getClass().getSimpleName())) {
+                activityStack_.remove(activityStack_.get(activityStack_.size() - 2));
+                activityStack_.remove(activityStack_.get(activityStack_.size() - 1));
```
aaustin: Is there any risk that this gets called from two threads at once? Or is it only called on the main thread?

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
```diff
@@ -62,6 +63,11 @@ public ServerRequestCreateUrl(Context context, final String alias, final int typ
 
         linkPost_ = new BranchLinkData();
         try {
+            if (canonicalID != null && canonicalID.length() > 0) {
+                JSONObject paramsJson = new JSONObject(params);
```
aaustin: Aren't you assembling the params in the Universal Object? Why do you need to insert this here?

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
sojanpr: Yes .  The BranchUniversalObject is created before calling the share sheet. (`BranchUniversalObject` is created on `onCreate()` and the Share sheet is called on a button click event.  Probably I should move to to the first statement in `onCreate()` to avoid a confusion

Branch-SDK/src/io/branch/referral/ServerRequestRegisterView.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void clearCallbacks() {
+        callback_ = null;
+    }
+
+    @Override
+    public String getRequestUrl() {
+        return "http://54.153.121.111:5000/v0.1/register_view/";
```
aaustin: do we have the real API yet?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Since this is a System UI event it is always called in only in UI thread. This method is adopted for finding the the path to a current Activity (e.g.`home/recipe-list/recipe-12345`). Since we are not using this now we can remove this

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
sojanpr: Removed the duplicate setting for Canonical ID

Branch-SDK/src/io/branch/referral/ServerRequestRegisterView.java
sojanpr: Added production Api end point

sojanpr:
This PR is moved to https://github.com/BranchMetrics/Android-Deferred-Deep-Linking-SDK/pull/152

PR #150, opened 2015-10-14
Fix for issue Resource conflict  error code is not added to server response

sojanpr:
In case of Resource conflict `(Status =409)` getInputStream will throw
FileNotFoundException. Handling it here in order to send the right
status code

@aaustin @Sarkar


Sarkar:
good try/catch. :+1:

aaustin:
:+1:

PR #154, opened 2015-10-20
Fixing a activity reference leak on Manual session

sojanpr:
On manual session there were couple of possible memory leak
1.`contenxt_`  is assigned with activity instance. This will keep the
first activity reference leaked .Changed to `context_ ` to save only
application context.

2. `currentActivity_` is saving the activity instance on each
initSessionCall() but up on calling  `closeSession` last activity
instance is  not freed up .

@aaustin @Sarkar @EvangelosG


aaustin:
:+1: looks good

EvangelosG:
@sojanpr let's test this with android studio's memory monitor when I get into the office today.

Sarkar:
:+1:

PR #160, opened 2015-11-03
Cleanup And Reorg

sojanpr:
PR for Clean Up and and some missing functionalities to the SDK
@aaustin @Sarkar @EvangelosG


aaustin:
Looks like it's coming together nicely! Were there other items left on the backlog?

aaustin:
As far as I can tell :+1: But if you add more here, let me know.

sojanpr:
@Sarkar I tried with true but due to some AS lint issue it is aborting the makeJar. That why we ignore the Lint

Sarkar:
makes sense. this looks good to me :+1:

aaustin:
Looks good to me :+1:

PR #164, opened 2015-11-13
Adding an error on on Share sheet if Branch instance is not created

sojanpr:
Very rare case :- In case any one one trying to access share sheet
functionality using  BUO  when Branch instance is not created (In case
forget to instantiate branch or missed BranchApp in manifest). This
will show an error log / callback with appropriate message instead of
an exception.

@aaustin  @Sarkar


Sarkar:
:+1:

aaustin:
:+1:

PR #165, opened 2015-11-13
Fixing a n intermittent crash issue

sojanpr:
Making sure the  context is live until the share dialog is dismissed.
Release  the context only on dismissing Share dialog

@Sarkar @aaustin


Sarkar:
:+1:

aaustin:
You ran leak canary on this? I thought we added those nulls for a leak before.

sojanpr:
Yes I have ran Leak Canary and couldn't  find leak. The null is essentially added to prevent the leak, but we are setting the `context_` to null on dismiss of the dialog. So when ever the dialog is dismissed we are releasing the context

aaustin:
:+1:

PR #166, opened 2015-11-21
Extension Support for Branch Referral SDK

sojanpr:
1) Opening necessary end points to aces features of core SDK to
extending SDKs.

2) Adding option to set parameters to get call

@aaustin

### DIFF
--- a/Branch-SDK/src/io/branch/referral/RemoteInterface.java
+++ b/Branch-SDK/src/io/branch/referral/RemoteInterface.java
@@ -104,17 +104,18 @@ class RemoteInterface {
     }
 
     /**
-     * <p>Make a RESTful GET request, by calling {@link #make_restful_get(String, String, int, int, boolean)}
+     * <p>Make a RESTful GET request, by calling {@link #make_restful_get(String, JSONObject, String, int, int, boolean)}
      * with the logging {@link Boolean} parameter pre-populated.</p>
      *
      * @param url     A {@link String} URL to request from.
+     * @param params  A {@link JSONObject} with additional parameters and their values to send with the get request.
      * @param tag     A {@link String} tag for logging/analytics purposes.
      * @param timeout An {@link Integer} value containing the number of milliseconds to wait
      *                before considering a server request to have timed out.
      * @return A {@link ServerResponse} object containing the result of the RESTful request.
      */
-    public ServerResponse make_restful_get(String url, String tag, int timeout) {
-        return make_restful_get(url, tag, timeout, 0, true);
+    public ServerResponse make_restful_get(String url, JSONObject params, String tag, int timeout) {
+        return make_restful_get(url, params, tag, timeout, 0, true);
     }
 
     private boolean addCommonParams(JSONObject post, int retryNumber) {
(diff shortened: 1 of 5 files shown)


Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -156,6 +158,16 @@ private ServerResponse make_restful_get(String baseUrl, String tag, int timeout,
             timeout = DEFAULT_TIMEOUT;
         }
         if (addCommonParams(getParameters, retryNumber)) {
+            if (params != null) {
```
aaustin: Why add this in here and not in the ServerRequest getRequestPath of the subclass for the request? That's what is done with other GET requests.

aaustin:
This seems excessive. Please explain why we can't use the existing GET structure and need to introduce a new one.

sojanpr:
We have no Get APIs which take custom parameters.(IdentityID is added to the get command as part of path) .We add the common parameters (Such as Branch Key,Session ID etc) in `addCommonPrams() `  method in `RemoteChannel` .  
  To avoid the complexity of checking for Except params and adding params (checking for "?" and Appending with `&`) , seems it is better to add prams in one shot.

Branch-SDK/src/io/branch/referral/ServerRequest.java
```diff
@@ -22,6 +22,8 @@
     /*True if there is an error in creating this request such as error with json parameters.*/
     public boolean constructError_ = false;
 
+    private final JSONObject getParamObject;
```
aaustin: Do we need to split out getParams from normal parameters on a request? If the request is type GET, can you just use the other JSONObject?

Branch-SDK/src/io/branch/referral/RemoteInterface.java
aaustin: After discussing, we've decided this is the correct implementation

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -156,6 +158,16 @@ private ServerResponse make_restful_get(String baseUrl, String tag, int timeout,
...
+                Iterator keys = params.keys();
+                while (keys.hasNext()) {
```
aaustin: Is there a risk that the request will be modified while iterating? we had a nasty bug for a while because we did iteration here and still allowed the server requests to be modified.

Branch-SDK/src/io/branch/referral/ServerRequest.java
sojanpr: Removed the extra JSON object and `param_` variable stores get /post params now

Branch-SDK/src/io/branch/referral/RemoteInterface.java
sojanpr: Good catch.  Converted `addGetParams` in the `ServerRequest` class is converted  to `protected`. So only get params can be modifed inside the class .

aaustin:
Looks good :+1:

PR #167, opened 2015-11-24
Server API Retry

sojanpr:
1) Updated Request time out to 5.5.sec and retry count to 3

2) Added to `lrtt` (last round trip time in milli sec ) to the request
parameters.

@aaustin


aaustin:
Can we change the client timeout messaging to something unique, like "Client timed out before response sent from Branch"? Other than that :+1:

sojanpr:
Added a custom message  for Request time out

PR #169, opened 2015-11-24
Adding a broadcaster for session state

sojanpr:
Broadcaster for init session
@aaustin @Sarkar @danwalkowski

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -367,6 +367,9 @@ public class Branch {
     /* Sets to true when the init session params are reported to the app though call back.*/
     private boolean isInitReportedThroughCallBack = false;
 
+    /* List of session state listeners registered */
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3375,10 +3379,12 @@ else if (thisReq_ instanceof ServerRequestLogout) {
                                     thisReq_.onRequestSucceeded(serverResponse, branchReferral_);
                                 }
                             }
+                            BroadcastSessionState(initState_);
```
Sarkar: Wondering if this should perhaps be moved up into the `if (thisReq_ instanceof ServerRequestInitSession)` block - since that's the only time initState is updated to initialized? just a thought, would want to prevent excess broadcasts

Sarkar:
looks like a good design :+1:

aaustin:
:+1:

PR #173, opened 2015-12-01
Fixing Issue GAID fetch hang

sojanpr:
On Nexus 9 and Nexus 10 , some edge cases can lead to resource conflict
and dead lock while accessing GAID on background thread. Adding a timer
thread to access the GAID sms to solve this issue.
Issue is seen only with one specific partner and this solution seems
working fine with them.

@aaustin @Sarkar @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ServerRequest.java
+++ b/Branch-SDK/src/io/branch/referral/ServerRequest.java
@@ -7,6 +7,9 @@ import android.content.pm.PackageManager;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
 /**
  * Abstract class defining the structure of a Branch Server request.
  */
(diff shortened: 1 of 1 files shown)


aaustin:
looks good :+1:

Branch-SDK/src/io/branch/referral/ServerRequest.java
```diff
@@ -262,25 +265,43 @@ private static ServerRequest getExtendedServerRequest(String requestPath, JSONOb
...
-            String advertisingId = sysObserver.getAdvertisingId();
-            if (advertisingId != null && getPost() != null) {
-                getPost().put(Defines.Jsonkey.GoogleAdvertisingID.getKey(), advertisingId);
-            }
-            int latVal = sysObserver.getLATValue();
-            if (getPost() != null) {
-                getPost().put(Defines.Jsonkey.LATVal.getKey(), latVal);
+    public void updateGAdsParams(final SystemObserver sysObserver) {
+        final CountDownLatch  latch  = new CountDownLatch(1);
+        skipOnTimeOut = false;
```
Sarkar: hm - you have a skipOnTimeOut up there which is default set to false, do we need it again here?

Sarkar:
:+1:

Branch-SDK/src/io/branch/referral/ServerRequest.java
sojanpr: This is to counter the case of calling "updateGAdsParams()" multiple time on same server request (Even though we  don't do it now)

PR #174, opened 2015-12-02
Add timings to server requests

dmitrig01:
As discussed with Sahil:

In order to gain a better sense of how long things are actually taking on the client side, I'd like to add the ability to be able to measure times on the client side, and send them back to the server. This shouldn't get in the way of any existing functionality, so it's a best-effort at most.

The basic model is, we have a timer repository where you can associate a timer id with a time value; then on every request, we will retrieve the currently stored timings, and send them up to the server, where we can do some kind of analysis.

The current things I have timed are: (1) time between when a ServerRequest object is created and when the request is actually started (hypothesis: it's possible the request is sitting on the AsyncTaskQueue for a while), and (2) time between request start and request end (we have server-side latencies, but it'd be good to get data on what the client-side latencies are).

Obviously, we'll only be able to collect request time data after a request is finished. We aren't going to make a separate request to send this timing data; instead, we'll just attach it to whatever request happens to go out next (and if there's not request... well, we're doing best-effort here). Note that we'll still be able to know *what* was timed, because it was associated with a specific timer id (e.g., the "/v1/open request time" timer id).

Additionally, I've added the capability for users to be able to add their own timers. In most cases, this won't be necessary, but if we are doing a deep integration, it could be useful to instrument on our end. I've reserved the first 1000 timer ids for us, and allowed partners to specify ids outside of that. It's possible we'll want to comment this out for public release.

Untested - not sure how...

@Sarkar @sojanpr

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -367,6 +367,9 @@ public class Branch {
     /* Sets to true when the init session params are reported to the app though call back.*/
     private boolean isInitReportedThroughCallBack = false;
 
+    /* Key added to all post request to denote timings */
+    private static final String KEY_TIMINGS = "timings";
+
     /**
      * <p>The main constructor of the Branch class is private because the class uses the Singleton
      * pattern.</p>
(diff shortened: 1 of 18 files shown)


dmitrig01:
closing in favor of https://github.com/BranchMetrics/Android-Deferred-Deep-Linking-SDK/pull/175

PR #179, opened 2015-12-04
Caturing external URI and Extra always

sojanpr:
Earlier we used to capture this only if link-click-is is not present in
the intent URI. Capturing this always will needed considering the case
of google app search or other search app launching the app with
deplaned content.

 The Info is added with Open/Install post with following format
```”external_intent_uri": "branchtest:\/\/open",
   "external_intent_extra": "{\"android.intent.extra.REFERRER_NAME\":\"android-app:\\\/\\\/com.google.android.googlequicksearchbox\\\/https\\\/www.google.com\”}”,```

@aaustin @derrickstaten @dmitrig01


aaustin:
:+1:

sojanpr:
merged

PR #183, opened 2015-12-07
Push Notification Support

sojanpr:
App will check for extra value for key  “branch” in intent. If
available this value is  send with the install/open api as  `
"push_identifier": “my_branch_url”,`
Clients need  to add the branch url as “branch” in their extra
values for push notification

@aaustin @Sarkar @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -1289,6 +1289,28 @@ public class Branch {
     }
 
     private boolean readAndStripParam(Uri data, Activity activity) {
+
+        // Capture the intent URI and extra for analytics in case started by external intents such as  google app search
+        try {
+            if (data != null) {
+                prefHelper_.setExternalIntentUri(data.toString());
+            }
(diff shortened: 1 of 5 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1289,6 +1289,37 @@ private void executeClose() {
     }
 
     private boolean readAndStripParam(Uri data, Activity activity) {
+
+        // Capture the intent URI and extra for analytics in case started by external intents such as  google app search
+        try {
```
aaustin: this is the change from the other PR merged in, correct?

aaustin:
If you confirm the above :+1:

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Yes this Branch is cut from the branch for "Capture_External_URI" fix Branch

sojanpr:
Merged

PR #184, opened 2015-12-07
Adding 100% deplane debug support

sojanpr:
Adding API setDeepLinkDebugMode() to set the deep link debug
parameters. These parameters are returned back 100 % time with init
session callbacks

Note: This method should be called from onCreate() of activity which
listens to Branch Init Session

@aaustin @Sarkar

### DIFF
--- a/Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
+++ b/Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
@@ -486,7 +486,12 @@ public class BranchUniversalObject implements Parcelable {
         Branch branchInstance = Branch.getInstance();
         try {
             if (branchInstance != null && branchInstance.getLatestReferringParams() != null) {
+                // Check if link clicked. Unless deepvlink debug enabled return null if there is no link click
                 if (branchInstance.getLatestReferringParams().has("+clicked_branch_link") && branchInstance.getLatestReferringParams().getBoolean("+clicked_branch_link")) {
+                    branchUniversalObject = createInstance(branchInstance.getLatestReferringParams());
+                }
+                // If debug params are set then send BUO object even if link click is false
+                else if (branchInstance.getDeeplinkDebugParams() != null && branchInstance.getDeeplinkDebugParams().length() > 0) {
                     branchUniversalObject = createInstance(branchInstance.getLatestReferringParams());
                 }
             }
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -285,6 +285,9 @@ public class Branch {
      */
     private static final int PREVENT_CLOSE_TIMEOUT = 500;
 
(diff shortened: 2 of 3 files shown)


Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
```diff
@@ -486,9 +486,14 @@ public static BranchUniversalObject getReferredBranchUniversalObject() {
         Branch branchInstance = Branch.getInstance();
         try {
             if (branchInstance != null && branchInstance.getLatestReferringParams() != null) {
+                // Check if link clicked. Unless deepvlink debug enabled return null if there is no link click
                 if (branchInstance.getLatestReferringParams().has("+clicked_branch_link") && branchInstance.getLatestReferringParams().getBoolean("+clicked_branch_link")) {
                     branchUniversalObject = createInstance(branchInstance.getLatestReferringParams());
                 }
+                // If debug params are set then send BUO object even if link click is false
+                else if (branchInstance.getDeeplinkDebugParams() != null && branchInstance.getDeeplinkDebugParams().length() > 0) {
```
Sarkar: I think there should be a flag to toggle this mode, since I'm worried about the case where the above line is false `(+clicked_branch_link == 0)` and someone ships an .apk without commenting out: 

`Branch.getInstance().setDeepLinkDebugMode(JSONObject ...)`

Thoughts?

Sarkar:
lmk what you think about my comment, otherwise :+1:

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
aaustin: I see it similar to setDebug. We'll want to through some loud console errors when they use it so it's not forgotten.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1717,7 +1730,33 @@ public JSONObject getFirstReferringParams() {
      */
     public JSONObject getLatestReferringParams() {
         String storedParam = prefHelper_.getSessionParams();
-        return convertParamsStringToDictionary(storedParam);
+        JSONObject latestParams = convertParamsStringToDictionary(storedParam);
```
aaustin: Let's put a loud warning here that says. "You're currently in deep link debug mode. Please comment out 'setDeepLinkDebugMode' to receive the deep link parameters from a real Branch link". Please put this in the above function as well.

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
Sarkar: got it - since we don't have setDebug here, it can be another key entry in the Manifest?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Added a debug deeplink warning

PR #189, opened 2015-12-15
Support for non Encoded referrer

sojanpr:
Support for Back end Change:- Removing encoded "=" in the referrer for
adding link-click-id. Link click id is added in the format
"link_click_id-ident". This to fix the issue with few Android version
where play store page is not correctly displayed when "=" is encoded to
%3D

@aaustin @derrickstaten @dmitrig01


aaustin:
:+1:

PR #193, opened 2016-01-01
Fixing crash on concurrent exeption

sojanpr:
Preventing chance for concurrent modification exception and  handling
properly in case of exception to prevent crash.

@aaustin


aaustin:
Ahh nice :+1:

aaustin:
This is clean. Nice :+1:

PR #194, opened 2016-01-04
Fixing init issue with app links

sojanpr:
Marking the app link in the intent as used once it is used for getting
deep link params in order to prevent providing wrong deep link data.

@aaustin


sojanpr:
Tested and Review ready

aaustin:
Does this still require onNewIntent and `singleTask` launch mode?

sojanpr:
Unfortunately `onNewIntent` and  singleTask is still needed. With out which the case  of clicking a link when app is in background will not work properly

aaustin:
:+1: looks great

PR #195, opened 2016-01-05
Adding Styling option for list view

sojanpr:
Since share sheet is  a custom  component minimal styling is added to
the sheet. Builder now provides option to add custom styles.

@aaustin


sojanpr:
Tested and Review reay

aaustin:
Looks good - can you add an example to the testbed so that others can see how this is used?

sojanpr:
Example added for setting custom style

aaustin:
:+1:

PR #196, opened 2016-01-07
safeguard against NPE if getNetworkInfo invalid

Sarkar:
@sojanpr @aaustin 

if someone specifies <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE "/>, we determine if wifi is connected upon a new user install.

In some cases, the class that retrieves this network property can be invalid, which returns null: http://developer.android.com/reference/android/net/ConnectivityManager.html#getNetworkInfo(android.net.Network)

we don't check for null before returning, leading to a crash: http://crashes.to/s/40d5b581afa


sojanpr:
Very nice ! LGTM
(I think null is retuned on devices which does not have Wifi adapter)

sojanpr:
:+1:

PR #197, opened 2016-01-09
Fixed compatibility with Android 6.0

Evgeniy-Sinyak:
I have recently faced a bug in my app that an ActionMode adds an additional Action Bar instead of using the existing one. This bug is reproducible only on Android 6.0. Resource into the issue shows that this happens because BranchWindowCallback doesn't implement onWindowStartingActionMode(ActionMode.Callback ,int) which was added in API level 23


sojanpr:
@Evgeniy-Sinyak Thank you very much for this PR. We are testing to reproduce this crash.
Could you please  help with some input to see how the Action mode create this issue
? Really curious since our test bed app with SDK version 23 works fine with Branch lib

Evgeniy-Sinyak:
https://drive.google.com/file/d/0B57Ez5L61GnHZThYZjJMbW1iOEU/view?usp=sharing

@sojanpr Here is small project which demonstrates the issue described above. This bug reproduces on devices with Android 6.0 or above. Before run it, please, update "io.branch.sdk.BranchKey" and "io.branch.sdk.BranchKey.test" in the manifest file.

Also I have made two screenshots with BranchIO enabled and than commented it, so take a look on wrong_result.png and expected_result.png in the root of the archive.

sojanpr:
@Evgeniy-Sinyak Really appreciate the the details and sample app. Will check it out and  verify against this PR . Update you asap

sojanpr:
@Evgeniy-Sinyak  With recent changes we have removed the Window Callback. So this should not be a any issue any more but we will update SDK with 23 after testing

Sarkar:
@Evgeniy-Sinyak thanks for this - if this is still occurring, let us know, otherwise you should be good to go with our latest - as we've opted to remove everything related to how we use the Window + callbacks. 

closing now unless anything else is needed. thanks for your help!

PR #201, opened 2016-01-18
Format Change for Long link

sojanpr:
long link format changed to  `[link base] + & + [query parameters]`
since link already contains ? now.

@dmitrig01  @aaustin


dmitrig01:
hm, actually on second thought - it looks like sometimes it generates an /a/ link.

instead, can we just check if the URL has a ?, if so add &, if not add ?

aaustin:
agreed

PR #209, opened 2016-02-10
Setting for  updating Gads param on request

sojanpr:
Provides a convenient option on Server request to specify whether it
need GAds params with the request

@aaustin @Sarkar


aaustin:
:+1: Looks good

Sarkar:
nice one :+1:

PR #212, opened 2016-02-18
Fixing link data match ussie

sojanpr:
Fixing issue with link-data match checking for async short url creation.
Safe handling for non url encoded characters in Install referrer string
@Sarkar @EvangelosG @aaustin


Sarkar:
for async url creation - this retrieves the contents of a created link from the request object - which is stored in the cache from a successful response

:+1:

aaustin:
Was concerned that adding more keys to the cache would cause the cache to work but it looks like we've implemented a custom equals method to determine if link data is the same. :+1:

EvangelosG:
Thanks for adding the extra exception handling, partner notified.

PR #215, opened 2016-02-25
Removing strong references to to Context

sojanpr:
Strong reference to the context variable may create mem leak depending
on how activities are handled in the app implementation.

@aaustin @Sarkar

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
+++ b/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
@@ -22,7 +22,6 @@ import android.widget.TextView;
 import android.widget.Toast;
 
 import java.lang.ref.WeakReference;
-import java.nio.channels.NonWritableChannelException;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -43,7 +42,7 @@ class ShareLinkManager {
     /* Background color for the list view in disabled state. */
     private final int BG_COLOR_DISABLED = Color.argb(20, 17, 4, 56);
     /* Current activity context.*/
-    private WeakReference <Activity> context_;
+    private Activity context_;
     /* Default height for the list item.*/
     private static int viewItemMinHeight_ = 100;
     /* Indicates whether a sharing is in progress*/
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/ShareLinkManager.java
```diff
@@ -319,31 +322,31 @@ public boolean isEnabled(int position) {
...
-            this.setMinWidth(context_.getResources().getDisplayMetrics().widthPixels);
+            this.setMinWidth(context_.get().getResources().getDisplayMetrics().widthPixels);
         }
 
         public void setLabel(String appName, Drawable appIcon, boolean isEnabled) {
             this.setText("\t" + appName);
             this.setTag(appName);
             if (appIcon == null) {
-                this.setTextAppearance(context_, android.R.style.TextAppearance_Large);
+                this.setTextAppearance(context_.get(), android.R.style.TextAppearance_Large);
```
aaustin: for these 3, should we check if not null before trying to use it?

aaustin: also, they aren't using the ShareLinkBuilder, so I'd say we shouldn't over correct this class. Thoughts?

sojanpr: Agree. Will Add the null check before getting anything using weak reference

Sarkar: i think ShareLinkBuilder had a memory leak before, so it's better to account for this especially if we're making Activity reachable by weak reference in Branch.java anyway

Sarkar:
Nice tactic using WeakReference - I think this is a good way to not leak on Activity :+1:

Branch-SDK/src/io/branch/referral/ShareLinkManager.java
sojanpr: Figured out how exactly the leak happens with ShareLinkManager and added  fix for the issue. Removed previous corrections with weak reference for the share link manager

sojanpr:
Moved to  PR https://github.com/BranchMetrics/Android-Deferred-Deep-Linking-SDK/pull/216

PR #216, opened 2016-02-25
Leak Protection 2

sojanpr:
1. Changed activity context reference to a weak reference

2. Fixed leak issue with share link manager

@aaustin @Sarkar

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
+++ b/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
@@ -228,6 +228,8 @@ class ShareLinkManager {
                     }
                 }
                 isShareInProgress_ = false;
+                // Cancel the dialog and context is released on dialog cancel
+                cancelShareLinkDialog(false);
             }
         });
     }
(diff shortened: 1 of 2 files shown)


aaustin:
Looks good :+1:

Branch-SDK/src/io/branch/referral/ShareLinkManager.java
```diff
@@ -228,6 +228,8 @@ public void onLinkCreate(String url, BranchError error) {
                     }
                 }
                 isShareInProgress_ = false;
+                // Cancel the dialog and context is released on dialog cancel
+                cancelShareLinkDialog(false);
```
Sarkar: was this added back in?

sojanpr: It was not there. The reason for adding it is 
1) on launching a sharing end point a new activity comes to the top of the stack and GC will try to recycle the current activity.
2) We cannot set `context =null` since the dialog termination is asynchrounos. if it is not closed when user press back, the share dialog will try to set the visible rows again and will result in crash.

Sarkar: works for me!

Sarkar:
:+1:

PR #218, opened 2016-02-26
NPE fix for currentActivityReference

sojanpr:
currentActivityReference is initialized only on Activity starts or when
init session called internally. This crash is caused by application
calling branch feature such as “userActionComplete” before initsession
happens

@aaustin @Sarkar


aaustin:
:+1:

Sarkar:
:+1:

PR #221, opened 2016-03-02
Adding BUO example for deep link router

sojanpr:
Deep link router example was using legacy deep link routing example.
Updating with BUO example.

@aaustin @Sarkar

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,12 +172,17 @@ If you are not creating or using an Application class throughout your project, a
 
 ```java
 Branch branch = Branch.getInstance(getApplicationContext());
-branch.initSession(new BranchReferralInitListener(){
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -172,12 +172,17 @@ If you are not creating or using an Application class throughout your project, a
 
 ```java
 Branch branch = Branch.getInstance(getApplicationContext());
-branch.initSession(new BranchReferralInitListener(){
+branch.initSession(new Branch.BranchUniversalReferralInitListener() {
```
aaustin: Because the BUO is not the only way to create links, we can't make this the default unfortunately : / You can add this as an option below the main JSONObject version, as we do in a couple other repos.

PR #223, opened 2016-03-23
Fixing corner case  latest params reset

sojanpr:
This is a really corner case but prevalent when app is killed by user
or process.
Here is the steps how it happens

1. On a slower network user press home button and immediately  kill the
app.

2. A close request is enqueued and not responded

3. Opening app next time queue is retrieved and so open request  is
immediately followed by close and wipe off the session params

Testing : Tested by adding delay to close req execute . Pressing recent app and removing app will push one close request to queue which is retrieved on app init

@aaustin @Sarkar

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
+++ b/Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
@@ -16,9 +16,9 @@ import java.util.List;
 import java.util.NoSuchElementException;
 
 /**
- *<p>The Branch SDK can queue up requests whilst it is waiting for initialization of a session to
- *complete. This allows you to start sending requests to the Branch API as soon as your app is
- *opened.</p>
+ * <p>The Branch SDK can queue up requests whilst it is waiting for initialization of a session to
+ * complete. This allows you to start sending requests to the Branch API as soon as your app is
+ * opened.</p>
  */
 class ServerRequestQueue {
     private static final String PREF_KEY = "BNCServerRequestQueue";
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
```diff
@@ -105,7 +104,10 @@ public void run() {
                     JSONObject json = jsonArr.getJSONObject(i);
                     ServerRequest req = ServerRequest.fromJSON(json, context);
                     if (req != null) {
-                        result.add(req);
+                        // No need to retrieve close or logout from previous session
+                        if (!(req instanceof ServerRequestRegisterClose || req instanceof ServerRequestLogout)) {
```
Sarkar: quick q here. in the case of a logout not being queued up to be fired, we'd still use an old identity ID inside the following v1/open right?

Sarkar:
code makes sense to me, just the clarifying q on v1/logout :+1:

Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
sojanpr: Yes open will fire with the previous identity ID. Identity ID will rest only on a successful logout

aaustin:
Looks good, as long as retrieve on only fetched on startup, which it appears that it is :+1:

sojanpr:
Yes retrieve is called only on creating Branch singleton instance

PR #228, opened 2016-03-30
Adding NPE protection

sojanpr:
There seems like a corner case were some partners see
activity.getIntent().getDataString() is null. This may be caused by
application setting the  intent data to null before Branch read it.
Adding a safe check here

@aaustin @Sarkar @EvangelosG


Sarkar:
solid :+1:

EvangelosG:
always good to check for null :) :+1:

PR #230, opened 2016-04-02
first commit of custom message and title based on channel

EvangelosG:
@sojanpr @aaustin @Sarkar first attempt at supporting custom share title/message based on selected channel. will need a better name for new interface, other than BranchLinkShareListener2. Also, I'm not 100% sure if everything was done correctly, looking forward to reviewing w/ you guys. TestBed from this branch utilizes new interface, if you want to see how it works, look @ MainActivity.

### DIFF
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -259,7 +259,7 @@ public class MainActivity extends Activity {
                 // Define custom styel for the share sheet list view
                 //.setStyleResourceID(R.style.Share_Sheet_Style);
 
-                branchUniversalObject.showShareSheet(MainActivity.this, linkProperties, shareSheetStyle, new Branch.BranchLinkShareListener() {
+                branchUniversalObject.showShareSheet(MainActivity.this, linkProperties, shareSheetStyle, new Branch.BranchLinkShareListener2() {
                     @Override
                     public void onShareLinkDialogLaunched() {
                     }
@@ -274,6 +274,20 @@ public class MainActivity extends Activity {
 
                     @Override
                     public void onChannelSelected(String channelName) {
+                    }
+
+                    @Override
+                    public String getSharingTitleForChannel(String channel) {
+                        return channel.contains("messenger") ? "title for messenger" :
+                                channel.contains("slack") ? "title for slack" :
+                                channel.contains("Gmail") ? "title for gmail" : null;
+                    }
+
+                    @Override
+                    public String getSharingMessageForChannel(String channel) {
+                        return channel.contains("messenger") ? "message for messenger" :
+                                channel.contains("slack") ? "message for slack" :
+                                channel.contains("Gmail") ? "message for gmail" : null;
                     }
                 });
             }
--- a/Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
+++ b/Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
@@ -463,6 +463,50 @@ public class BranchUniversalObject implements Parcelable {
         }
     }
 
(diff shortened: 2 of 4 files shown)


Branch-SDK/src/io/branch/referral/ShareLinkManager.java
```diff
@@ -31,6 +31,7 @@
     /* The custom chooser dialog for selecting an application to share the link. */
     AnimatedDialog shareDlg_;
     Branch.BranchLinkShareListener callback_;
+    Branch.BranchLinkShareListener2 callback2_;
```
aaustin: can we name these a little more distinctly now that they have different functions?

aaustin: applies to all v2 labels above

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3169,6 +3169,45 @@ public void onActivityDestroyed(Activity activity) {
 
     /**
      * <p>An Interface class that is implemented by all classes that make use of
+     * {@link BranchLinkShareListener}, defining methods to listen for link sharing status.</p>
+     */
+    public interface BranchLinkShareListener2 {
```
aaustin: Instead of just duplicating the functionality, not sure, but is there a way to just extend the previous. Also, name change please.

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
```diff
@@ -259,7 +259,7 @@ public void onClick(View view) {
                 // Define custom styel for the share sheet list view
                 //.setStyleResourceID(R.style.Share_Sheet_Style);
 
-                branchUniversalObject.showShareSheet(MainActivity.this, linkProperties, shareSheetStyle, new Branch.BranchLinkShareListener() {
+                branchUniversalObject.showShareSheet(MainActivity.this, linkProperties, shareSheetStyle, new Branch.BranchLinkShareListener2() {
```
sojanpr: Can we keep this interface  unchanged? We can provide a separate builder method to set this call back so that existing partners are not affected with this change.

sojanpr: Can we name this new interface as `IChannelProperties` and builder method as `setChannelProperties(IChannelProperties)`

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
```diff
@@ -275,6 +275,20 @@ public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchE
                     @Override
                     public void onChannelSelected(String channelName) {
                     }
+
+                    @Override
+                    public String getSharingTitleForChannel(String channel) {
+                        return channel.contains("messenger") ? "title for messenger" :
+                                channel.contains("slack") ? "title for slack" :
```
sojanpr: Seems like this hard codings are accidental. Let us not dependent on any hard coded channel

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
```diff
@@ -463,6 +463,50 @@ public void showShareSheet(@NonNull Activity activity, @NonNull LinkProperties l
         }
     }
 
+    public void showShareSheet(@NonNull Activity activity, @NonNull LinkProperties linkProperties, @NonNull ShareSheetStyle style, @Nullable Branch.BranchLinkShareListener2 callback) {
+        if (Branch.getInstance() == null) {  //if in case Branch instance is not created. In case of user missing create instance or BranchApp in manifest
```
sojanpr: We don't need this changes if we set a builder method

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Seems like we can achieve this without changing much as follows

PR #232, opened 2016-04-07
Fixing orientation issue

sojanpr:
Fixing bug
“Branch views are not shown in a  session after Branch view is
dismissed by orientation change.”

@aaustin @Sarkar @EvangelosG


Branch-SDK/src/io/branch/referral/util/BranchViewHandler.java
```diff
@@ -296,5 +298,12 @@ public void updateUsageCount(Context context, String branchViewID) {
         void onBranchViewError(int errorCode, String errorMsg, String action);
     }
 
+    public void onActivityDestroyed(Activity activity) {
```
Sarkar: this is a bit confusing since it's the same as the `onActivityDestroyed` provided by Application lifecycle, can we rename? at first i thought you were attempting to hook into the application lifecycle callback.

Sarkar:
:+1: makes sense to me.

aaustin:
:+1:

PR #233, opened 2016-04-11
settings activity

EvangelosG:
@sojanpr 

Added settings activity (long press "Share Link" to launch activity)

Allows you to set Branch Key w/ UI.
Allows you to set deeplinkdebug params w/ UI.

UI auto populates currently set debugparams when killing/re-entering activity. I also included "next" and "enter" functionality to make adding debugparams a friendly experience. Press Enter after typing in a key -> focus goes to next edittext, press enter and both edittexts clear + the debugparams are set, and focus goes back to original edittext.

Current blocker: I thought that calling setBranchKey("") would cause app to use original Branch Key from XML file, but it does not. If you set an empty Branch Key (using PrefHelper) and exit->enter app, app is still using empty String. If you set an empty Branch Key (using PrefHelper) and exit to main activity, app is still using empty String and not original XML String.

### DIFF
--- a/Branch-SDK-TestBed/AndroidManifest.xml
+++ b/Branch-SDK-TestBed/AndroidManifest.xml
@@ -45,6 +45,8 @@
 
         <activity android:name="io.branch.branchandroiddemo.CreditHistoryActivity"/>
 
+        <activity android:name="io.branch.branchandroiddemo.SettingsActivity"/>
+
         <activity android:name="io.branch.branchandroiddemo.ReferralCodeActivity"/>
 
         <activity android:name=".AutoDeepLinkTestActivity">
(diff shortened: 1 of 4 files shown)


sojanpr:
Clear setting seems to be not working. The old Branch key is used after clearing settings

EvangelosG:
Updated clearing of branch Key. Switched from prefHelper.getBranchKey to prefHelper.readBranchKey. Added isTestModeEnabled from BranchUtil to activity.

EvangelosG:
Switched
io.branch.sdk.TestMode
from true/false. The reset function respects this setting correctly. Will get to other tests tomorrow!

EvangelosG:
First time I switch to a new key, init session counts as an install, subsequent inits count as open. Would love for @sojanpr @Sarkar to test out setting a new key w/ testbed app, and using clear settings button to revert back to original key. In test mode you will notice that clearing settings reverts to test key, same for live.

PR #240, opened 2016-05-12
Adding try-catch for additional safety

sojanpr:
Unmarshalling errors are reported on some corner cases while reading
data from intent

@aaustin @EvangelosG @Sarkar


EvangelosG:
Looks like a simple enough change to me 👍

PR #243, opened 2016-05-19
Add extras to requests

Sarkar:
@sojanpr 

the goal is to add an extra entry to all requests, called "metadata". inside it, anyone can specify a key with any type of value. 

we whitelist values on the server - something to keep in mind. 

usage: should be added after get instance but before initSession. can be added after getAutoInstance.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -679,7 +679,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents {
     /**
      * <p>Add key value pairs to all requests</p>
      */
-    public void addMetadataToRequests(@NonNull String key, JSONObject value) {
+    public void addMetadataToRequests(@NonNull String key, Object value) {
         prefHelper_.setRequestMetadata(key, value);
     }
 
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -679,7 +679,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents {
     /**
      * <p>Add key value pairs to all requests</p>
      */
-    public void addMetadataToRequests(@NonNull String key, Object value) {
+    public void setRequestMetadata(@NonNull String key, @NonNull String value) {
         prefHelper_.setRequestMetadata(key, value);
     }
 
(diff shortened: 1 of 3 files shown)


Branch-SDK/src/io/branch/referral/PrefHelper.java
```diff
@@ -1024,6 +1032,18 @@ public void disableSmartSession() {
         BNC_Smart_Session = false;
     }
 
+    public void setRequestMetadata(@NonNull String key, Object value) {
```
sojanpr: Since retadata Object is added and retrieved as  JSON Object internally is there any reason for specifying param type as `Object`

Sarkar: No - I originally had it as JSONObject, but moved it to generic Object when I modified the exposed class in `Branch.java`. It allows for someone to do:

`addRequestToMetadata("key", "value")`

let me know thoughts

Branch-SDK/src/io/branch/referral/ServerRequest.java
```diff
@@ -137,6 +138,11 @@ public String getRequestUrl() {
      *             as key-value pairs.
      */
     protected void setPost(JSONObject post) {
+        try {
+            post.put("metadata", prefHelper_.getRequestMetadata());
```
sojanpr: Also can we rename this form `metadata` to some thing else. Already ActionCompleted Request and RegisterView Request has metadata param

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -677,6 +677,13 @@ public void disableSmartSession() {
     }
 
     /**
+     * <p>Add key value pairs to all requests</p>
+     */
+    public void addMetadataToRequests(@NonNull String key, Object value) {
```
derrickstaten: there's `addMetadataToRequests` here, then in Pref helper there's `setRequestMetadata`. is there a reason why these methods have slightly different signatures?

Branch-SDK/src/io/branch/referral/PrefHelper.java
```diff
@@ -1024,6 +1032,26 @@ public void disableSmartSession() {
         BNC_Smart_Session = false;
     }
 
+    public void setRequestMetadata(@NonNull String key, @NonNull String value) {
```
derrickstaten: so how does someone remove a key if value has `@NonNull`?

Sarkar: unfortunately, these annotations cant account for run time null values, so we remove the entry below.

derrickstaten: got it, you can tell that I'm a novice :)

sojanpr:
LGTM 👍

PR #244, opened 2016-05-20
App Link  Install Support

sojanpr:
Adding support for FB deferred app link Data

@aaustin

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -17,6 +17,7 @@ import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.StyleRes;
+import android.text.TextUtils;
 import android.util.Log;
 
 import org.json.JSONArray;
@@ -2935,7 +2936,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents {
         processNextQueueItem();
     }
 
-    private void initializeSession(BranchReferralInitListener callback) {
+    private void initializeSession(final BranchReferralInitListener callback) {
         if ((prefHelper_.getBranchKey() == null || prefHelper_.getBranchKey().equalsIgnoreCase(PrefHelper.NO_STRING_VALUE))
                 && (prefHelper_.getAppKey() == null || prefHelper_.getAppKey().equalsIgnoreCase(PrefHelper.NO_STRING_VALUE))) {
             initState_ = SESSION_STATE.UNINITIALISED;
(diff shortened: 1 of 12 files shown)


Branch-SDK/src/io/branch/referral/ServerRequestRegisterInstall.java
```diff
@@ -94,6 +95,7 @@ public void onRequestSucceeded(ServerResponse resp, Branch branch) {
             prefHelper_.setExternalIntentExtra(PrefHelper.NO_STRING_VALUE);
             prefHelper_.setAppLink(PrefHelper.NO_STRING_VALUE);
             prefHelper_.setPushIdentifier(PrefHelper.NO_STRING_VALUE);
+            prefHelper_.setIsAppLinkTriggeredInit(false);
```
aaustin: The primary pathway this can be triggered on on install. We'll need to pass the boolean up here.

aaustin: oops - I misread .. sorry. But you'll need this in the `onRequestSucceeded` for RegisterOpen too I believe, right?

Branch-SDK/src/io/branch/referral/DeferredAppLinkDataHandler.java
```diff
@@ -0,0 +1,76 @@
...
+class DeferredAppLinkDataHandler {
+    private static final String NATIVE_URL_KEY = "com.facebook.platform.APPLINK_NATIVE_URL";
+
+    public static Boolean fetchDeferredAppLinkData(Context context, final AppLinkFetchEvents callback) {
+        boolean isRequestSucceeded = true;
+        try {
+            // Init FB SDK
+            Class<?> FacebookSdkClass = Class.forName("com.facebook.FacebookSdk");
+            Method initSdkMethod = FacebookSdkClass.getMethod("sdkInitialize", Context.class);
+            initSdkMethod.invoke(null, context);
```
aaustin: Did you find that this was required? I think I was able to get it to work without calling init on the FB SDK.

Branch-SDK/src/io/branch/referral/DeferredAppLinkDataHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+            Object completionListenerInterface = Proxy.newProxyInstance(AppLinkDataCompletionHandlerClass.getClassLoader()
+                    , new Class<?>[]{AppLinkDataCompletionHandlerClass}
+                    , ALDataCompletionHandler);
+
+            String fbAppID = context.getString(context.getResources().getIdentifier("facebook_app_id", "string", context.getPackageName()));
+            if (TextUtils.isEmpty(fbAppID)) {
+                isRequestSucceeded = false;
+            } else {
+                fetchDeferredAppLinkDataMethod.invoke(null, context, fbAppID, completionListenerInterface);
```
aaustin: Is it possible to use the method w/o the app id? That was the one I'm using in BMF.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2949,9 +2954,43 @@ private void initializeSession(BranchReferralInitListener callback) {
...
+                        if (!TextUtils.isEmpty(bncLinkClickId)) {
+                            prefHelper_.setLinkClickIdentifier(bncLinkClickId);
+                            prefHelper_.setIsAppLinkTriggeredInit(true);
+                        }
+                    }
+                    BranchReferralInitListener deferredCallback = null;
+                    if (deferredInitListener_ != null) {
+                        deferredCallback = deferredInitListener_.get();
+                    }
+                    registerAppInit(deferredCallback);
```
aaustin: Can you test in airplane mode and the like to ensure that the Facebook SDK always calls back. We don't want them to introduce an edge case that causes us to fail to initialize.

Branch-SDK/src/io/branch/referral/DeferredAppLinkDataHandler.java
sojanpr: I have seen the error "Unable to fetch deferred params since SDK is not initialized" .I couldn't find the exact error checking the code but FB code seems like calling initialize if we don't do it when calling `fetchDeferredParams`

We wanted to catch all possible error before calling `fetchDeferredParams` to avoid waiting for errors in callback

Branch-SDK/src/io/branch/referral/DeferredAppLinkDataHandler.java
sojanpr: The methods with out app id internally calls `fetchDeferredAppLinkFromServer(Context context, String applicationId, AppLinkData.CompletionHandler completionHandler) ` method and they read the app id (The `facebook_app_id` from string resource) .
IN case user  does not have it provided , using this approach we can find it before hands rather than waiting for the callback.

Branch-SDK/src/io/branch/referral/ServerRequestRegisterInstall.java
sojanpr: Good catch. Will update this

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: I have already tested with no internet condition and it seems it called back with in 500 ms.

sojanpr:
Fixed  the issue with `facebook_app_link_checked` state. Review  ready

Branch-SDK/src/io/branch/referral/DeferredAppLinkDataHandler.java
aaustin: Nice

aaustin:
Looks good ! 👍 Please fix conflicts

PR #247, opened 2016-05-26
Fixing a potential case of callback getting nullified

sojanpr:
Call back may get nullified if  init seen is called  before auto
session start and  if Fb  app link check is on going

@aaustin @Sarkar @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -1092,8 +1092,8 @@ public class Branch implements BranchViewHandler.IBranchViewEvents {
             if (initState_ == SESSION_STATE.INITIALISING) {
                 if (callback != null) {
                     deferredInitListener_ = new WeakReference<>(callback);
+                    requestQueue_.setInstallOrOpenCallback(callback);
                 }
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2945,7 +2947,9 @@ private void registerInstallOrOpen(ServerRequest req, BranchReferralInitListener
         // be cleared if the app is terminated while an Open/Install is pending.
         else {
             // Update the callback to the latest one in initsession call
-            deferredInitListener_ = new WeakReference<>(callback);
+            if (callback != null) {
+                deferredInitListener_ = new WeakReference<>(callback);
+            }
             requestQueue_.setInstallOrOpenCallback(callback);
```
Sarkar: does requestQueue ignore null here?

sojanpr: Good point sahil , We already do `if (callback != null) {` check inside the method `setInstallOrOpenCallback`

aaustin:
seems good to me 👍

EvangelosG:
👍

Sarkar:
👍

PR #251, opened 2016-06-06
Adding Fabric res to SDK

sojanpr:
According to the Doc the Fabric properties file should be added  to the
SDK as a Java resource in the generated  war file. Adding properties as
separate resource as per the doc

@aaustin @Sarkar


Sarkar:
👍

aaustin:
👍

PR #252, opened 2016-06-07
A possible case for init callback fail

sojanpr:
This is a  corner case when calling `BranchInit` with BUO on devices
4.4 or less.
Basically the reference to init wrapper is created inside the method
and loses  the reference by the time FB `onAppLinkFetchFinished` method
called. Leasing to the issue init callback is not called.

@aaustin @Sarkar @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -363,6 +363,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents {
     private final ConcurrentHashMap<String, String> instrumentationExtraData_;
 
     private WeakReference<BranchReferralInitListener> deferredInitListener_;
+    BranchUniversalReferralInitWrapper branchUniversalReferralInitWrapper_;
 
     /**
      * <p>The main constructor of the Branch class is private because the class uses the Singleton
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3472,7 +3473,11 @@ else if (thisReq_ instanceof ServerRequestLogout) {
                 } catch (JSONException ex) {
                     ex.printStackTrace();
                 }
+                if (thisReq_ instanceof ServerRequestInitSession) {
+                    branchUniversalReferralInitWrapper_ = null;
```
aaustin: If we nullify the wrapper on response, then, when the app is opened again and the Application detects a new Activity added to the stack, there's a risk that the callback won't be referenced. Isn't that right?

aaustin: This is why we don't nullify the callback elsewhere

sojanpr: The call back is nullified only after response is sent for "initSession". Since there is an   additional async operation  `fetchDeferredAppLinkData` the callback wrapper need a strong reference until  the request queued

aaustin: I got it. But what if Auto Session Management kicks in, triggering a new session next time the app is opened. The callback won't be called because you've nullified it? I didn't check this path but if you did, I'm 👍

sojanpr: The CallbackWrapper is nullified only on receiving install/open response and only after calling  `onInitFinished`.  So once callback set it will get nullified only after calling `onInitFinished`

sojanpr:
A better solution is added in this PR https://github.com/BranchMetrics/android-branch-deep-linking/pull/253

PR #254, opened 2016-06-09
Fabric answers stuff

sojanpr:
Adding support for reporting install / open / share events to fabric answers

Install / open / share events are send to Answers if Fabric is enabled
for the  application. A Kit Event with desecrate key-value pairs
associated with the event  is send to Answers through Answers-Shim APIs

@aaustin @ahmednawar


Branch-SDK/src/io/branch/referral/ExtendedDataProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Pass the Branch event to integrated third party SDKs
+     *
+     * @param request  {@link ServerRequest} associated with the Branch event
+     * @param response {@link ServerResponse} associated with the Branch event
+     */
+    public void provideData(ServerRequest request, ServerResponse response) {
+        // If Fabric is enabled
+        if (prefHelper_.getIsFabricEnabled()) {
```
aaustin: Can we remove this whole class for now and just put this logic in the appropriate place (install, open, share)? I don't really like that we're adding a new static data provider in every new SDK installation while only a small percentage of apps will use it.

Branch-SDK/src/io/branch/referral/ServerRequest.java
```diff
@@ -30,6 +30,7 @@
     private SystemObserver systemObserver_;
     long queueWaitTime_ = 0;
     private boolean disableAndroidIDFetch_;
+    protected  final ExtendedDataProvider extendedDataProvider_;
```
aaustin: if following the logic above, we can get rid of these references/instantiation

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
```diff
@@ -117,6 +118,7 @@ public void onRequestSucceeded(ServerResponse resp, Branch branch) {
         } catch (Exception ex) {
             ex.printStackTrace();
         }
+        extendedDataProvider_.provideData(this, resp);
```
aaustin: probably want to use that boolean you've worked so hard to pass in here ;)

Branch-SDK/src/io/branch/referral/ExtendedAnswerProvider.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * <p>
+     * Method for sending the kit events to Answers for Branch Events.
+     * Method create a Json Object to flatten and create the {@link KitEvent} with key-values
+     * </p>
+     *
+     * @param request  {@link ServerRequest} corresponding to the event
+     * @param response {@link ServerResponse} corresponding to the event
+     */
+    public static void provideData(ServerRequest request, ServerResponse response) {
```
aaustin: Can you just pass in the event name like Ahmed does [here[(https://github.com/BranchMetrics/ios-branch-deep-linking/pull/367/files)? You have that context in every original call? By passing in a server request, you're creating a dependency of this class on external code structure.

Branch-SDK/src/io/branch/referral/ExtendedAnswerProvider.java
```diff
@@ -0,0 +1,152 @@
...
+                    jsonData.put(Defines.Jsonkey.Data.getKey(), new JSONObject(response.getObject().getString(Defines.Jsonkey.Data.getKey())));
+                }
+            } else if (request instanceof ServerRequestCreateUrl && ((ServerRequestCreateUrl) request).isReqStartedFromBranchShareSheet()) {
+                BranchLinkData linkData = ((ServerRequestCreateUrl) request).getLinkPost();
+                if (linkData != null && linkData.has(Defines.Jsonkey.Data.getKey())) {
+                    kitEvent = new ExtendedKitEvent(KIT_EVENT_SHARE);
+                    jsonData = new JSONObject();
+                    JSONObject linkDataJson = new JSONObject(linkData.getString(Defines.Jsonkey.Data.getKey()));
+
+                    jsonData.put(Defines.LinkParam.Channel.getKey(), linkData.getChannel());
```
aaustin: Is there anyway to iterate through the data rather than hard code all these specific calls? I'd err on sending more data than not. See Ahmed's implementation.

Branch-SDK/src/io/branch/referral/ExtendedAnswerProvider.java
```diff
@@ -0,0 +1,152 @@
...
+
+        public void addBranchAttributes(String keyPathPrepend, String key, String value) {
+            if (!TextUtils.isEmpty(value)) {
+                String modifiedKey = keyPathPrepend + key;
+                if (key.startsWith(CTRL_PARAM_NOTATION)) {
+                    modifiedKey = key.replaceFirst(CTRL_PARAM_NOTATION, "");
+                } else if (modifiedKey.startsWith(CTRL_PARAM_NOTATION)) {
+                    modifiedKey = modifiedKey.replaceFirst(CTRL_PARAM_NOTATION, "");
+                }
+                PrefHelper.Debug("kittest", modifiedKey + " = " + value);
```
aaustin: make sure we remove this log

Branch-SDK/src/io/branch/referral/BranchLinkData.java
```diff
@@ -397,4 +397,30 @@ public int hashCode() {
...
+     *
+     * @return {@link JSONObject} with link params and link properties
+     */
+    public JSONObject getLinkDataJsonObject() {
+        JSONObject linkDataJson = null;
+        try {
+            if (has(Defines.Jsonkey.Data.getKey())) {
+                linkDataJson = new JSONObject(getString(Defines.Jsonkey.Data.getKey()));
+
+                linkDataJson.put("~" + Defines.LinkParam.Channel.getKey(), getChannel());
```
aaustin: do we need to worry about these being null?

aaustin:
after that, looks good 👍

Branch-SDK/src/io/branch/referral/BranchLinkData.java
sojanpr: Ahh good catch . It cannot be null with BUO but yes if we implement the share sheet without BUO. Will Add the safe check

Branch-SDK/build.gradle
```diff
@@ -6,6 +6,7 @@ def jarFileName
 dependencies {
     compile fileTree(dir: 'libs', include: '*.jar')
     compile 'com.android.support:support-annotations:22.2.0'
+    compile(name:'answers-shim', ext:'jar')
```
stevewilber: The shim is in jCenter now, you can add it with ```compile('com.crashlytics.sdk.android:answers-shim:0.0.3@aar')```

sojanpr: @stevewilber just noticed that there is change in package name than from the jar used earlier 
`com.crashlytics.sdk.android.answers.shim.AnswersOptionalLogger`  to `com.crashlytics.android.answers.shim.AnswersOptionalLogger`. Is that intentional?

stevewilber: Yeah we adjusted the package name to be more consistent with the sdks

aaustin:
Let's get this guy deployed once we get approval from QA

Branch-SDK/src/io/branch/referral/ServerRequestInitSession.java
```diff
@@ -71,6 +71,26 @@ public boolean handleBranchViewIfAvailable(ServerResponse resp) {
...
 
+    @Override
+    public void onRequestSucceeded(ServerResponse response, Branch branch) {
+        // Check for any Third party SDK for data handling
+        String appOrigin = "";
+        try {
+            if (response.getObject().has(Defines.Jsonkey.AppOrigin.getKey())) {
+                appOrigin = response.getObject().getString(Defines.Jsonkey.AppOrigin.getKey());
+            }
+            boolean isFabricEnabled = appOrigin.equals(Defines.Jsonkey.AppOriginTwitterFabric.getKey());
```
stevewilber: With this check, people will only be able to use Branch with Fabric if they originally provisioned the Branch account through Fabric, right? If they already have a Branch account and then decide they wanted to use it with Fabric they'd be totally out of luck.

What I'd suggest here is dropping this check and pass the event to ```ExtendedAnswerProvider().provideData()```. The AnswersOptionalLogger will noop if the client doesn't have Fabric/Answers installed.

sojanpr: @stevewilber  This will get called each time they open the app. So existing users if they decide to use Answers with Branch, this will be true next time when the app opens.

stevewilber: How does the AppOrigin get set in the backend?

aaustin: Hey Steve - Just for the sake of not surprising any existing customers by messing up their Answers dash, we decided to only enable the data integration for users who added Branch through Fabric. So, to answer your question, App Origin is set when the app is created or updated via Fabric. 

Looking forward, this value is something that we can easily expose as an integration to our partners via the dash, but it would be opt in for existing Branch folks.

stevewilber: Hey Austin, we definitely want this to work for existing Branch/Fabric customers without requiring any extra work on their part. We're big believers in "it just works". Can we remove this or should we get on a call to discuss?

jzazove: I agree.  This was our shared narrative around a true partnership - where things work automagically for our mutual customers.  @aaustin: do you want to jump on a call to talk?

aaustin: I think we should. I'll follow up in the thread!

aaustin:
👍

PR #255, opened 2016-06-14
Improved Branch request handling in slow network

sojanpr:
On slow networks If the hostname resolves to multiple IP addresses, URL
connection will attempt to connect the next if one connection fails.
This may result is multiple timeouts elapse before the connect attempt
throws an exception. In order to make sure the Branch callback is
called with in the timeout specified we resolve the host first and then
try only with first resolved ip.
This will ensure that Branch Callbacks are retuned in the time out
specified (Default is 3sec sec and 3 attempts. so 9 sec max).
This fix will solve the issue “Branch init callback returned after
long wait on slow networks”

@aaustin @Sarkar @EvangelosG @hubt

### DIFF
--- a/Branch-SDK/src/io/branch/referral/RemoteInterface.java
+++ b/Branch-SDK/src/io/branch/referral/RemoteInterface.java
@@ -15,11 +15,17 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
+import java.net.InetAddress;
+import java.net.MalformedURLException;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
 import java.net.URL;
 import java.net.UnknownHostException;
 import java.util.Iterator;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLSession;
 
 /**
  * <p>This class assists with RESTful calls to the Branch API, by using
@@ -152,7 +158,7 @@ class RemoteInterface {
     private ServerResponse make_restful_get(String baseUrl, JSONObject params, String tag, int timeout, int retryNumber, boolean log) {
         String modifiedUrl = baseUrl;
         JSONObject getParameters = new JSONObject();
-        HttpURLConnection connection = null;
+        HttpsURLConnection connection = null;
         if (timeout <= 0) {
             timeout = DEFAULT_TIMEOUT;
         }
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -173,13 +179,22 @@ private ServerResponse make_restful_get(String baseUrl, JSONObject params, Strin
...
             if (log) PrefHelper.Debug("BranchSDK", "getting " + modifiedUrl);
             lastRoundTripTime_ = 0;
             long reqStartTime = System.currentTimeMillis();
             URL urlObject = new URL(modifiedUrl);
-            connection = (HttpURLConnection) urlObject.openConnection();
+            connection = (HttpsURLConnection) urlObject.openConnection();
             connection.setConnectTimeout(timeout);
             connection.setReadTimeout(timeout);
+            // Default host verifier will provide SSL error on host verification since ip used in request is not mapped to a certificate
+            connection.setHostnameVerifier(new HostnameVerifier() {
```
aaustin: hmm - can you help me understand what this is for?

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -431,4 +456,18 @@ private String convertJSONtoString(JSONObject json) {
...
     }
+
+
+    private String getResolvedApiUrl( String url) throws MalformedURLException, UnknownHostException {
+        // If the hostname resolves to multiple IP addresses, URL connection will attempt to connect the next if one connection fails
+        // This may result is multiple timeouts elapse before the connect attempt throws an exception. In order to make sure the
+        // Branch callback is called with in the timeout specified we resolve the host first and then try only with first resolved ip
+
+        String apiHost = new URL(url).getHost();
+        InetAddress inetAddress = InetAddress.getByName(apiHost);
```
aaustin: What if this is a stale IP for an API machine that we've shut down? Wouldn't a better way to do this be to just clear the list of mapped IPs for our hostname every so often?

Branch-SDK/src/io/branch/referral/RemoteInterface.java
sojanpr: Android internally check certificate for any secure connection. The way it find the SSL certificate is using the CommonName of the host. When we change the host with its IP Android won't be able to  verify SSL and throw `hostNotVerified` exception. This code will override Android SSL check and return it as always verified

aaustin: Yuck - can you see if the path I suggest below will work?

aaustin:
👍

Branch-SDK/src/io/branch/referral/RemoteInterface.java
sojanpr: This has been tested and since the cached entries expires after TTL this seems not an issue.  Also    existing connections honored until TTL in anycase.

hongrich:
I'm not familiar with this codebase, but overriding the default hostname verifier to allow everything seems like a very dangerous thing to do at first glance. If the hostname is not being verified, a man-in-the-middle attack can easily trick this into sending data to an IP that you don't control (but with a valid SSL certificate otherwise).

Just want to share my concern.

sojanpr:
@hongrich  Thank you very much for the review and valuable comment. We are only overriding SSL verification only on the connection instance that we use internally in Branch. So do you think this can introduce an hole for middle man attack?

hongrich:
Yes. This is an SDK that other developers include in their apps, so it's almost always run in a network environment that you don't control and skipping hostname verification is probably not the best idea.

sojanpr:
@hongrich  Thank you very much for the suggestion. We understand the security concern and will revert this change. Thanks

sojanpr:
@hongrich  We have created a new PR https://github.com/BranchMetrics/android-branch-deep-linking/pull/258 and reverted the changes. Thank you very much

PR #256, opened 2016-06-16
Increase retry interval to 1 s

aaustin:
There are reports of high client timeout errors despite impeccable
server performance. I have a hypothesis that by not waiting some time
between retries, we don’t give the connection time to recover,
therefore rendering the retry code useless. I don’t see how else with 3
retries we’d have the same timeout issues.

@sojanpr


sojanpr:
LGTM 👍

aaustin:
yay

PR #261, opened 2016-06-21
Adding answers-shim from Maven

sojanpr:
Answers shim 0.0.3 is added to maven and using with Branch SDK to get
rid off jcenter dependency.
This would be changed when Answers-Shim is available on both maven and
center

@aaustin @stevewilber


aaustin:
👍

PR #264, opened 2016-07-05
Adding full width share sheet

sojanpr:
Adding option to set a full width share sheet with option to customize
divider

@aaustin @Sarkar @EvangelosG


EvangelosG:
Compiles / works. I agree w/ #android chat that some header text would look great.

aaustin:
looks great 👍

PR #266, opened 2016-07-09
Faster GAId Fetch

sojanpr:
Optimizing the way of fetching google ads params

@aaustin @EvangelosG


aaustin:
Nice catch - looks good to me 👍

PR #271, opened 2016-07-22
Fix Stale Intent Issue

sojanpr:
On newer Android devices there seems an issue with calling
`onNewIntent()` with `singleTop`/`SingleInstance` activities.
`onNewIntent()` is called after `onStart` method where  it is expected
to be before `onStart`. To fix the issue moving init requests execution
to  `onResume` in lifecycle

@aaustin @EvangelosG @Sarkar


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3072,9 +1962,22 @@ private void registerAppInit(BranchReferralInitListener callback, ServerRequest.
...
             request.addProcessWaitLock(ServerRequest.PROCESS_WAIT_LOCK.GAID_FETCH_WAIT_LOCK);
         }
+        if (intentState_ != INTENT_STATE.READY) {
+            request.addProcessWaitLock(ServerRequest.PROCESS_WAIT_LOCK.INTENT_PENDING_WAIT_LOCK);
+        }
         registerInstallOrOpen(request, callback);
     }
 
+    private void onIntentReady(Activity activity) {
+        Log.d("BranchSDK", "onIntentReady");
```
aaustin: note about log here

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -3142,18 +2045,19 @@ private void setActivityLifeCycleObserver(Application application) {
...
         public void onActivityCreated(Activity activity, Bundle bundle) {
+            intentState_ = INTENT_STATE.PENDING;
             if (BranchViewHandler.getInstance().isInstallOrOpenBranchViewPending(activity.getApplicationContext())) {
                 BranchViewHandler.getInstance().showPendingBranchView(activity);
             }
         }
 
         @Override
         public void onActivityStarted(Activity activity) {
+            intentState_ = INTENT_STATE.PENDING;
```
aaustin: Doesn't this mean that we won't do any initialization until onResume is called now? I thought we discussed just doing this in the scenario where we determine the intent has stale data?

aaustin: And if so, how much additional latency do you anticipate this introducing?

aaustin:
Other than the above, the cleanup looks great!

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Some times even on create is calling with null data on second link click.
So there seems no better way to figure out a stale intent. Will check
furrher and update the code if there is a unique way to find satle intent
always

sojanpr: The latency with this solution raelly depends on where the ini session is
called. Assuming it is called in "onstart" the latency really depends on
the code added to the onStart

aaustin: Can you test with testbed and just let me know the delta?

sojanpr: For the test bed App the additional delay this adds seems to be `2 - 3 msec`. 
(For TestBed app there is no operation carried out in `onStart()` other than Branch#InitSession)

aaustin: Instead of adding this additional latency, can you only do this in Marshmallow or greater? Then any other instance you can just set `INTENT_STATE.READY` in Started.

aaustin:
Nice - Looks great 👍

PR #273, opened 2016-07-22
BranchUniversalObject.getShortUrl() return long url when called inside AsyncTask

bubinimara:


sojanpr:
@bubinimara  Thank you for submitting the PR. With this solution `getShortUrl` will work only form a Background thread. It will create `NetworkOnMainThread` if called form UI thread.

If you are  trying to use `getShortUrl()` with an Async Task, do you think it is possible to call `getShortUrl()` method from `onPreExecute()` of the task and use the link in `doInBackground` method?

bubinimara:
From `onPreExectute` and `onPostExecute` will work because the function is called from the UI thread but the ui will be no smooth, it block until the call end.

When I see that the function is documented as sync, I  had supposed to call it from a background thread.
In fact I use `generateShortUrl()` for other calls used in  the UI thread .

I think that a method to create a short url from a background thread is really useful.

sojanpr:
@bubinimara Agree with you on the above. It is good to have sync create url method that can be called from background thread also. Do you think the following can be a good approach since it can be called from any thread or Async task.
```private String generateShortLinkSync(final ServerRequestCreateUrl req) {
        if (initState_ == SESSION_STATE.INITIALISED) {
            final ServerResponse[] response = {null};
            try {
                //response = new getShortLinkTask().execute(req).get(timeOut, TimeUnit.MILLISECONDS);
                final CountDownLatch latch = new CountDownLatch(1);
                new Thread(new Runnable() {
                    @Override
                    public void run() {
                        response[0] = kRemoteInterface_.createCustomUrlSync(req.getPost());
                        latch.countDown();
                    }
                }).run();
                
                int timeOut = prefHelper_.getTimeout() + 2000; // Time out is set to slightly more than link creation time to prevent any edge case
                latch.await(timeOut, TimeUnit.MILLISECONDS);
                
                
            } catch (InterruptedException ignore) {
            }
            String url = req.getLongUrl();
            if (response[0] != null && response[0].getStatusCode() == HttpURLConnection.HTTP_OK) {
                try {
                    url = response[0].getObject().getString("url");
                    if (req.getLinkPost() != null) {
                        linkCache_.put(req.getLinkPost(), url);
                    }
                } catch (JSONException e) {
                    e.printStackTrace();
                }
            }
            return url;
        } else {
            Log.i("BranchSDK", "Branch Warning: User session has not been initialized");
        }
        return null;
    }```

sojanpr:
Since this is not the case  for most of our partners, we suggest to use the above approach  if you like to make the `generateShortLinkSync` call from a  background thread.

PR #275, opened 2016-07-24
Possible NPE with context going null

sojanpr:
Adding a proactive null check for the share link item selected.

@aaustin @EvangelosG


aaustin:
👍

PR #277, opened 2016-07-25
Fix initSession return value

sojanpr:
With the latest code the `initSession` operation is always expected to
be asynchronous and the actual results should be given with the
`BranchReferralInitListener` callback. The return value for this method
actually specifies whether the initSession Operation is started and
hence should be always true. Any errors for init session is provided
with `BranchErrors` in the callback

@aaustin @EvangelosG @Sarkar


aaustin:
👍

PR #280, opened 2016-07-29
White list for External intent URI

sojanpr:
Support for a  given list of URI Scheme as the external Uri white list.
Branch will collect  * external intent uri only for Uris in white list.
If no URI is added to the white list branch will collect all external
intent uris

@aaustin


aaustin:
Looks great 👍 Thanks for the fast turn around!

PR #281, opened 2016-08-01
Branch Content Discover

sojanpr:
Branch Content discover Implementation

@aaustin @dmitrig01 @shekar9


Branch-SDK/src/io/branch/referral/ServerRequestInitSession.java
```diff
@@ -72,7 +74,10 @@ public boolean handleBranchViewIfAvailable(ServerResponse resp) {
     }
 
     @Override
+
     public void onRequestSucceeded(ServerResponse response, Branch branch) {
+       ContentDiscoverManifest.getInstance(context_).onBranchInitialised(response.getObject());
```
aaustin: Nitpick but can you change this to `ContentDiscoveryManifest` - add the y.

Branch-SDK/src/io/branch/referral/ServerRequestRegisterClose.java
```diff
@@ -30,6 +33,18 @@ public ServerRequestRegisterClose(Context context) {
             if (!prefHelper_.getLinkClickID().equals(PrefHelper.NO_STRING_VALUE)) {
                 closePost.put(Defines.Jsonkey.LinkClickID.getKey(), prefHelper_.getLinkClickID());
             }
+            JSONObject ciObject = ContentDiscoverer.getInstance().getContentDiscoverDataForCloseRequest(context);
+            if (ciObject != null) {
+                closePost.put(Defines.Jsonkey.ContentDiscovery.getKey(), ciObject);
+            }
+            try {
+                PackageInfo info = context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
```
aaustin: A little worried about garbage collection & activity lifecycle with this code. Can you put this is in your `updateRequestWithDeviceParams` instead?

Branch-SDK/src/io/branch/referral/ContentDiscoverer.java
```diff
@@ -0,0 +1,239 @@
+package io.branch.referral;
```
aaustin: I thought we left this to put them in `indexing`

Branch-SDK/src/io/branch/referral/ContentDiscoverer.java
```diff
@@ -0,0 +1,239 @@
...
+
+            try {
+                if (cdManifest_.isCDEnabled() && lastActivityReference_ != null && lastActivityReference_.get() != null) {
+                    Activity activity = lastActivityReference_.get();
+                    JSONObject contentEvent = new JSONObject();
+                    contentEvent.put(TIME_STAMP_KEY, System.currentTimeMillis());
+                    contentEvent.put(ACTION_KEY, "v");
+                    contentEvent.put(NAV_PATH_KEY, contentNavPath_);
+                    contentEvent.put(REFERRAL_LINK_KEY, triggerUri_);
+                    contentEvent.put(CONTENT_LINK_KEY, "");
```
aaustin: eh? empty string?

aaustin: and empty dict below

aaustin: oh I see. deleting comment

aaustin: actually, yes, why is this and the below a blank string + dictionary.

Branch-SDK/src/io/branch/referral/ContentDiscoverer.java
```diff
@@ -0,0 +1,239 @@
...
+
+        int viewRenderWait = VIEW_SETTLE_TIME;
+        if (isSessionStart) {
+            triggerUri_ = null;
+        }
+
+        if (activity != null) {
+            Intent intent = activity.getIntent();
+            if (intent != null && intent.getData() != null && intent.getData().getScheme() != null) {
+                triggerUri_ = intent.getData().toString();
```
aaustin: This should respect our new whitelist array.

Branch-SDK/src/io/branch/referral/Defines.java
```diff
@@ -96,7 +96,18 @@
         BranchViewAction("action"),
         BranchViewNumOfUse("number_of_use"),
         BranchViewUrl("url"),
-        BranchViewHtml("html");
+        BranchViewHtml("html"),
+
+        Path("plath"),
```
aaustin: Correcting this on my local version to `path`, but noting in case you actually meant `plath`

Branch-SDK/src/io/branch/referral/PrefHelper.java
```diff
@@ -169,7 +171,8 @@ public static PrefHelper getInstance(Context context) {
      * API uses.
      */
     public String getAPIBaseUrl() {
-        return "https://api.branch.io/";
+        //return "https://api.branch.io/";
+        return "http://chandra.branch.io:8080/";
```
aaustin: note that we need to change this prior to release

Branch-SDK/src/io/branch/referral/RemoteInterface.java
```diff
@@ -148,7 +148,7 @@ private boolean addCommonParams(JSONObject post, int retryNumber) {
     private ServerResponse make_restful_get(String baseUrl, JSONObject params, String tag, int timeout, int retryNumber, boolean log) {
         String modifiedUrl = baseUrl;
         JSONObject getParameters = new JSONObject();
-        HttpsURLConnection connection = null;
+        HttpURLConnection connection = null;
```
aaustin: and all this stuff

Branch-SDK/src/io/branch/referral/ContentDiscoverer.java
sojanpr: They were added for passing more meta data in  future. Removed for now.

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
```diff
@@ -0,0 +1,256 @@
...
+ * Note that this feature can be controlled from the dashboard.
+ * </p>
+ */
+public class ContentDiscoverer {
+    private static ContentDiscoverer thisInstance_;
+
+    private Handler handler_;
+    private WeakReference<Activity> lastActivityReference_;
+    private static final int VIEW_SETTLE_TIME = 1000; /* Time for a view to load its components */
+    //private static final int APP_LAUNCH_DELAY = 2500; /* Time for an app to do initial pages like splash or logo */
```
aaustin: why leave this in there?

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
```diff
@@ -0,0 +1,256 @@
...
+            handler_.postDelayed(readContentRunnable, VIEW_SETTLE_TIME);
+        }
+    }
+
+
+    public void onActivityStopped(Activity activity) {
+        if (lastActivityReference_ != null && lastActivityReference_.get() != null
+                && lastActivityReference_.get().getClass().getName().equals(activity.getClass().getName())) {
+            handler_.removeCallbacks(readContentRunnable);
+            lastActivityReference_ = null;
```
aaustin: I'm worried we have a potential exception case here. If the activity stops and this is set to null in between line 119 and 120, we could see a crash. Perhaps we can use a semaphore to ensure that doesn't happen?

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
```diff
@@ -0,0 +1,256 @@
...
+                    contentEvent.put(TIME_STAMP_KEY, System.currentTimeMillis());
+                    contentEvent.put(NAV_PATH_KEY, contentNavPath_);
+                    if (!TextUtils.isEmpty(referredUrl_)) {
+                        contentEvent.put(REFERRAL_LINK_KEY, referredUrl_);
+                    }
+                    String viewName = "/" + activity.getClass().getSimpleName();
+                    contentEvent.put(VIEW_KEY, viewName);
+
+
+                    ViewGroup rootView = (ViewGroup) activity.findViewById(android.R.id.content);
```
aaustin: Perhaps null check if rootView is still present in the below. Since we have a weak reference to the activity, it could possibly be cleaned up prior to our referencing it?

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
```diff
@@ -0,0 +1,256 @@
...
+            } else {
+                String viewName = res.getResourceEntryName(childView.getId());
+                updateElementData(viewName, childView, isClearText, contentDataArray, contentKeysArray);
+            }
+        }
+    }
+
+    private void updateElementData(String viewName, View view, boolean isClearText, JSONArray contentDataArray, JSONArray contentKeysArray) {
+        String viewVal;
+        if (view instanceof TextView) {
```
aaustin: not sure if this handles nulls, but we should definitely do that in the case that our server-saved ID doesn't return anything (referring to line 173)

aaustin:
Other than my concerns above, this is looking great!

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
sojanpr: Before setting `LastActivity` weak reference to Null we are removing all `readContentRunnable` from the handler. So there is no chance to hit line 119 -120 after on stop for that activity. Is there any other path that I am missing for  potential NPE?

aaustin: The runnable happens in the background - right? It would seems it's possible for the runnable to start executing, pass through line 119, then this code to be executed resulting in you calling `get()` on null in 120.

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
sojanpr: This case should never happen since this runnable is removed on `onStop`. So this `runnable ` should run only in between `onStart` and `onStop` for a given activity . Will add a null check  anyways

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
sojanpr: `instanceof` won't throw NPE even if the compared object is `null`

Branch-SDK/src/io/branch/indexing/ContentDiscoverer.java
sojanpr: `readContentRunnable ` is happening in the UI thread since it is invoked from `Handler#postDelayed`. So there is no option for the runnable to run after `onStop` for an activity?

aaustin:
👍 let's do this - get it in BMF first?

PR #285, opened 2016-08-09
Possible NPE when API response is timeout

sojanpr:
NPE protection in case of Branch API response parse error
@aaustin @EvangelosG @Sarkar


Sarkar:
this makes sense to me 👍

aaustin:
👍

aaustin:
I thought we were doing a staging branch this time?

PR #286, opened 2016-08-22
Changing to Crashlytics Answers-Shim

sojanpr:
Replacing `io.branch.external.answersshim` dependency with
`com.crashlytics.sdk.android:answers-shim` in order to avoid dependency
hierarchy  issue when Branch is used with other Fabric components which
use `answersshim`

@aaustin @Sarkar


Sarkar:
👍

aaustin:
👍

PR #288, opened 2016-09-01
BUO Enhancements

sojanpr:
Making BUO great again
1) Adding Price and currency support
2) Adding userCompletedAction()  method to report user events on BUO
3) Fixing issue: Index mode is not send with link data 

@aaustin @ahmednawar @EvangelosG

### DIFF
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -19,7 +19,7 @@ import io.branch.referral.Branch.BranchReferralStateChangedListener;
 import io.branch.referral.BranchError;
 import io.branch.referral.BranchViewHandler;
 import io.branch.referral.SharingHelper;
-import io.branch.referral.util.CurrencyType;
+import io.branch.referral.util.CURRENCY_TYPE;
 import io.branch.referral.util.LinkProperties;
 import io.branch.referral.util.ShareSheetStyle;
 
@@ -51,7 +51,7 @@ public class MainActivity extends Activity {
                 .setContentIndexingMode(BranchUniversalObject.CONTENT_INDEX_MODE.PUBLIC)
                 .setContentType("application/vnd.businessobjects")
                         //.setContentExpiration(new Date(1476566432000L)) // set contents expiration time if applicable
-                .setPrice(5.00, CurrencyType.USD)
+                .setPrice(5.00, CURRENCY_TYPE.USD)
                 .addKeyWord("My_Keyword1")
                 .addKeyWord("My_Keyword2")
                 .addContentMetadata("Metadata_Key1", "Metadata_value1")
@@ -268,10 +268,12 @@ public class MainActivity extends Activity {
 
                             @Override
                             public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchError error) {
+                                branchUniversalObject.userCompletedAction(BranchUniversalObject.BUO_USER_ACTIONS.SHARE_COMPLETED);
                             }
 
                             @Override
                             public void onChannelSelected(String channelName) {
+                                branchUniversalObject.userCompletedAction(BranchUniversalObject.BUO_USER_ACTIONS.SHARE_STARTED);
                             }
 
                         },
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -260,6 +260,7 @@ public class MainActivity extends Activity {
 
                             @Override
                             public void onShareLinkDialogLaunched() {
+                                branchUniversalObject.userCompletedAction(BranchUniversalObject.BUO_USER_ACTIONS.SHARE_STARTED);
                             }
 
                             @Override
@@ -273,7 +274,6 @@ public class MainActivity extends Activity {
 
                             @Override
                             public void onChannelSelected(String channelName) {
-                                branchUniversalObject.userCompletedAction(BranchUniversalObject.BUO_USER_ACTIONS.SHARE_STARTED);
                             }
 
                         },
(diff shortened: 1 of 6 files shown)


Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
```diff
@@ -266,10 +268,12 @@ public void onShareLinkDialogDismissed() {
 
                             @Override
                             public void onLinkShareResponse(String sharedLink, String sharedChannel, BranchError error) {
+                                branchUniversalObject.userCompletedAction(BranchUniversalObject.BUO_USER_ACTIONS.SHARE_COMPLETED);
                             }
 
                             @Override
                             public void onChannelSelected(String channelName) {
+                                branchUniversalObject.userCompletedAction(BranchUniversalObject.BUO_USER_ACTIONS.SHARE_STARTED);
```
aaustin: Let's move this to dialog launched

aaustin: @ahmednawar - this is a good idea.. adding it to the share start/completed

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
```diff
@@ -80,6 +105,8 @@ public BranchUniversalObject() {
         type_ = "";
         indexMode_ = CONTENT_INDEX_MODE.PUBLIC; // Default content indexing mode is public
         expirationInMilliSec_ = 0L;
+        price_ = 0.0;
+        currency_ = CURRENCY_TYPE.NONE;
```
aaustin: let's default to null

aaustin: for both of these

aaustin: honestly.. I'm not sure why any of those values `canonicalIdentifier`, `canonicalUrl`, etc should be initialized to an empty string?

Branch-SDK/src/io/branch/referral/ServerRequestRegisterView.java
```diff
@@ -106,6 +107,10 @@ private JSONObject createContentViewJson(BranchUniversalObject universalObject,
 
         paramsObj.put(Defines.Jsonkey.ContentKeyWords.getKey(), universalObject.getKeywordsJsonArray());
         paramsObj.put(Defines.Jsonkey.PublicallyIndexable.getKey(), universalObject.isPublicallyIndexable());
+        if (!universalObject.getCurrencyType().equals(CURRENCY_TYPE.NONE.toString())) {
+            paramsObj.put(Defines.Jsonkey.PurchaseAmount.getKey(), universalObject.getPrice());
```
aaustin: I think we should do these separately, and note that we default to USD in currency.

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
EvangelosG: in my history i've initialized vars to empty strings to avoid getters throwing nullpointerexceptions.

Branch-SDK/src/io/branch/referral/ServerRequestRegisterView.java
aaustin: my main objective is that requests from iOS and Android are exactly the same in call edge cases.

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
aaustin: The problem I have with that is as I mention below: it causes the Android SDK to send empty strings for values, where the iOS SDK doesn't send anything, which pollutes the backend and has us store unnecessary data.

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
ahmednawar: that's pretty cool. will add it to iOS

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
sojanpr: The reason for initializing the above variables is that this class is implementing parcellable interface. "Null" is not auto unboxed through parcel and it can throw NPE in case of null (Unless we write more code in parcel logic  to handle null with some other default value).

sojanpr: Will add logic so that value is not set it won't send to Branch

sojanpr: Added better handling for price. Wont be updated to link data or user action payload if not set by user

Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
ahmednawar: P.S This will also populate the events on the Dashboard for the partners and they didn't explicitly call it. There is a chance they get confused.

sojanpr: @ahmednawar This is called only from the testbed app not from SDK. So it wont affect other apps

ahmednawar: Oh, I thought we are adding this to the SDK, no?

aaustin: Oh my bad. I thought so too but that's a good point. However, I think these are some of those events that we could add a ton of value to the dash without being distracting.

aaustin: So let's add them to the share sheet?

aaustin:
looks great - thoughts on auto events? Otherwise 👍

PR #290, opened 2016-09-06
Adding Set Debug Back

sojanpr:
Adding setDebug() Api for supporting debugging on other SDKs
@aaustin


aaustin:
Looks good 👍 This has all of the other stuff in it as well in case you didn't notice.

PR #294, opened 2016-09-10
Cookie Based Matching

sojanpr:
Implementation for cookie based Matching using chrome custom Tab

@derrickstaten @aaustin


Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
```diff
@@ -0,0 +1,145 @@
...
+                if (matcher.find()) {
+                    hostString = matcher.group(0);
+                }
+            }
+
+            if (!TextUtils.isEmpty(hostString)) {
+                String uriString = "";
+                uriString += hostString + "_strong_match?os=" + deviceInfo.getOsName();
+                // Add HW ID
+                uriString += "&" + Defines.Jsonkey.HardwareID.getKey() + "=" + deviceInfo.getHardwareID();
```
aaustin: can you add GAID as well if present?

aaustin:
So, high level, we cannot use the referring link as a flag here. The idea of strong matching is to match with 100% accuracy through _install_. If there's a referring link, clearly it's not an install. We need to figure out how to let the dev specify the domain.

sojanpr:
Should we extend the init method to have an optional domain name in case of
strong match check needed?

aaustin:
I think we should have use app.link by default and let people specify a custom domain if they want. I'd suggest a method that could be called before get auto instance like 'setCookieMatchDomain'.

Branch-SDK-TestBed/build.gradle
```diff
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 dependencies {
-    compile fileTree(dir: 'libs', include: '*.jar')
     compile project(':Branch-SDK')
-    debugCompile 'com.squareup.leakcanary:leakcanary-android:1.+'
-    releaseCompile 'com.squareup.leakcanary:leakcanary-android-no-op:1.3.1'
+    //debugCompile 'com.squareup.leakcanary:leakcanary-android:1.+'
+    //releaseCompile 'com.squareup.leakcanary:leakcanary-android-no-op:1.3.1'
```
aaustin: did you mean to leave these commented out?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -380,6 +380,8 @@
 
     String sessionReferredLink_; // Link which opened this application session if opened by a link click.
 
+    private static String cookieBasedMatchDomain_ = null; // Domain name used for cookie based matching
```
aaustin: So, you're thinking that we don't use a default and that we just ask the dev to specify?

sojanpr: Yes , I feel that way we can cover all cases

aaustin: so, per our discussion, you're updating this to be "app.link".

aaustin:
now there are conflicts

Branch-SDK-TestBed/build.gradle
sojanpr: Yes. Since we are adding more dependency to SDK there is chance  for  dex over flow (Since we don't want to complicate test bed  with multi-dex implementation)  for test bed app. This is needed  only when we run memory leak test.

PR #295, opened 2016-09-11
Branch Events Defines

sojanpr:
Migrating from BUO_USER_ACTIONS to BRANCH_EVENT
 BRANCH_EVENT Defines pre defines user action string which can be used
with Branch#userCompletedAction or buo#userCompletedAction.

@aaustin

### DIFF
--- a/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
+++ b/Branch-SDK-TestBed/src/io/branch/branchandroiddemo/MainActivity.java
@@ -19,7 +19,7 @@ import io.branch.referral.Branch.BranchReferralStateChangedListener;
 import io.branch.referral.BranchError;
 import io.branch.referral.BranchViewHandler;
 import io.branch.referral.SharingHelper;
-import io.branch.referral.util.CURRENCY_TYPE;
+import io.branch.referral.util.CurrencyType;
 import io.branch.referral.util.LinkProperties;
 import io.branch.referral.util.ShareSheetStyle;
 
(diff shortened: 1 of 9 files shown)


aaustin:
Also, I can't seem to find these 3 constants in the current project. Can you add them to Branch event?

```
CANONICAL_ID_LIST: ”$canonical_identifier_list”
PURCHASE_AMOUNT: ”$amount”
PURCHASE_CURRENCY: ”$currency”
```

sojanpr:
Are these constants supposed to be used by devs?

aaustin:
Yes they are

aaustin:
The other thing that feels a bit off is the ALL CAPS classname `BRANCH_EVENT`. Maybe I'm just a perfectionist, but looking at these imports, it looks weird:

```
import io.branch.referral.SharingHelper;
import io.branch.referral.util.BRANCH_EVENT;
import io.branch.referral.util.CURRENCY_TYPE;
import io.branch.referral.util.LinkProperties;
import io.branch.referral.util.ShareSheetStyle;```

Some are caps, some are prefixed with `util`. Let's change `CURRENCY_TYPE` and `BRANCH_EVENT` to `CurrencyType` and `BranchEvent` respectively to be more consistent with the rest of the libraries. It's a bummer that SharingHelper isn't in `utils` but it's too late for that.

sojanpr:
I agree completely with the naming suggestions. Apologies for the misplaced SharingHelper. It was a while back and we didn't have util package at that time

sojanpr:
With the above addition of currency, amount and canonical ID List the class name `BranchEvent` seems not the best. Any suggestion

aaustin:
How about `EventHelper`? and move `CurrencyType` in there as well.

aaustin:
Oh, you added them here. Let's just keep it like this. 👍

PR #297, opened 2016-09-14
Updating min Sdk version

sojanpr:
Updating min SDK version to 15 to use chrome tabs

@EvangelosG


EvangelosG:
👍 makes sense to me!

PR #299, opened 2016-09-15
Stripping Unncessary Dependency

sojanpr:
Strippping support v4 module from 
support:customtabs dependecy

@aaustin @EvangelosG


aaustin:
Looks good to me 👍

PR #300, opened 2016-09-16
Support for Whitelisting path

sojanpr:
Suppress adding schme or a Path for white listing external intent

@aaustin

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -1221,8 +1221,15 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
             try {
                 if (data != null) {
                     if (externalUriWhiteList_.size() > 0) {
-                        String externalIntentScheme = Uri.parse(data.toString()).getScheme();
-                        if (externalIntentScheme != null && externalUriWhiteList_.contains(externalIntentScheme)) {
+                        String uriString = data.toString();
+                        boolean foundWhiteListMatch = false;
+                        for (String whiteListScheme : externalUriWhiteList_) {
+                            if (uriString.contains(whiteListScheme)) {
+                                foundWhiteListMatch = true;
+                                break;
+                            }
+                        }
+                        if (foundWhiteListMatch) {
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1221,8 +1221,15 @@ private boolean readAndStripParam(Uri data, Activity activity) {
             try {
                 if (data != null) {
                     if (externalUriWhiteList_.size() > 0) {
-                        String externalIntentScheme = Uri.parse(data.toString()).getScheme();
-                        if (externalIntentScheme != null && externalUriWhiteList_.contains(externalIntentScheme)) {
+                        String uriString = data.toString();
+                        boolean foundWhiteListMatch = false;
+                        for (String whiteListScheme : externalUriWhiteList_) {
```
TheMarshal99: It's not quite this simple, as we need to automatically whitelist "open" for default behavior, and I have concerns about treating whitelisted schemes and whitelisted paths the same.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1220,25 +1222,41 @@ private boolean readAndStripParam(Uri data, Activity activity) {
...
-                            sessionReferredLink_ = data.toString();
-                            prefHelper_.setExternalIntentUri(data.toString());
+                    } else {
+                        foundSchemeMatch = true;
+                    }
+
+                    if (externalPathWhiteList_.size() > 0) {
+                        String uriPath = data.getPath();
+
+                        for (String whiteListPath : externalPathWhiteList_) {
```
sojanpr: can we simplify this for loop with
`externalUriWhiteList_.contains(data.getScheme())`

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1220,25 +1222,41 @@ private boolean readAndStripParam(Uri data, Activity activity) {
...
+                            if (uriPath.startsWith(whiteListPath)) {
```
sojanpr: We need a null check here since `getPath` can return null

PR #302, opened 2016-09-19
Skip Path

sojanpr:
Changing form Path white list to black list

@TheMarshal99 @aaustin

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -377,7 +377,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
     private boolean isGAParamsFetchInProgress_ = false;
 
     private List<String> externalUriWhiteList_;
-    private List<String> externalPathWhiteList_;
+    private List<String> skipExternalUriPaths_;
 
     String sessionReferredLink_; // Link which opened this application session if opened by a link click.
 
@@ -412,7 +412,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
             intentState_ = INTENT_STATE.READY;
         }
         externalUriWhiteList_ = new ArrayList<>();
-        externalPathWhiteList_ = new ArrayList<>();
+        skipExternalUriPaths_ = new ArrayList<>();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1220,18 +1222,20 @@ private boolean readAndStripParam(Uri data, Activity activity) {
...
-                        String externalIntentScheme = Uri.parse(data.toString()).getScheme();
-                        if (externalIntentScheme != null && externalUriWhiteList_.contains(externalIntentScheme)) {
-                            sessionReferredLink_ = data.toString();
-                            prefHelper_.setExternalIntentUri(data.toString());
-                        }
-                    } else {
+                        foundSchemeMatch = externalUriWhiteList_.contains(data.getScheme());
+                    }
+                    if (skipExternalUriPaths_.size() > 0) {
+                        skipThisPath = skipExternalUriPaths_.contains(data.getPath());
```
TheMarshal99: Gonna have to check the black-listed path against the beginning of the URI path. So a black-listed path of "products" should blacklist "app://products" as well as "app://products/1234". Right now this only blocks the former.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1220,18 +1222,28 @@ private boolean readAndStripParam(Uri data, Activity activity) {
...
-                            sessionReferredLink_ = data.toString();
-                            prefHelper_.setExternalIntentUri(data.toString());
-                        }
+                        foundSchemeMatch = externalUriWhiteList_.contains(data.getScheme());
                     } else {
+                        foundSchemeMatch = true;
+                    }
+                    if (skipExternalUriPaths_.size() > 0) {
+                        for (String path : skipExternalUriPaths_) {
+                            String externalPath = data.getPath();
```
aaronhe42: Actually this didn't solve the problem. Let's say we have the Uri `sth://callback`. The `data.getPath()` would simply return `/`. I guess `data.getHost` should be used?

sojanpr: @aaronhe42 Sorry for the confusion. We have released 2.4.4 which has the option to skip collecting external intent data on specified hot name. Could you please try that.

PR #303, opened 2016-09-20
Switched URI blacklist from 'path' to 'host' per desired functionality.

TheMarshal99:
Previous functionality: blacklist "callback" if URI is of form "tinder://hostname/callback"
Desired functionality: be able to blacklist "callback" if URI is of form "tinder://callback?..."

Updated to function as the latter.


TheMarshal99:
@sojanpr

sojanpr:
LGTM

PR #307, opened 2016-09-20
Adding Trouble shooting section in Read me

sojanpr:
@aaustin @aeromusek @dwestgate


aeromusek:
Content is above my head technically, but seems important to have. I'd vote for trying to minimize what goes into the README here, since the eventual goal is to merge everything into the real docs portal, but I'll leave that call to @dwestgate in his new role :)

PR #309, opened 2016-09-23
Circle CI Fix

sojanpr:
Circle CI fix to update the AVD


aaustin:
Still not passing it seems. Let me know if you need anything

PR #310, opened 2016-09-28
IAE on getting intent extra

sojanpr:
This is really a corner case and not able to reproduce. The crash log
shows the `intent.getExtra` throws illegal argument exception. It seems
like some Android version or device has issue in unmarshaling some
extra values with intent. Adding a crash protection

@aaustin @Sarkar @EvangelosG


aaustin:
Sure? 👍

PR #312, opened 2016-10-02
Temp Transaction Too Large Crash Fix

sojanpr:
Adding a temp fix for Transaction too large issue. This will result in
auto deep link failure on getting this exception.

This need to be replaced with a better fix but involves some serious
changes with the way auto-deep link is configured in manifest

@aaustin @EvangelosG


aaustin:
Did you mean to make this PR against staging? The first change I already reviewed but the second looks ok too 👍

PR #314, opened 2016-10-05
Support for forced session restart

sojanpr:
Implementation  for forced session restart. The Branch session is
restarted if the incoming intent has $branch_force_new_session set to
true.
This is for supporting opening a deep link path while app is
already running in the foreground. Such as clicking push notification
while app in foreground.

@aaustin @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -2199,19 +2199,22 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
                 if (BranchUtil.isTestModeEnabled(context_)) {
                     prefHelper_.setExternDebug();
                 }
-
-
-                Uri intentData = null;
-                if (activity.getIntent() != null) {
-                    intentData = activity.getIntent().getData();
-                }
-                initSessionWithData(intentData, activity); // indicate  starting of session.
+                startSession(activity);
+            } else if (checkIntentForSessionRestart(activity.getIntent())) { // Case of opening the app by clicking a push notification while app is in foreground
+                initState_ = SESSION_STATE.UNINITIALISED;
+                // no need call close here since it is session forced restart. Don't want to wait till close finish
+                startSession(activity);
             }
             activityCnt_++;
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2250,6 +2253,30 @@ public void onActivityDestroyed(Activity activity) {
...
+
+    /*
+     * Check for forced session restart. The Branch session is restarted if the incoming intent has $branch_force_new_session set to true.
+     * This is for supporting opening a deep link path while app is already running in the foreground. Such as clicking push notification while app in foreground.
+     *
+     */
+    private boolean checkIntentForSessionRestart(Intent intent) {
+        boolean isRestartSessionRequested = false;
+        if (intent != null) {
+            isRestartSessionRequested = intent.getBooleanExtra(Defines.Jsonkey.ForceNewBranchSession.getKey(), false);
```
aaustin: If the use cloud push, can they still modify the intent parameters? Right now, we tell devs to put the key "branch" in the payload of their push, and the link URL as the value. Is this an additional parameter they'd need to add to the payload?

aaustin:
Overall, why do they need to add a new boolean? Shouldn't we just check if the `branch` key is present in the push payload and automatically do this by default? This seems to overcomplicate things quite a bit unless I misunderstand.

Branch-SDK/src/io/branch/referral/Defines.java
```diff
@@ -69,6 +69,7 @@
         AndroidAppLinkURL("android_app_link_url"),
         AndroidPushNotificationKey("branch"),
         AndroidPushIdentifier("push_identifier"),
+        ForceNewBranchSession("$branch_force_new_session"),
```
sojanpr: Thanks for catching it . Will remove  the `$`

sojanpr:
The reason for adding a new flag is to avoid any Branch behavior  changes for existing apps.  Depending  only on `Branch` key in the intent will have ill effect on applications checking for the session params using `getLatestPrams`

aaustin:
alright sounds good 👍

PR #315, opened 2016-10-06
Adding locale string to device params for matching check

sojanpr:
Locate string is added to device params in branch request .A string
representing the locale in format "ISO2Language-ISO2Country" (Eg.
"ml-IN"). An empty string is returned on error.  Locale string is sent
with key `locale`

@dmitrig01 @aaustin @E-B-Smith


dmitrig01:
can we just send two different fields, 'language' and 'country'?

sojanpr:
Added `country` and `language` (Both are ISO2 Codes)

Branch-SDK/src/io/branch/referral/SystemObserver.java
```diff
@@ -322,6 +323,33 @@ public String getPhoneModel() {
...
     /**
+     * Gets default ISO2 Country code
+     *
+     * @return A string representing the ISO2 Country code (eg US, IN)
+     */
+    public String getISO2CountryCode() {
+        if (Locale.getDefault() != null) {
+            return Locale.getDefault().getCountry();
+        } else {
+            return "";
```
aaustin: Is there a case where these return an empty string? Or is this just to be safe. I don't think we should get in the habit of sending empty strings in the POST body.

aaustin:
👍

PR #317, opened 2016-10-21
Adding Local IP  for matching

sojanpr:
Adds the local ipv4 address for  non local net Interface with other
device params for more matching checks.

@dmitrig01 @aaustin @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/SystemObserver.java
+++ b/Branch-SDK/src/io/branch/referral/SystemObserver.java
@@ -24,6 +24,9 @@ import org.json.JSONObject;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Method;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


Branch-SDK/src/io/branch/referral/SystemObserver.java
```diff
@@ -641,4 +644,31 @@ protected void onPostExecute(Void aVoid) {
...
+     */
+    public static String getLocalIPAddress() {
+        String ipAddress = "";
+        try {
+            List<NetworkInterface> netInterfaces = Collections.list(NetworkInterface.getNetworkInterfaces());
+            for (NetworkInterface netInterface : netInterfaces) {
+                List<InetAddress> addresses = Collections.list(netInterface.getInetAddresses());
+                for (InetAddress address : addresses) {
+                    if (!address.isLoopbackAddress()) {
+                        String ip = address.getHostAddress();
```
aaustin: I guess you're using the try catch to avoid this, but can this return null?

sojanpr: Not only for that but I suspect a permission denial on some device with future android versions. As per adndroid policy ip is not allowed to access unless providing network permissions

PR #318, opened 2016-10-21
Capture intent extra only for white listed hosts

sojanpr:
Capture the intent extra  for any whitelisted hosts.

@EvangelosG @aaustin


EvangelosG:
@sojanpr the code makes sense to me + compiles, how would you like me to test?

aaustin:
👍

aaronhe42:
May I know the reason why intent extras even need to be collected? Does any functionality rely on the data passed into the intent? I'd assume the `Uri` is probably something Branch only cares about?

sojanpr:
@aaronhe42 We are actually on another app-app linking platform which can help dev's to connect to other apps eazily using Branch. Intent extras are collected for this reason. You can see this open source project at https://github.com/BranchMetrics/Roots-Android

PR #321, opened 2016-10-24
Fixing Activity life cycle re init issue

sojanpr:
`getAutoInstance()` or `getAutoTestInstance()` are supposed to be
called only from `Application#onCreate()`. After this Branch instance
can be accessed anywhere using `getInstance()` method.  This fix the
issue of  reinitializing the activity life cycle observer if
`getAutoInstance()` is called  to get instance accidentally which will
lead to issues with Branch session management
@aaustin @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -506,14 +506,15 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
                 branchReferral_.linkCache_.clear();
                 branchReferral_.requestQueue_.clear();
             }
-        }
-        branchReferral_.context_ = context.getApplicationContext();
 
-        /* If {@link Application} is instantiated register for activity life cycle events. */
-        if (context instanceof BranchApp) {
-            isAutoSessionMode_ = true;
-            branchReferral_.setActivityLifeCycleObserver((Application) context);
+            branchReferral_.context_ = context.getApplicationContext();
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -506,15 +506,16 @@ private static Branch getBranchInstance(@NonNull Context context, boolean isLive
...
-        }
-        branchReferral_.context_ = context.getApplicationContext();
 
-        /* If {@link Application} is instantiated register for activity life cycle events. */
-        if (context instanceof BranchApp) {
-            isAutoSessionMode_ = true;
-            branchReferral_.setActivityLifeCycleObserver((Application) context);
+            branchReferral_.context_ = context.getApplicationContext();
+            /* If {@link Application} is instantiated register for activity life cycle events. */
+            if (context instanceof BranchApp) {
```
aaustin: won't this mean that `setActivityLifeCycleObserver` is only called if they've subclassed our Application? That's not recommended as a standard flow: https://dev.branch.io/getting-started/sdk-integration-guide/guide/android/#enable-auto-session-management

aaustin: Oops - this line: https://github.com/BranchMetrics/android-branch-deep-linking/pull/321/files#diff-11fc8f6026de99b60216e9aca37d6d26L564

aaustin:
👍

PR #322, opened 2016-10-25
Fix: Deep link params are not supplied on cold start with poor internet connection

sojanpr:
If the app is put to background before   receiving the  init response
close is not called (Since init is not happened yet. ) When init
response received finally SDK mark session as valid.
Upon next link click start since SDK thinks session started it returned
empty param

It is only reproducible  on a slower network. need to put app into
background before init finished.

Here is  customer reported issue on this
https://github.com/BranchMetrics/android-branch-deep-linking/issues/319

@aaustin @EvangelosG

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -2200,6 +2200,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
                 }
             }
             if (activityCnt_ < 1) { // Check if this is the first Activity.If so start a session.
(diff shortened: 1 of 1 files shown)


aaustin:
Seems harmless to me 👍

EvangelosG:
👍 Data is returned for link_1 when app is in the background. I click link_2, app opens and data for link2 is returned. In both cases correct data is delivered, even with Sleep(5000) to ensure I'm "slowing" my internet down.

pikaboo:
We are experiencing this issue *some* of the time. What can we do besides updating to version 2.4.6 or 2.4.7 which cause an even greater problem of init callback never returning

PR #325, opened 2016-10-31
enable disable logging

EvangelosG:
@sojanpr 
added static 
void enableLogging, void disableLogging, boolean isLogging, calls so you can set before Branch initializes.

known: if you disable logging with disableLogging(), the app will still log your next close() call, but not your next open, so the session still finishes up.


sojanpr:
Do we really need the `build.gradle` file changes and the `gradle.warpper` changes?

PR #333, opened 2016-11-07
Handle status 400

sojanpr:
Handle status 400  in Branch error
@aaustin @EvangelosG


aaustin:
You confirm that this correction is the same logic as iOS? If so 👍

PR #334, opened 2016-11-07
Handling exception proguard obfuscated for Facebook SDK

sojanpr:
Adding exception handling and Read me update

@aaustin @EvangelosG @rubinsingh


EvangelosG:
👍 (For anyone wondering about catch exception vs catch throwable: http://stackoverflow.com/questions/2274102/difference-between-using-throwable-and-exception-in-a-try-catch)

aaustin:
👍

PR #340, opened 2016-11-10
Handling for API level issue with some android device

sojanpr:
Some devices seem to give a wrong os version   number or not providing
features as per the SDK level. Adding a safe handling

@aaustin @EvangelosG


aaustin:
👍

PR #344, opened 2016-11-15
Adding more options to Preferred Sharing Option

ryee1:
Added pinterest, gmail, wechat, instagram, hangout


sojanpr:
@ryee1  Thank you very much for making share sheet better! Will add this is to our next release

PR #348, opened 2016-11-16
Fixing issue : callback never returned when called from onCreate method

sojanpr:
Support init from oncreate 
We see few partners prefer to still call ini session form onCreate(). This was broken with fix for

https://github.com/BranchMetrics/android-branch-deep-linking/issues/319.

@aaustin @EvangelosG


EvangelosG:
this fix works for me. 👍

PR #351, opened 2016-11-22
Fix bool representation compatibility issue Android -iOS

sojanpr:
iOS - Android  compatibility issue. iOS send 0/1 instead of true or
false. Adding handling for fields which are parsed

@aaustin @EvangelosG @E-B-Smith


aaustin:
👍

PR #356, opened 2016-12-04
Default check Facebook App Link local to false

aaustin:
@EvangelosG For some reason, this was defaulted to true. Since there also appears to be a bug with Facebook's SDK, it's causing trouble for partners. Separately, I'll file a Facebook bug for the deferred deep link API.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -284,7 +284,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
 
     private static boolean disableDeviceIDFetch_;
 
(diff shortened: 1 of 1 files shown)


EvangelosG:
Interesting. Here's the commit message w/ the introduction of true. 
screen shot 2016-12-03 at 5 25 50 pm But the previous commit messages says it was introduced to disable the FB app link check by default.

aaustin:
Must have been a mistake. It's not used anywhere else.

sojanpr:
This is definitely a mistake happened with some previous check in

EvangelosG:
kk 👍

EvangelosG:
closing, created a new PR against staging: https://github.com/BranchMetrics/android-branch-deep-linking/pull/357

PR #357, opened 2016-12-05
set fbapplinkcheck to false

EvangelosG:


EvangelosG:
created PR against staging, instead of master

PR #359, opened 2016-12-06
cleaning up setdebug

EvangelosG:
set debug should be static, also it really shouldn’t be called set
debug at this point if the manifest variable we use is test mode? i
added static enabletestmode/disabletestmode, and kept setdebug which
calls enabletestmode. or i can go the other direction and choose to use
the name set debug to mirror iOS.


EvangelosG:
@aaustin review please

aaustin:
Sure - this is rather frivolous though. Have there been requests for this?

EvangelosG:
No, I'm just creating PRs here and there to better understand our SDK as a part of my transition process. Some will be meaningful, some frivolous. :)

aaustin:
👍

PR #363, opened 2016-12-13
Func to init Branch properly programmatically

EvangelosG:
@Sarkar func to init branch from app class as follows: Branch.getAutoInstance(this, "key");

No need for anything in the manifest regarding key(s).

Can you gimme a test?

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -601,6 +601,38 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
     }
 
     /**
+     * <p>Singleton method to return the pre-initialised, or newly initialise and return, a singleton
+     * object of the type {@link Branch}.</p>
+     *
+     * <p>Use this whenever you need to call a method directly on the {@link Branch} object.</p>
+     *
+     * @param context A {@link Context} from which this call was made.
+     * @param branchKey A {@link String} value used to initialize Branch.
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -601,6 +601,38 @@ public static Branch getAutoInstance(@NonNull Context context, boolean isReferra
...
+     * one was not already requested during the current app lifecycle.
+     */
+    @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+    public static Branch getAutoInstance(@NonNull Context context, String branchKey) {
+        isAutoSessionMode_ = true;
+        customReferrableSettings_ = CUSTOM_REFERRABLE_SETTINGS.USE_DEFAULT;
+        boolean isLive = !BranchUtil.isTestModeEnabled(context);
+        getBranchInstance(context, isLive);
+
+        if (branchKey.startsWith("key_")) {
```
Sarkar: Would be safe to have a check for null / empty here since it's a string

EvangelosG: good call

aaustin:
👍

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: This context should be an Application instance. I think it would be better to specify that in the comment

PR #371, opened 2016-12-26
Modified Listener for further channel customization

ryee1:
Example use case: Facebook/twitter/pinterest needs different aspect ratio images to show display correctly without cropping. With the new changes, we are able to set the image url of the short link builder class depending on which share channel the user selected


sojanpr:
We will be adding this as part of upcoming revamping to the share sheet

PR #374, opened 2017-01-06
share sheet header item

EvangelosG:
disable the sharesheet header item from being clickable, which was causing crashes. @sojanpr -- when creating a share sheet, if you include `.setSharingTitle("share using");` a clickable header item is added to the sharesheet that crashes when clicked. I changed the constructor being used to include isSelectable = false when adding the header item. Please let me know if this is the right course of action.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
+++ b/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
@@ -157,7 +157,7 @@ class ShareLinkManager {
         shareOptionListView.setBackgroundColor(Color.WHITE);
 
         if (builder_.getSharingTitleView() != null) {
-            shareOptionListView.addHeaderView(builder_.getSharingTitleView());
+            shareOptionListView.addHeaderView(builder_.getSharingTitleView(), null, false);
         } else if (!TextUtils.isEmpty(builder_.getSharingTitle())) {
             TextView textView = new TextView(context_);
             textView.setText(builder_.getSharingTitle());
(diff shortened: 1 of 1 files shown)


EvangelosG:
@Sarkar could use a review if you have time today

Branch-SDK/src/io/branch/referral/ShareLinkManager.java
```diff
@@ -166,7 +166,7 @@ private void createShareDialog(List<SharingHelper.SHARE_WITH> preferredOptions)
             textView.setTextAppearance(context_, android.R.style.TextAppearance_Medium);
             textView.setTextColor(context_.getResources().getColor(android.R.color.darker_gray));
             textView.setPadding(leftMargin, padding, padding, padding);
-            shareOptionListView.addHeaderView(textView);
+            shareOptionListView.addHeaderView(textView, null, false);
```
sojanpr: Instead of changing this constructor can we do 
`textView.setClickable(false);`
Please note that this constructor is used in user defined Title views also where they control the behavior of short title (Line # 160)

EvangelosG: That was the first thing I tried, and the crash persists, try it:
```
textView.setClickable(false);
shareOptionListView.addHeaderView(textView);
//shareOptionListView.addHeaderView(textView, null, false);
```

sojanpr: Sorry it is my bad. Didn't realize the list header is selectable  by default. The above change look good but we need to do the same on Line #160

EvangelosG: done!

sojanpr:
LGTM 👍

PR #375, opened 2017-01-11
Skip strong match on no branchkey

sojanpr:
Skip calling strong match end point when Branch key is not specified.

@EvangelosG @dmitrig01 @aaustin


EvangelosG:
👍

PR #379, opened 2017-01-16
commerce update

EvangelosG:
removing default values for smaller event payloads @derrickstaten

### DIFF
--- a/Branch-SDK/src/io/branch/referral/util/CommerceEvent.java
+++ b/Branch-SDK/src/io/branch/referral/util/CommerceEvent.java
@@ -122,7 +122,7 @@ public class CommerceEvent {
             jsonObject.put("tax", this.tax);
             jsonObject.put("coupon", this.coupon);
             jsonObject.put("affiliation", this.affiliation);
-            try { jsonObject.put("products", getProducts()); } catch ( NullPointerException e ) {}
+            if ( getProducts() != null ) jsonObject.put("products", getProducts());
         } catch (JSONException e) {
 
         }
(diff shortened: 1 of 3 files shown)


EvangelosG:
@sojanpr -- a PR for Derrick, updating commerceEvent for a simplified function without required empty null statements, default constructors cleaned out, and setproductcategory replaced w/ setcategory.

Branch-SDK/src/io/branch/referral/util/CommerceEvent.java
```diff
@@ -137,6 +131,7 @@ public JSONObject getCommerceJSONObject() {
     }
 
     public List<JSONObject> getProducts() {
+        if (this.products == null) return null;
```
sojanpr: should we return null here? Can we return an empty List in order to prevent any accidental  NPE

sojanpr: line break after if

EvangelosG: I returned null there because @derrickstaten doesn't want empty items in the payload, and this will return an empty product item.

derrickstaten: Can we make this line safe instead: https://github.com/BranchMetrics/android-branch-deep-linking/blob/21ede340d9e03d99f4df68a85609b6774c5730ac/Branch-SDK/src/io/branch/referral/util/CommerceEvent.java#L125

derrickstaten: That should be enough to prevent any NPE

PR #383, opened 2017-01-20
Fixing a possible crash

sojanpr:
This fix is to wrap a possible crash when the resource cannot be
figured out with the ID. Also fixing an issue related to view iteration
logic.

@aaustin @EvangelosG @shekar9


EvangelosG:
👍

aaustin:
yuck 👍

PR #385, opened 2017-01-21
Fix proguard warnings that halt build

mikemee:
I had the same problem described here: 

https://mobilegrowth.org/questions/question/gradle-proguard-warning-with-exclude-module-answers-shim-option/

This PR has the fix proposed there, which also fixed my (identical) problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Just add `compile 'io.branch.sdk.android:library:2.+'` to the dependencies secti
 - If you don't plan to use the `Fabric Answers` integration, and don't want to import the `answers-shim`, just import your project as follows:
 ```
 compile ('io.branch.sdk.android:library:2.+') {
-  exclude module: 'answers-shim'
+  exclude module: 'answers.shim'
 }
 ```
 - This supports minimum sdk level 15. If you want to support minimum sdk level 9 please consider using version 1.14.5


sojanpr:
@mikemee Thanks for the PR. Could you please help to  recreate and check the warning in your logs. Which version of gradle build tools you are  using?
Following is the additional dependency that we add to the  SDK
`compile 'com.crashlytics.sdk.android:answers-shim:0.0.6'` . Just wondering how `answers.shim` is working not `answers-shim`

mikemee:
@sojanpr I think it's because the class name that proguard is looking for is `com.crashlytics.android.answers.shim`. It makes sense to me that `answers-shim` won't work as a proguard instruction, because it needs to be a valid portion of class name. _Though I hardly claim to be a proguard expert!)_

I.e. the `-` vs `.` is the only problem here...

I'm using the Android gradle plug version 2.2.3, though I think it was failing with 2.2.2. (`com.android.tools.build:gradle:2.2.3`)

EvangelosG:
Hi @mikemee -- I've included Branch as follows:

```
compile ('io.branch.sdk.android:library:2.+') {
        exclude module: 'answers-shim'
}
```

& Enabled proguard for my debug build:
`proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'`

What else must I do to reproduce your issue?

mikemee:
Hi @EvangelosG - sorry this is turning into a snark hunt!

Hmm, are you also including the lines suggested in the readme in your `proguard-rules.pro` per https://github.com/BranchMetrics/android-branch-deep-linking#proguard-settings-for-leveraging-branchs-pooled-matching

I wonder if my proguard config is set to "fail on warning" and the default is not (though I don't see a setting for that with a quick look).  Fwiw, here is an extract from my `build.config`:

```
    buildTypes {
        release {
            minifyEnabled true
            shrinkResources true
            signingConfig signingConfigs.release
            proguardFile 'proguard-google-play-services.pro'
            proguardFile 'proguard-support-v7-appcompat.pro'
            proguardFile 'proguard-square-picasso.pro'
            proguardFile 'proguard-square-retrofit.pro'
            proguardFile 'proguard-gson.pro'
            proguardFile 'proguard-project.pro'
            proguardFile 'proguard-amazon.pro'
            proguardFile 'proguard-realm.pro'
            proguardFile 'proguard-branch-io.pro'
            proguardFiles getDefaultProguardFile('proguard-android.txt')
        }
        debug {
            ext.enableCrashlytics = false
        }
    }
```
I'll dig a bit more to see if there are more clues.

At the risk of stating the obvious, and/or looking silly myself!, given the errors [from here](https://mobilegrowth.org/questions/question/gradle-proguard-warning-with-exclude-module-answers-shim-option/) that look like this:
```
Warning: io.branch.referral.ExtendedAnswerProvider: can’t find referenced class com.crashlytics.android.answers.shim.
```
you'll probably need your test case app to include `crashlytics` (from `fabric.io`). Mine does anyway. Perhaps that's a factor?

mikemee:
FYI, I include branch in my `build.gradle` as follows (using `answers-shim` causes the proguard warning that stops the build):
```
    compile ('io.branch.sdk.android:library:2.5.6') {
        exclude module: 'answers.shim'
    }
```
_(I use the specific version number so lint doesn't complain.)_

sojanpr:
Tested and seem like both `answers-shim` and `answers.shim` works fine.  Since lot of our partners are already using `answers-shim` and have no issues , can we keep this unchanged and mention this in the trouble shooting session of  README

sojanpr:
Updated Read me with the trouble shooting tip
https://github.com/BranchMetrics/android-branch-deep-linking/commit/11a81f799296cf9d6968e05c1e03a23b720ef12a

PR #399, opened 2017-02-17
Enabling delayed intents for all sdk versions

sojanpr:
Since delayed intents  handling  seem more reliable and not causing any
integration issues adding this to all supported SDK versions.
This will make single route for all `StartSession` process .
This will fix the issue of Strong match url not being kicked from pre M
device

@EvangelosG


EvangelosG:
👍

PR #403, opened 2017-02-28
Branch Search 2

sojanpr:
Refactoring to add new search interfaces

@EvangelosG @zeeshabnc


Branch-SDK/src/io/branch/search/DeleteFromSearch.java
```diff
@@ -0,0 +1,35 @@
...
+import io.branch.referral.Branch;
+
+/**
+ * Created by sojanpr on 3/2/17.
+ * <p>
+ *     Provides methods to Delete the contents added to Local search
+ *     @see {@link BranchUniversalObject#listOnSamsungSearch()} and {@link SearchBuilder}
+ * </p>
+ */
+public class DeleteFromSearch {
```
zeeshabnc: Where does this get used??

Branch-SDK/src/io/branch/search/RecommendationBuilder.java
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Add the given app to the app preference list. Caller can add any number of preferred apps
+     * The recommended content list will include only contents from the preferred apps.
+     * If no apps added, the content recommendations are left unfiltered with apps
+     *
+     * @param packageName {@link String} package name for the apps to be added to the preferred list
+     * @return RecommendationBuilder instance for chaining
+     */
+    public RecommendationBuilder addPreferredApp(String packageName) {
```
zeeshabnc: Whats the usecase for preferred apps?

Branch-SDK/src/io/branch/indexing/BranchUniversalObject.java
```diff
@@ -925,11 +926,69 @@ public void onChannelSelected(String channelName) {
...
+     *
+     * @param userAction A {@link String }with value of user action name. See {@link io.branch.referral.util.BranchEvent} for Branch defined user events.
+     * @return {@code true} if content successfully added to samsung search
+     */
+    public boolean listOnSamsungSearch(String userAction) {
+        boolean isContentAdded;
+        Context context = Branch.getInstance().getAppContext();
+        String url = getShortUrl(context, new LinkProperties().setChannel("Branch Search"));
+        isContentAdded = BranchSearchServiceConnection.getInstance().addToSharableContent(this, context.getPackageName(), url);
+        if (!TextUtils.isEmpty(userAction)) {
```
sojanpr: These are the service connection methods and service connection make call to aidl. Agree that it is good to have same names

Branch-SDK/src/io/branch/search/DeleteFromSearch.java
sojanpr: This class is to hold the deleteFromIndex and deleteAllFromIndex Apis. Don't wanted to add that functionality to Branch.java
usage would be
`DeleteFromSearch.deleteFromSamsungSearch()`

Branch-SDK/src/io/branch/search/RecommendationBuilder.java
sojanpr: preferred apps provide a a filtering for results in recommendation. If preference added only  contents from those apps will appear in th recomemndation

Branch-SDK/src/io/branch/search/SearchBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+        this.maxContentResults = maxContentResults;
+        return this;
+    }
+
+    /**
+     * Add the preferred  apps to search with. Setting this option will return only result associated with the specified app
+     * @param packageName package name for the preferred app
+     * @return SearchBuilder instance for  chaining
+     */
+    public SearchBuilder addPreferredApp(String packageName) {
```
zeeshabnc: I don't think we have a usecase for this right now, so lets not add it just yet. We are not providing the ability to search just within specific apps, so lets add it only when we want to document / support the use case. But useful to know that we can support it out of the box

Branch-SDK/src/io/branch/search/RecommendationBuilder.java
zeeshabnc: From a recommendation standpoint, lets not add this usecase. We should just power recommendations based on what we think is most useful. When we have a solid usecase for this, we can add it in then. Lets keep the API to the small subset of what we think is a valuable value prop right now.

Branch-SDK/src/io/branch/search/SearchBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Created by sojanpr on 3/1/17.
+ * <p>
+ * Builder for Branch local search query. Use this builder to customise your search for contents
+ * with Branch search
+ * </p>
+ */
+public class SearchBuilder implements Parcelable {
+    private static final int DEF_MAX_APP_RESULT = 3;
```
zeeshabnc: Lets have 4 apps as default?

PR #405, opened 2017-03-07
Avoid race condition

mateherber:
Fetch WeakReference value to local variable to avoid race condition which ends up in npe.


EvangelosG:
Hi @mateherber -- how is this different from the line you removed?

`Context context = (currentActivityReference_ != null && currentActivityReference_.get() != null) ? currentActivityReference_.get().getApplicationContext() : null;`

Let me know!

mateherber:
Hi @EvangelosG!

In the second case after **currentActivityReference_.get() != null** and before **currentActivityReference_.get().getApplicationContext()** weakreference can be dismissed so second get can be null even if the nullcheck passed just before. If you cache it in a local variable (hold the strong reference) it is not the case

sojanpr:
@mateherber  Thanks for mentioning it. Since this method is executed on Main thread do you think there is chance for Weak reference of Activity can change before and after of the  `if` statement?

mateherber:
@sojanpr We're getting nullptr exceptions using this version right there in production.

sojanpr:
@mateherber  Thanks for the  update. Even though couldn't find the way for this to happen, agree with you that above is the best practice when dealing with weak references. We will add this with our next release

PR #423, opened 2017-04-17
Support for Full app conversion attribution and deep linking

sojanpr:
Support for Full app conversion attribution and deep linking
@aaustin @EvangelosG @Sarkar @dmitrig01 @derrickstaten

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -291,8 +291,11 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
 
     private static boolean isLogging_ = false;
 
-    private static boolean checkInstallReferrer_ = false;
-    private static long PLAYSTORE_REFERRAL_FETCH_WAIT_FOR = 5000;
+    // Branch will wait for 1.5 sec to capture the install referrer inorder to provide
+    // accurate attribution . This will be only on the first time user open the app.
+    // @see to skip this please call {@link #disablePlayStoreReferrer() }
+    private static boolean checkInstallReferrer_ = true;
+    private static long PLAYSTORE_REFERRAL_FETCH_WAIT_FOR = 1500;
 
     /**
      * <p>A {@link Branch} object that is instantiated on init and holds the singleton instance of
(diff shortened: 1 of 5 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -291,8 +291,11 @@
 
     private static boolean isLogging_ = false;
 
-    private static boolean checkInstallReferrer_ = false;
-    private static long PLAYSTORE_REFERRAL_FETCH_WAIT_FOR = 5000;
+    // Branch will wait for 1.5 sec to capture the install referrer inorder to provide
+    // accurate attribution . This will be only on the first time user open the app.
+    // @see to skip this please call {@link #disablePlayStoreReferrer() }
+    private static boolean checkInstallReferrer_ = true;
```
Sarkar: my only concern with this PR is setting this default to true, what are your thoughts @sojanpr

EvangelosG: I'm conflicted as well @Sarkar -- it's probably one of those things devs/users are never going to notice, but as a performance queen I would not want *any* SDK by default taking 1.5 seconds extra to gather something that 1) might not show up in time (1.5 seconds seems like it won't?) 2) I might not care about.

sojanpr: This is only for Install  and that happens only one time. Subsequent open will not have  this delay. We have option to reset it for any one who is worried  about this deal over accurate tracking.

EvangelosG: Yeah I saw it's only for installs, that's when you want the user to have the least delayed experience possible IMO. In my metric lacking opinion, playstore referrer information is a less utilized/more advanced feature(?). If you think **every** dev can benefit from this feature then it should be on, but our on-boarding/integration docs will have to cover it, I can already see the tickets in support "why does my app take longer to boot up on install???? I only want deep linking, why does it have to wait for something I'M NOT USING????!?!"

derrickstaten:
Per my comment in slack, it'd be great to have the Android and iOS functionality match up. On iOS, partners can specify the delay themselves (it's not fixed to 1.5 seconds). See https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/Branch.h#L519

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -459,6 +459,9 @@ public void setDebug() {
         enableTestMode();
     }
 
+    // Since play store referrer broadcast from google is few millisecond delayed, delay Branch init by calling this method for more accurate
+    // tracking and attribution. This will be only on the first time user open the app.
+    // Note : This method should be called in order to handle IA - Full app conversion deeplinking and attribution
```
aaustin: Let's add another note and please update with correctness:
`// Note: Recommend 1500 to capture more than 90% of the install referrer cases per our testing as of 4/2017`

aaustin:
Almost there. Let's just add the guidance on number to wait for

aaustin:
Looks good 👍

PR #424, opened 2017-04-19
Updating environment variable value

sojanpr:
Values are “INSTANT_APP”, and “FULL_APP”.
Note :Back end should provide backward compatibility to previous values “instant_app”, “native”

@derrickstaten @EvangelosG


EvangelosG:
👍

PR #426, opened 2017-04-26
Scroll view content discovery

sojanpr:
Adding scroll event capture to trigger scrape
Adding Max Discovery repeat time (mdr). Default is set to 15. This
is a variable which can control the max scrape scrape per path.
(Repeated discovery happens when DRI or ENABLE_SCROLL_WATCH is enabled)

Adding switch to enable scroll based trigger for list view
discovery.bnc_esw, Enables scroll watch for collection views.
Provided as part of Collection view data JSON

The new data structure for content keys and content data are updated in the below document along with the new control keys

@aaustin @EvangelosG @E-B-Smith


sojanpr:
moved to PR 427

PR #429, opened 2017-05-02
Adding support for Fragmented URLs

sojanpr:
Support for URLs with fragments. Fragments need to be last part of the
URL. Adding Branch query params before the fragments

Test Link  :https://bnctestbed.test-app.link/Q1cJyzAiPC#.456

Steps: 
1) Open testbed app by clicking above app link (App should show auto deeplink page)

2)  Move back to MainActivity and background the app

3) Resume the app (App should not navigate to auto deep link activity)

@aaustin @EvangelosG @mparul


aaustin:
looks good 👍

PR #432, opened 2017-05-10
enablesimulateinstalls / testmode GAID fix

EvangelosG:
+is_first_session should always be true when using TestMode or enableSimulateInstalls, but it won't be if we're passing up GAID. This checks to see if TestMode is enabled before passing GAID.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
+++ b/Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
@@ -35,6 +35,7 @@ class BranchStrongMatchHelper {
 
     private boolean isCustomTabsAvailable_ = true;
     boolean isStrongMatchUrlLaunched = false;
+    private Context context_;
 
     Class<?> CustomTabsClientClass;
     Class<?> CustomServiceTabConnectionClass;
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
```diff
@@ -164,7 +166,7 @@ private Uri buildStrongMatchUrl(String matchDomain, DeviceInfo deviceInfo, PrefH
             String hardwareIDTypeVal = deviceInfo.isHardwareIDReal() ? Defines.Jsonkey.HardwareIDTypeVendor.getKey() : Defines.Jsonkey.HardwareIDTypeRandom.getKey();
             uriString += "&" + Defines.Jsonkey.HardwareIDType.getKey() + "=" + hardwareIDTypeVal;
             // Add GAID if available
-            if (systemObserver.GAIDString_ != null) {
+            if (systemObserver.GAIDString_ != null && !BranchUtil.isTestModeEnabled(context_)) {
```
sojanpr: Please pass `context` to `buildStrongMatchUrl()` method instead of creating a class member variable

sojanpr:
Test Plan
A) Install test
1. Gogole play dependency added to testbed app (To collect GAID) 
2. Set Debug = true in manifest
3. Uninstall and install test bed
4. v1/install should return `is_first_session` as `true` always

B) Cookie based Matching test
1. Cookie based strong match should not work when test mode is enabled.
2. Match guaranteed should return `false` even if `Cookie based matching ` enabled 

@mparul ^

sojanpr:
LGTM 👍

PR #437, opened 2017-05-24
fix: int to Integer so you can pass nulls

ethanneff:
Declared right as a property [line 14](https://github.com/BranchMetrics/android-branch-deep-linking/blob/f92c238846c46614bf1f8f3a199c95d58f0f414a/Branch-SDK/src/io/branch/referral/util/Product.java#L14)

Declared wrong as constructor [line 78](https://github.com/BranchMetrics/android-branch-deep-linking/pull/437/files)

Need to pass nulls, example:
screen shot 2017-05-24 at 9 59 58 am

### DIFF
--- a/Branch-SDK/src/io/branch/referral/util/Product.java
+++ b/Branch-SDK/src/io/branch/referral/util/Product.java
@@ -75,7 +75,7 @@ public class Product {
     public Product() {
     }
 
-    public Product(String sku, String name, Double price, int quantity, String brand, String variant, ProductCategory category) {
+    public Product(String sku, String name, Double price, Integer quantity, String brand, String variant, ProductCategory category) {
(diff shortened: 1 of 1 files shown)


EvangelosG:
@derrickstaten is it OK for quantity to be passed as null?

derrickstaten:
Yeah and if not we should make it okay.

derrickstaten:
Yes, let me know if you run into any issues during testing.

EvangelosG:
thanks @derrickstaten . hey @ethanneff 👍

sojanpr:
Added to commit https://github.com/BranchMetrics/android-branch-deep-linking/commit/e330ec09585e44a45182b528b4feeaaec5b2ad34

PR #438, opened 2017-05-25
fix: lower to upper case to match other Product Categories

ethanneff:


ethanneff:
@derrickstaten can you confirm these should be uppercase?

derrickstaten:
@ethanneff are you seeing problems with these values, or doing cleanup? This is an enum on the backend so I believe casing does not matter. If you have a reproducible issue, please let me know ASAP. Thanks!

ethanneff:
@derrickstaten I recommend we uppercase the value to make Android consistent with iOS 
screen shot 2017-05-25 at 4 28 30 pm

derrickstaten:
👍

sojanpr:
Changes added to this commit https://github.com/BranchMetrics/android-branch-deep-linking/commit/3e25ab1bca93c3afaa00d8cd724d4d467d82f6cc

PR #443, opened 2017-06-01
Network Abstraction

sojanpr:
Adding an abstract layer for Branch Remote interface. This will allows
developers to create their own network layer and handle the network
communication in their own way.

By default Branch provides Implementation for
1) HttpUrlConnection
or
2) OKHttp (If dependency is added to the project)

Also this has a lot of clean up and arch changes

@aaustin @EvangelosG


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -405,6 +406,8 @@
     /* Flag for checking of Strong matching is waiting on GAID fetch */
     private boolean performCookieBasedStrongMatchingOnGAIDAvailable = false;
 
+    public static final String SDK_VERSION = "2.8.0";
```
aaustin: Don't we have another place to store these constants?

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2631,7 +2645,9 @@ private boolean checkIntentForSessionRestart(Intent intent) {
     private class getShortLinkTask extends AsyncTask<ServerRequest, Void, ServerResponse> {
         @Override
         protected ServerResponse doInBackground(ServerRequest... serverRequests) {
-            return kRemoteInterface_.createCustomUrlSync(serverRequests[0].getPost());
+            String urlExtend = "v1/url";
+            return branchRemoteInterface_.make_restful_post(serverRequests[0].getPost(), prefHelper_.getAPIBaseUrl() + urlExtend, Defines.RequestPath.GetURL.getPath(), prefHelper_.getBranchKey());
+//            return branchRemoteInterface_.createCustomUrlSync(serverRequests[0].getPost());
```
aaustin: remove this comment before release

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: Good point. Actually we  have  this defined in couple of places. We can define one in gradle and use the same anywhere.

sojanpr: done

Branch-SDK/src/io/branch/referral/network/BranchRemoteInterface.java
```diff
@@ -0,0 +1,276 @@
...
+ *
+ * @see {@link #doRestfulGet(String)}
+ * {@link #doRestfulPost(String, JSONObject)}
+ * {@link io.branch.referral.network.BranchRemoteInterface.BranchResponse}
+ * {@link BranchRemoteException}
+ * </p>
+ */
+public abstract class BranchRemoteInterface {
+    private static final String BRANCH_KEY = "branch_key";
+    public static final String RETRY_NUMBER = "retryNumber";
```
aaustin: any way we can move these constants out of this class as well?

sojanpr: This is intentionally put inside this class assuming the custom implementations also want o send the retry number. They can use the same key

aaustin: It looks like only retry number is used in our implementation. Can you add a little documentation on the requirements of when to add it in?

Branch-SDK/src/io/branch/referral/network/BranchRemoteInterface.java
```diff
@@ -0,0 +1,285 @@
...
+    /**
+     * Key for adding retry numbers for the request. This will help better network issue analysis and debugging.
+     * Add the retry number to the GET Request as a query param and as a JSon key value for post
+     */
+    public static final String RETRY_NUMBER = "retryNumber";
+
+    //----------- Abstract methods-----------------------//
+
+    /**
+     * <p/>
```
EvangelosG: </p> should be <p> -- will cause javadoc gen/jar build error

EvangelosG:
@sojanpr run makeJar with this PR and you will see the ~6 comment related issues in BranchRemoteInterface.java which breaks JAR generation.

Branch-SDK/src/io/branch/referral/network/BranchRemoteInterface.java
```diff
@@ -0,0 +1,285 @@
...
+     * @see {@link io.branch.referral.network.BranchRemoteInterface.BranchRemoteException}
+     * {@link io.branch.referral.network.BranchRemoteInterface.BranchResponse}
+     *
+     * NOTE: For better debugging purpose conside adding {@link #RETRY_NUMBER} as a query params if you implement multiple retries for your request
+     * </p>
+     */
+    public abstract BranchResponse doRestfulGet(String url) throws BranchRemoteException;
+
+    /**
+     * <p/>
```
EvangelosG: `</p>` should be `<p>` -- will cause javadoc gen/jar build error

PR #446, opened 2017-06-13
Network abstraction okie http

EvangelosG:
updated buildscript to support optional dependencies resulting in the following behavior:

if the app using Branch has okhttp dependency, our SDK will use the okhttp interface, otherwise it falls back to the httpurl interface.

@sojanpr


sojanpr:
Closing this for now since SDK provides option to extend  the network interface to any developer preferred model

PR #449, opened 2017-06-21
Fixing an 	instant app related issue

sojanpr:
Fixing an instant app redirect to full app issue

@aaustin @EvangelosG


aaustin:
👍

PR #452, opened 2017-07-03
Install Listner Fix

sojanpr:
There are below cases fixed 
1) Considering the case that url may contain query params with `=` and `&`
with it and may cause issue when parsing play store referrer.

On the receiving end the referrer is split based on ”&“ and each key
value  is detected by “=“ . The kay and value and again url decoded and
hence taking care of the double url encoding case

2)In case play store can deliver install listener faster or the install
listener get delayed, the current version misses the referrrer.  Added
changes to capture and save referrer and report it as soon as SDK ask
for it

@EvangelosG @aaustin @Sarkar

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -26,8 +26,10 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.io.UnsupportedEncodingException;
 import java.lang.ref.WeakReference;
 import java.net.HttpURLConnection;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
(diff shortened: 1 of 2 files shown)


sojanpr:
Test Steps
1)  Need  to verify the referrer getting  passed  using ADB install referrer broadcast
2) Need  to verity the referrer string with urls containing  multiple query params are handled (Eg Url encoded version of https://branchster.app.link/5dWl9ZfcaE?key1=value1&key2=value2" )
3) Need  to verify the broad cast occurring time
    - Broadcast the  ADB referrer broadcast and then wait for few seconds, then launch the app
    - Broadcast the  ADB referrer broadcast and immediately  launch the app

aaustin:
Looks good to me 👍 

cc @wlindemann FYI

EvangelosG:
**Need to verify the broad cast occurring time**
> Broadcast the ADB referrer broadcast and then wait for few seconds, then launch the app
> Broadcast the ADB referrer broadcast and immediately launch the app

Neither of these scenarios will deliver the referrer payload. App has to already be launched. Tested on a Pixel, which determines if the referrer was delivered almost instantaneously. When we were testing previously, payloads were taking a few additional seconds. Running more tests on other devices now. @sojanpr If you have a specific adb referrer payload to broadcast let me know.

PR #453, opened 2017-07-12
Prevent double reading of App Links and URI schemes

aaustin:
It's possible that we read and set an App Link URL in addition to a URI scheme with this current logic if readAndStripParams is called multiple times. @sojanpr


aaustin:
No longer relevant

PR #490, opened 2017-10-10
Add retry for IOException on doRestfulPost

moriadamri:
We having a lot of "Poor connectivity" errors from our users, we want to be able to retry the branch deep-link request to the server on those cases.


sojanpr:
@moriadamri thank you very much for the PR. 
Since the expected behavior on network retry is different for many of our partners we will not be able to merge this PR

We have already opened an interface for partners to build their on interface for Branch network requests. This way you can use any other  libs like  Volley, OKHTTP etc  with Branch. 
(in your case  you can extend `BranchRemoteInterfaceUrlConnection` class and override the method you wish.  Please see this API to see how you can set your custom network interface

https://github.com/BranchMetrics/android-branch-deep-linking/blob/master/Branch-SDK/src/io/branch/referral/Branch.java#L447

EvangelosG:
Here's an [unpublished] example of a custom BranchRemoteInterface using OKHttp: https://github.com/BranchMetrics/android-branch-deep-linking/pull/444/files#diff-0adeea19b472fd64b52322e059f24263

PR #497, opened 2017-10-31
Fixing check for Instant deep linking only while app is uninitialized

sojanpr:
Setting `instantDeepLinkPossible`  only while app is uninitialized

@aaustin @EvangelosG @ahmednawar


aaustin:
I didn't review the previous PR, so I don't have full context on this, but the code looks fine 👍 Seems like it won't affect any of the existing functionality.

PR #499, opened 2017-11-06
persisting and retrieving link data

sojanpr:
Fixing a crash caused by creating url from a persisted `ServerRequestCreateUrl` as described in the issue here

https://github.com/BranchMetrics/android-branch-deep-linking/issues/498

@EvangelosG @aaustin @mparul

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
+++ b/Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
@@ -96,6 +96,12 @@ class ServerRequestCreateUrl extends ServerRequest {
 
     public ServerRequestCreateUrl(String requestPath, JSONObject post, Context context) {
         super(requestPath, post, context);
+        // Retrieve persisted link data
+        linkPost_ = new BranchLinkData();
+        if (post.has(Defines.Jsonkey.CustomBranchLinkDataObj.getKey())) {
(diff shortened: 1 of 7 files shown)


Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
```diff
@@ -256,6 +262,20 @@ void setIsReqStartedFromBranchShareSheet(boolean startedByShareSheet) {
     boolean isReqStartedFromBranchShareSheet() {
         return isReqStartedFromBranchShareSheet_;
     }
+    
+    @Override
+    public JSONObject toJSON() {
+        JSONObject jsonObject = super.toJSON();
+        try {
+            jsonObject.putOpt(Defines.Jsonkey.CustomBranchLinkDataObj.getKey(), linkPost_.getLinkDataJsonObject());
```
aaustin: If the link data isn't getting persisted to preferences, shouldn't we just store the link data correctly rather than add a whole separate key to store the data in?

aaustin: Just read through the code to refresh myself. Seems like you need to store the other properties in here as well isReqStartedFromBranchShareSheet_, defaultToLongUrl_, isAsync_. Plus, there'll be no reference to the callback anymore after reloading which I imagine would be an issue, right?

sojanpr:
On further review it looks like there is no use case for persisting a create url request since there is no callback. So skipping persisting create url requests.

aaustin:
nice! 👍

PR #508, opened 2017-11-21
Ensuring backward compatibility in case using a deprecated BUO method for latest SDK

sojanpr:
`BUO# addContentMetadata()` is deprecated and new integration should use `BUO#setContentMetadata()`.
This fix ensure backward compatibility in case the app is still using the deprecated  methods

@aaustin @EvangelosG @derrickstaten


EvangelosG:
👍

PR #513, opened 2017-11-30
Fixing a possible for  thread sync issue

sojanpr:
Fix for  a related issue mentioned  here
https://github.com/BranchMetrics/android-branch-deep-linking/issues/512

Unable to reproduce this issue but it look like the Monitor Object was set to `queue` which is a variable inside the class. Changing the monitor object to the Container object (`ServerRequestQueue`) in order to make sure the synchronous black is not interleaved

@aaustin @EvangelosG


aaustin:
looks safe 👍

sojanpr:
Test steps.
1) Launch testbed app
2) After Branch init completed, turn off (internet connection)
3) Click `Track Custom Event` and `Track Custom Event` buttons
4) Turn on internet connection
5) Click `report view` button
6) Close and reopen app

Expected Result 
1) SDK should queue the request for tracking events  when there is no internet connection and replay it when there is internet. At step 5 the `registerView` event should be send to server after sending queued standard and custom events.
2) SDK should clear out the queue properly. At step 6 there should be only av1/open call

PR #522, opened 2017-12-21
fix synchronization in ServerRequestQueue

dmkessler:


sojanpr:
@dmkessler Thanks for the PR. We are reviewing it and will merge to the next release  after QA pass

dmkessler:
@sojanpr Thank you for taking a look at this.  Let me know if there's anything else I can do to help.

Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
```diff
@@ -106,7 +104,7 @@ public void run() {
         if (jsonStr != null) {
             try {
                 JSONArray jsonArr = new JSONArray(jsonStr);
-                for (int i = 0; i < Math.min(jsonArr.length(), MAX_ITEMS); i++) {
+                for (int i = 0, size = Math.min(jsonArr.length(), MAX_ITEMS); i < size; i++) {
```
sojanpr: @dmkessler While this is ideal for iteration in a multi threaded condition, Is there a real chance for this to happen. The JSON array is created by a string retrieved from shared preference. Shared preference is thread safe. Is there a chance in your application using multiple process like a remote service which can access the preference simultaneously. (just thinking of way to reproduce the issue)

sojanpr:
@dmkessler Thank you very much for this PR. Awesome !!  

Agree that all the queue operations should be  synchronized on a  single monitor object. We copied these changes to a new PR https://github.com/BranchMetrics/android-branch-deep-linking/pull/525 since our QA setup prefer to have PRs against staging.

Also, agree with you completely that many of the exceptions (`indexOutOfBounds`, `NoSuchElement`, `ConcurrentModification`) are not needed  since all the queue operations are  synchronized. They are kept as it is now for additional safety but will remove once we have a good amount testing on this.

Would really  appreciate if you can help with a review in the above PR

Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
dmkessler: Because jsonArr is initialized with jsonStr, and because java is pass by value rather than pass by reference, we can get the size of the array once when we start the loop rather than each iteration which it was doing previously.  It's just a trivial optimization.  If you are unsure about it this does not affect the fix in this PR and can be left out.

dmkessler:
@sojanpr Glad I could help out.  I left some feedback with the reasoning but that line change is not necessary for the actual fix and can be left out or moved into a new PR if you'd like.  Let me know if you need anything else.

sengoontoh:
Will someone please merge this?

sojanpr:
@sengoontoh This PR is in QA now. Will be released with v2.14.4 planned for early next week

sengoontoh:
Thanks for the update @sojanpr 
I'm actually using the cordova branch plugin: https://github.com/BranchMetrics/cordova-ionic-phonegap-branch-deep-linking

Does that plugin then reference this code? How do I make sure it picks up this fix?
Thanks!

sojanpr:
@sengoontoh We will be updating all our cross platform SDKs once we  make this release  on the native SDK. Will update you once it is released.

dmkessler:
@sojanpr i guess we can close this PR now that the fix was moved to the staging branch?  do you have a timeline on when your next release will be in maven?  we have a release coming up and i'd like to update our app back to the newest branch.io dependency if the fix is working

sojanpr:
@dmkessler Changes are  already released with v2.14.4. Thank for your help

dmkessler:
Your welcome!  Thank you for evaluating it.

PR #524, opened 2017-12-22
DeepLink validator

agrimn:
- Added a SDK integration validator - Validates user's SDK integration:
1. Package Name
2. URI scheme
3. Applink domain
4. -alternate Applink domain

- Added a Deeplink Routing validator - append validate=true to your branch link to test your Deeplink routing.

- Fixed Testbed -alternate configuration


Branch-SDK/src/io/branch/referral/ApkParser.java
```diff
@@ -47,7 +53,8 @@
      *
      * @return A {@link String} containing the result of the decompression action.
```
sojanpr: wrong return type

Branch-SDK/src/io/branch/referral/ApkParser.java
```diff
@@ -98,41 +105,56 @@ public String decompressXML(byte[] xml) {
...
+                        int attrNameSi = LEW(xml, off+1*4);  // AttrName String Index
+                        int attrValueSi = LEW(xml, off+2*4); // AttrValue Str Ind, or FFFFFFFF
+                        int attrResId = LEW(xml, off+4*4);  // AttrValue ResourceId or dup AttrValue StrInd
+                        off += 5*4;  // Skip over the 5 words of an attribute
+
+                        attrName = compXmlString(xml, sitOff, stOff, attrNameSi);
+                        if (attrName.equals("scheme")) {
+                            attrValue = attrValueSi != -1 ? compXmlString(xml, sitOff, stOff, attrValueSi) : "resourceID 0x" + Integer.toHexString(attrResId);
+                            if (validURI(attrValue)) {
+                                intentFilters.put("scheme", attrValue);
```
sojanpr: Are we reading only one scheme her? An intent filter can have multiple schemes.  
Also hosts are associated with a scheme. This code doesnot seem to make a relation between a scheme and hosts

Branch-SDK/src/io/branch/referral/ApkParser.java
```diff
@@ -98,41 +105,56 @@ public String decompressXML(byte[] xml) {
...
+                            }
+                        } else if (attrName.equals("host")) {
+                            attrValue = attrValueSi != -1 ? compXmlString(xml, sitOff, stOff, attrValueSi) : "resourceID 0x" + Integer.toHexString(attrResId);
+                            JSONArray domainList;
+                            if (intentFilters.has("hosts")) {
+                                domainList = intentFilters.getJSONArray("hosts");
+                            } else {
+                                domainList = new JSONArray();
+                            }
+                            domainList.put(attrValue);
```
sojanpr: How about the path prefixes and paths? Deep link can change based on these.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2027,8 +2029,23 @@ String generateShortLinkInternal(ServerRequestCreateUrl req) {
...
+
+    /**
+     * <p>Gets the Branch Internal App Configuration for the last given live key</p>
+     *
+     */
+
+    JSONObject generateAppConfigInternal() {
+        ServerResponse response = null;
+        try {
+            response = new getAppConfigTask().execute().get(PrefHelper.TIMEOUT, TimeUnit.MILLISECONDS);
```
sojanpr: This Involves a blocking call. `getAppConfigTask.excute.get()` Waits if necessary for at most the given time for the computation. PrefHelper.TIMEOUT is `5500` sec. This can be potential ANR candidate (Since ANR watch dog timer is 5 sec)

Branch-SDK/src/io/branch/referral/IntegrationValidator.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.ArrayList;
+import java.util.jar.JarFile;
+
+/**
+ * Created by sojanpr on 9/15/17.
+ */
+
+public class IntegrationValidator {
+
+    public void validateSDKIntegration (Context context){
```
sojanpr: Do you think it is good to do all this logic in background thread. Preferably on  a Async Task with some progress updates

Branch-SDK/src/io/branch/referral/SystemObserver.java
```diff
@@ -164,7 +166,10 @@ private String getURIScheme(String packageName) {
                     xml = new byte[is.available()];
                     //noinspection ResultOfMethodCallIgnored
                     is.read(xml);
-                    scheme = new ApkParser().decompressXML(xml);
+                    JSONObject obj = new ApkParser().decompressXML(xml);
```
sojanpr: SDK  call `getURIScheme()` method on install request. Adding the extended logic with XML parsing in this method does not look good. It will slow down the v1/install call unnecessarily. Can we move this APK parsing for intents into separate method and use it only for integration validation

Branch-SDK/src/io/branch/referral/IntegrationValidator.java
```diff
@@ -0,0 +1,286 @@
...
+
+    public void validateDeeplinkRouting(final JSONObject validate_json,final WeakReference<Activity> currentActivityReference_) {
+        Activity current_activity = currentActivityReference_.get();
+        AlertDialog.Builder builder;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            builder = new AlertDialog.Builder(current_activity, android.R.style.Theme_Material_Dialog_Alert);
+        } else {
+            builder = new AlertDialog.Builder(current_activity);
+        }
+        builder.setTitle("Branch Deeplinking Routing")
```
sojanpr: Possible window leak if the activity providing context  is moved.

Branch-SDK/src/io/branch/referral/Branch.java
agrimn: Synced offline. Since this call is not meant to be made on the production deployment we should be okay with this.

Branch-SDK/src/io/branch/referral/IntegrationValidator.java
agrimn: Synced offline. Since this call is not meant to be made on the production deployment we should be okay with this.

PR #525, opened 2017-12-26
Syncing all queue operation on one monitor object

sojanpr:
This changes are copied from awesome suggestions and fixes from  PR https://github.com/BranchMetrics/android-branch-deep-linking/pull/522 on few basic synchronization issues. 

This changes sync all queue operations on a single monitor object and prevents any parallel queue modifications

@aaustin

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
+++ b/Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
@@ -27,7 +27,9 @@ class ServerRequestQueue {
     private SharedPreferences sharedPref;
     private SharedPreferences.Editor editor;
     private final List<ServerRequest> queue;
-
+    //Object for synchronising operations on server request queue
+    private static final Object reqQueueLockObject = new Object();
+    
     /**
      * <p>Singleton method to return the pre-initialised, or newly initialise and return, a singleton
      * object of the type {@link ServerRequestQueue}.</p>
@@ -47,7 +49,7 @@ class ServerRequestQueue {
         }
         return SharedInstance;
     }
-
+    
     /**
      * <p>The main constructor of the ServerRequestQueue class is private because the class uses the
      * Singleton pattern.</p>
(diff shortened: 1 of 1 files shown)


aaustin:
The problem all along being that locking on a `final` object didn't give us the same threadsafety as `static final`? Eesh. 👍 Need lots of high volume load testing

Branch-SDK/src/io/branch/referral/ServerRequestQueue.java
```diff
@@ -154,14 +155,16 @@ void enqueue(ServerRequest request) {
      */
     ServerRequest dequeue() {
         ServerRequest req = null;
-        try {
-            req = queue.remove(0);
-            persist();
-        } catch (IndexOutOfBoundsException | NoSuchElementException ignored) {
+        synchronized (reqQueueLockObject) {
+            try {
+                req = queue.remove(0);
```
dmkessler: The extraneous try/catches are removable if you return from the public functions that already exist with the try catches.  In this case it would be removeAt(0);

That way you're not duplicating code/error handling.  Not a deal breaker for the PR though and could be moved to a new one considering it isn't related to the actual fix this is addressing

dmkessler:
@aaustin the problem wasn't related to the visibility/modifiers of the list it was because the logic allowed for the class member List object to be accessed from multiple threads.  In order to prevent concurrent modifications all access to the queue class member needed to be synchronized on the same object.  The visibility and static modifier were just obvious choices for it's use in this case.

sojanpr:
@dmkessler We are aware of extraneous try/catches  and want to keep it until we have a strong testing on it. Also `enqueue()  method removing the element at 1 is for a specific  reason with handling of session init request. Will revisit for a better approach later since we don't want to  complicate the concurrency fix.

sojanpr:
Test steps.

Launch testbed app
After Branch init completed, turn off (internet connection)
Click Track Custom Event and Track Custom Event buttons
Turn on internet connection
Click report view button
Close and reopen app
Expected Result

SDK should queue the request for tracking events when there is no internet connection and replay it when there is internet. At step 5 the registerView event should be send to server after sending queued standard and custom events.
SDK should clear out the queue properly. At step 6 there should be only av1/open call

PR #531, opened 2018-01-11
Support for link data or property update on post share sheet events

sojanpr:
Branch share sheet enhancement for controlling the link creation on Share sheet events. Allows apps to modify the link data or link properties after user has selected a channel for sharing the link with.

@aaustin @E-B-Smith


aaustin:
Looks good to me! :+1:

PR #533, opened 2018-01-16
Fixing issue :Re-open after offline install event fails

sojanpr:
SDK was failing to do init session when there was a connectivity issue
while opening the app for first time.

This was caused by improperly nullifying the install listener callback
which will be null when app is opened from background causing app to
wait for install referrer finish forever.

@Sarkar @aaustin


Sarkar:
code makes sense to me 👍

aaustin:
seems harmless 👍

PR #536, opened 2018-01-24
Install Update Timestamps

sojanpr:
Adding
1) original_install_time
2) first_install_time
3) last_update_time
4) previous_update_time

 The Original install time will have a the very first install time only if the app allows preference back up.
Apps that need to distinguish between a fresh install and re-install need to allow backing up of preferences.
Previous install time stamp always carry the last last known update time. the value will be zero for any fresh install and will be the last update time with successive opens.
 Previous install time will have a value less than last update time ever since the app is updated.

@E-B-Smith @Sarkar @aaustin


Branch-SDK/src/io/branch/referral/Defines.java
```diff
@@ -60,6 +60,10 @@
         Language("language"),
         IsReferrable("is_referrable"),
         Update("update"),
+        OriginalInstallTime("original_install_time"),
+        FirstInstallTime("first_install_time"),
```
aaustin: original vs first is pretty unclear. Perhaps the following for clarity:
`original_install_time` -> `first_install_time`
`first_install_time` -> `latest_install_time`

Branch-SDK/src/io/branch/referral/ServerRequestInitSession.java
```diff
@@ -200,6 +231,65 @@ public void onPreExecute() {
...
+        PackageInfo packageInfo = null;
+        try {
+            packageInfo = context_.getPackageManager().getPackageInfo(context_.getPackageName(), 0);
+        } catch (PackageManager.NameNotFoundException ignored) {
+        }
+        if (PrefHelper.NO_STRING_VALUE.equals(prefHelper_.getAppVersion())) {
+            // Default, just register an install
+            installOrUpdateState = STATE_FRESH_INSTALL;
+            // if no app version is in storage, this must be the first time Branch is here
+            if (packageInfo != null && packageInfo.lastUpdateTime != packageInfo.firstInstallTime) {
```
aaustin: Just to be safe for this time compare and the one below on line 288, can you add a 24 hour buffer for equality? i.e. `abs(packageInfo.lastUpdateTime - packageInfo.firstInstallTime) > 24` I worry about the same issue happening here that happened on iOS where the times vary by a minute or two because of an OS change.

Branch-SDK/src/io/branch/referral/Defines.java
sojanpr: unfortunately Android classes calla latest_install_time as `first_install_time`. We added `original_install_time` to avoid the confusion. Do you think that make sense

aaustin: I see what it's called in the class, but this is our own internal reference. That means the data team will have to understand what it means, not necessarily other Android people. Because `original` and `first` are basically the same thing, I think it'd be helpful to change the description to reflect the functionality.

sojanpr: Make sense, will change this

PR #539, opened 2018-01-26
Support Install Referrer lib

sojanpr:
Adding support for installreferrer library for getting  the referrer
info.
This is need `installreferrer` lib added to the application.
This will be working only with compatible version of Play store app. In
case of an error this fallback to the install-referrer broadcast.

Capture following  additional info on app install 
1) clicked_referrer_ts
2) install_begin_ts

@aaustin @Sarkar

### DIFF
--- a/Branch-SDK/build.gradle
+++ b/Branch-SDK/build.gradle
@@ -8,10 +8,13 @@ dependencies {
     compile 'com.android.support:support-annotations:22.2.0'
     compile 'com.crashlytics.sdk.android:answers-shim:0.0.6'
 
-    // This is an optional dependency
-    //Please note that the Branch SDK does not require Firebase to operate. This dependency is listed here so there will not be build errors,
+    // --- optional dependencies -----
+    //Please note that the Branch SDK does not require any of the below optional dependencies to operate. This dependency is listed here so there will not be build errors,
     // but the library is *not* added to your app unless you do so yourself. Please check the code in gradle-mvn-push script to see how this works
+
     compile 'com.google.firebase:firebase-appindexing:10.0.1'
+    compile 'com.android.installreferrer:installreferrer:1.0'
+
 }
 
 android {
--- a/Branch-SDK/src/io/branch/referral/InstallListener.java
+++ b/Branch-SDK/src/io/branch/referral/InstallListener.java
@@ -52,6 +52,13 @@ public class InstallListener extends BroadcastReceiver {
     @Override
     public void onReceive(Context context, Intent intent) {
         String rawReferrerString = intent.getStringExtra("referrer");
+        processRawReferrer(context, rawReferrerString);
+        if (isWaitingForReferrer) {
+            reportInstallReferrer();
+        }
+    }
+    
+    private void processRawReferrer(Context context, String rawReferrerString) {
         if (rawReferrerString != null) {
             try {
                 rawReferrerString = URLDecoder.decode(rawReferrerString, "UTF-8");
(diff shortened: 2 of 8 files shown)


Branch-SDK/src/io/branch/referral/InstallListener.java
```diff
@@ -4,23 +4,33 @@
 import android.content.Context;
 import android.content.Intent;
 import android.os.Handler;
+import android.os.RemoteException;
 import android.text.TextUtils;
 import android.util.Log;
 
+import com.android.installreferrer.api.InstallReferrerClient;
```
aaustin: won't this trigger a build error if the developer doesn't include the library?

Branch-SDK/build.gradle
```diff
@@ -8,10 +8,13 @@ dependencies {
...
     compile 'com.crashlytics.sdk.android:answers-shim:0.0.6'
 
-    // This is an optional dependency
-    //Please note that the Branch SDK does not require Firebase to operate. This dependency is listed here so there will not be build errors,
+    // --- optional dependencies -----
+    //Please note that the Branch SDK does not require any of the below optional dependencies to operate. This dependency is listed here so there will not be build errors,
     // but the library is *not* added to your app unless you do so yourself. Please check the code in gradle-mvn-push script to see how this works
+
     compile 'com.google.firebase:firebase-appindexing:10.0.1'
+    compile 'com.android.installreferrer:installreferrer:1.0'
```
aaustin: Please make sure Sahil and you can update the docs appropriately if we're recommending this going forward

Branch-SDK/src/io/branch/referral/InstallListener.java
aaustin: A lot of devs will have included this class already but not have added the new dependency

Branch-SDK/src/io/branch/referral/InstallListener.java
```diff
@@ -52,6 +66,97 @@ public void run() {
...
+    
+    private static void onReferrerClientFinished(Context context, String rawReferrerString, long clickTS, long InstallBeginTS) {
+        processReferrerInfo(context, rawReferrerString, clickTS, InstallBeginTS);
+        if (isWaitingForReferrer) {
+            reportInstallReferrer();
+        }
+    }
+    
+    private static void onReferrerClientError() {
+        isReferrerClientAvailable = false;
```
aaustin: don't we need to handle the callback here in the error case?

Branch-SDK/src/io/branch/referral/InstallListener.java
sojanpr: This dependency is optional and Install listener class is already built with SDK. The exception will throw at run time only when Java try to load the class, which is already handled in the blow lines

aaustin:
Do you have a way to test this in a few conditions like high latency, no connection when app opened, etc?

Branch-SDK/src/io/branch/referral/InstallListener.java
sojanpr: Referrr client is not support on many play store app versions yet. So dev need to have the  broadcast listener at least for some more time. In case of ReferrerClientError we still need to check if we can get the referrer from broadcast.
Callback happens when the  time out is triggered in the above scenario

aaustin: Understood!

Branch-SDK/src/io/branch/referral/InstallListener.java
aaustin: Got it. I'd just make sure this is a part of the test cases: Upgrading from an older version of SDK without the gradle inclusion

aaustin:
Looks good to me! 👍

Branch-SDK/src/io/branch/referral/InstallListener.java
sojanpr: Will ensure this test

PR #544, opened 2018-02-05
fix incorrect option highlighted in Share Dialog

ibogolyubskiy:
The problem is when using headers in a ListView they are also considered as items in the onItemClickListener, so you need to subtract the number of added headers from the current position to get REAL clicked position in ListView.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
+++ b/Branch-SDK/src/io/branch/referral/ShareLinkManager.java
@@ -226,7 +226,7 @@ class ShareLinkManager {
                         builder_.getShortLinkBuilder().setChannel(((ResolveInfo) view.getTag()).loadLabel(context_.getPackageManager()).toString());
                         callback_.onChannelSelected(selectedChannelName);
                     }
-                    adapter.selectedPos = pos;
+                    adapter.selectedPos = pos - shareOptionListView.getHeaderViewsCount();
                     adapter.notifyDataSetChanged();
                     invokeSharingClient((ResolveInfo) view.getTag());
                     if (shareDlg_ != null) {


sojanpr:
@ibogolyubskiy Thanks for the PR. Will review and add it with next release

Branch-SDK/src/io/branch/referral/ShareLinkManager.java
```diff
@@ -216,7 +218,7 @@ private void createShareDialog(List<SharingHelper.SHARE_WITH> preferredOptions)
             public void onItemClick(AdapterView<?> adapterView, View view, int pos, long l) {
                 if (view.getTag() instanceof MoreShareItem) {
                     appList_ = cleanedMatchingAppsFinal;
-                    adapter.notifyDataSetChanged();
+                    shareOptionListView.setAdapter(adapter = new ChooserArrayAdapter());
```
sojanpr: @ibogolyubskiy Do you think this is needed. Each time creating the adapter again is costly

sojanpr:
@ibogolyubskiy  Not able to reproduce this issue. SDK is adding header views in non selectible way.
 https://github.com/BranchMetrics/android-branch-deep-linking/blob/master/Branch-SDK/src/io/branch/referral/ShareLinkManager.java#L203
Is there any other  thing added to the code making this issue?

ibogolyubskiy:
Just tested with actual repository version.
Bug still can be reproduced in the Test-Bed application.
You can see incorrect behavior in attached video.
https://www.dropbox.com/s/f8i7b39s5aahfeu/with_header.mp4?dl=0
Also you can see incorrect selection on this screenshot
image

Branch-SDK/src/io/branch/referral/ShareLinkManager.java
ibogolyubskiy: I suggest to set new list adapter instead adapter.notifyDataSetChanged() to avoid residual animation after click on "Show more" item. 
https://www.dropbox.com/s/em2yc5mfhj031dd/residual_animation.mp4?dl=0
As you can see in video, after click on "Show more" dataset for adapter changed, but previous ripple animation for selection on "Show more" not completely finished and after changing dataset animation finishes already on "Android Beam" item.

ibogolyubskiy:
Not sure it is device dependent bug.
Nexus 5 with pure Android 6.0 (API 23) shows same result.

sojanpr:
@ibogolyubskiy Thank you very much for the fix and the details. This fix looks good to me  👍 

Since we  have to pass QA before merging to master , moving this PR to https://github.com/BranchMetrics/android-branch-deep-linking/pull/566
Really appreciate if you can reviews  it. Will be added to the next version of SDK.

PR #546, opened 2018-02-13
BlackList URI Scheme v2

sojanpr:
Branch collect the URLs in the incoming intent for better attribution.
Branch SDK extensively check for any sensitive data in the URL and skip
if exist.
This implementation provides dynamic update feature to the blacklist
with regular expression support blacklist and whitelist. Apps can
also update the list with their own custom patterns

 @E-B-Smith @Sarkar @Prasanna-sr @aaustin


E-B-Smith:
👍

PR #548, opened 2018-02-15
Fix: Install  events blocked on wait lock on retry

sojanpr:
Fixing a corner case of install request being stuck in request queue up on initial install call is failed.

@aaustin @Sarkar


aaustin:
Looks good to me 👍

PR #554, opened 2018-03-07
Fixing issue - enable/ disable logging has no control when test key is used

sojanpr:
Fixing issue - enable / disable logging has no control when test key is used

@aaustin


aaustin:
👍

PR #557, opened 2018-03-09
Tracking Control

sojanpr:
Adding SDK complience with  upcoming policy changes outlined by GDPR.

Adds a way to Enable / Disable Tracking user data or actions.
Key feature are
 - When tracking disabled
       - Branch SDK will not cache or save any user tracking da
       - Branch SDK will not send any network request
       - Branch SDK will clear all previously collected user data
       - URL creation apis would still work but send a long url in place of short url
  
Support hot app init with Branch as  soon as tracking enabled.

Few other  fixes and changes to support the new feature

@E-B-Smith @aaustin @Sarkar


E-B-Smith:
LGTM 👍

aaustin:
I think this PR needs extensive load testing to attempt to cause some concurrent exception modifications. Like, can you write a test case that inits / fires off 100 events in a for loop and triggers the disable flow halfway through? Similar for re-init. I'm mostly worried about the queue/network request manipulation while it's in use.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1292,9 +1311,17 @@ public void closeSession() {
...
         executeClose();
         sessionReferredLink_ = null;
+        trackingController.updateTrackingState(context_); // Update the tracking state for next cold start
+    }
+    
+    /**
+     * Clears all pending requests in the queue
+     */
+    void clearPendingRequests() {
+        requestQueue_.clear();
```
aaustin: If this is called while a request is in progress (say v1/open), it'll still return and set all of the preference helper values after you cleared them in TrackingController, right?

aaustin:
you're planning to do a separate PR to handle the case of direct deep linking?

Branch-SDK/src/io/branch/referral/Branch.java
sojanpr: In tracking disabled mode, we are not adding any tracking related params to preference helper  when request is responded 
https://github.com/BranchMetrics/android-branch-deep-linking/blob/Tracking-Control/Branch-SDK/src/io/branch/referral/Branch.java#L2877

aaustin: I saw that later. Thanks! Deleting this comment

aaustin:
This seems like it should work. I'd definitely want that thread testing per my comment above on both iOS / Android.

sojanpr:
Added an automated test to exercise tracking mode switching and test link creation on both  tracking enabled  and disabled cases. Passes all the test
@aaustin

aaustin:
Nice 👍

E-B-Smith:
LGTM 👍2

PR #568, opened 2018-03-29
Update the dynamic device info params before sending  requests

sojanpr:
Update the dynamic device info params before sending  requests.

@aaaronlopez @aaustin


aaaronlopez:
LGTM 👍`updateDeviceInfo()` does the job.

PR #571, opened 2018-04-20
Fixing issue : Local reference get cleared when specifying Branch key through code

sojanpr:
Fixing issue : Local reference get cleared when specifying Branch key through code.

This  issue was causing v1/install to be called each time app killed and restarted if the Branch  key is  not specified in the manifest

@aaustin @Sarkar


Sarkar:
oh man. good find. 👍

aaustin:
👍

PR #573, opened 2018-05-04
Find Issue: Link data is unnecessarily escaped

sojanpr:
Branch v1/url request accept the `data` as JSONObj. Unfortunately the
data object is stringified and send to the  server. This causes
multiple escape for string format chars

@aaustin @Sarkar


aaustin:
Looking through blame, it does seem unnecessary. Looks good 👍

PR #579, opened 2018-05-17
Fix issue facebook_app_link_checked is always `false`

sojanpr:
Fix issue facebook_app_link_checked is always `false`

@aaustin


aaustin:
👍

PR #581, opened 2018-06-01
Remove request from retry queue in case of a bad request

sojanpr:
Remove request from retry queue in case of a bad request.

@Sarkar @aaustin


aaustin:
👍

PR #587, opened 2018-06-26
Adding support for Install metadata

sojanpr:
Adding API that  allows to tag the install with custom attribute. Add
any key-values that qualify or distinguish install .
Please make sure this method is called before the Branch init, which
is on the onStartMethod of first activity.
A better place to call this  method is right after
Branch#getAutoInstance()

@aaustin @Sarkar

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -880,8 +880,9 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
      * A better place to call this  method is right after Branch#getAutoInstance()
      * </p>
      */
-    public void addInstallMetadata(@NonNull String key, @NonNull String value) {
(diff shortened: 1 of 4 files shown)


aaustin:
👍

Sarkar:
quick question. if someone wants to specify this: 

{ 
~campaign: "some_campaign",
~channel: "some_channel",
K1: "v1",
K2: "v2"
}

it'd basically need to be called four times, right?

sojanpr:
Yes. They have to. Changed `addInstallMetadata()` method to return Branch instance to support chaining. So dev can call 
```Branch.getInstance().addInstallMetadata("~campaign", "some_campaign")
                                    .addInstallMetadata("~channel", "some_channel")```
[ Expectation is that partners are putting only minimal info here. It won't be a good idea for partners to put complex data here]

Sarkar:
Makes sense. Thanks 👍

PR #607, opened 2018-08-24
DEVEX-751: Always send standard events as standard even without object

aaaronlopez:
@E-B-Smith Android actually wasn't in parity with iOS here. With this change,

BranchEvent("PURCHASE")

will be treated the same as 

BranchEvent(BRANCH_STANDARD_EVENT.PURCHASE)


E-B-Smith:
Good catch.  Thanks! 👍

PR #608, opened 2018-08-28
DEVEX-761 fix custom event buo

aaaronlopez:


aaaronlopez:
@Sarkar another small one if you get the chance. We should be sending the BUO as content_items even if the event is not a "standard" one. iOS already does this.

Sarkar:
hm. thanks for answering that iOS does this. wonder why this was there in the first place? 👍

PR #610, opened 2018-08-31
DEVEX-766: Send GAID with v2/event

aaaronlopez:
@Sarkar @E-B-Smith @code-crusher


Sarkar:
wait, we were only getting AAID with v2/events? Yikes. Ok, this is good. 👍 

Let's make sure this goes through a round of QA review, and some level of automation is added such that we always confirm v2/events (and regular events) are passing GAID.  cc @GeneShay

aaaronlopez:
we were sending the GAID value in the key `user_data.aaid`, not the actual AAID, instead of `user_data.google_advertising_id` for all V2/event API calls

PR #613, opened 2018-09-10
DEVEX-776 instant app deep linking

aaaronlopez:
@E-B-Smith @code-crusher 

If we see the `instant=true` as a query param, we should do what we do for Chrome Intent Instant Deep Linking. This is only for the online case.

cc @GeneShay

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -1414,7 +1414,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
                             }
                         }
                     }
-                    // if not check the intent data to see if there is deep link params
+                    // If not check the intent data to see if there is deep link params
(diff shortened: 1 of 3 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -1394,41 +1394,53 @@ private boolean readAndStripParam(Uri data, Activity activity) {
         // If activity is created and launched then the intent can be readily consumed.
         // NOTE : IDL will not be working if the activity is launched from stack if `initSession` is called from `onStart()`. TODO Need to check for IDL possibility from any #ServerRequestInitSession
         if (!disableInstantDeepLinking) {
-            if (intentState_ == INTENT_STATE.READY || isActivityCreatedAndLaunched) {
```
aaaronlopez: This is the line I want to get rid of which will speed things up.

PR #642, opened 2018-11-02
feat: support for StrictMode#detectUntaggedSockets

code-crusher:
The TrafficStats.setThreadStatsTag() should be added just before using the OutputStreamWriter to abide the thread policy for Android O or above.

We don't need to use tagSocket() as we are not using any custom Socket, it's system default only.

Docs for ref: https://developer.android.com/reference/android/os/StrictMode.VmPolicy.Builder#detectUntaggedSockets()

Reviewers:
@Sarkar @aaaronlopez @apeterson-branch


Sarkar:
So nice. 👍

aaaronlopez:
👍 staging please my friend!

PR #643, opened 2018-11-02
[DEVEX-835] fix: unstable itemViewHeight

code-crusher:
Changes:
Replaced the getIntrinsicHeight() with getCurrent().getBounds().centerY()*ICON_SIZER

This way there is a uniformity among all the icons. The default ICON_SIZER is 3 to compensate with the near-ish safe distance followed by Branch SDK as of now.

Yes, this works for non-Samsung devices also. 

Reviewers:
@Sarkar @aaaronlopez @apeterson-branch


aaaronlopez:
lgtm, will test on staging 👍

PR #653, opened 2018-11-29
INTENG-4460 -- Protection around getSystemService

apeterson-branch:
## Reference
INTENG-4460 -- Exception in SystemObserver.getUIMode

## Description
Customer reports that an exception can be thrown when calling `getSystemService("uimode")`, if that returns null.

Note that I was not able to reproduce this issue in-house, however `lint` flagged that (and several other lines that were using `getSystemService()`) as potentially returning null.  I cleaned up the other calls in this class as well.  I also converted the Context to a WeakReference as well.

## Testing Instructions

## Risk Assessment [`MEDIUM`]
This is pretty core functionality.  It should be reviewed carefully.

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/io/branch/referral/SystemObserver.java
+++ b/Branch-SDK/src/io/branch/referral/SystemObserver.java
@@ -129,8 +129,9 @@ class SystemObserver {
         try {
             if (context_ != null) {
                 PackageInfo packageInfo = context_.getPackageManager().getPackageInfo(context_.getPackageName(), 0);
-                if (packageInfo.versionName != null)
+                if (packageInfo.versionName != null) {
                     return packageInfo.versionName;
+                }
             }
         } catch (NameNotFoundException ignored) {
         }
--- a/Branch-SDK/src/io/branch/referral/SystemObserver.java
+++ b/Branch-SDK/src/io/branch/referral/SystemObserver.java
@@ -3,7 +3,6 @@ package io.branch.referral;
 import android.Manifest;
 import android.app.UiModeManager;
 import android.content.Context;
-import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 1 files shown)


Branch-SDK/src/io/branch/referral/SystemObserver.java
```diff
@@ -429,32 +445,39 @@ static String getLocalIPAddress() {
...
-                break;
-            default:
-                mode = "UI_MODE_TYPE_UNDEFINED";
-                break;
+        String mode = "UI_MODE_TYPE_UNDEFINED";
+        Context context = contextReference_.get();
+        UiModeManager modeManager = null;
+
+        if (context != null) {
+            modeManager = (UiModeManager)context.getSystemService(UI_MODE_SERVICE);
```
aaustin: Feel like we should throw a try catch around this statement

aaustin:
So my general concern is regarding which context we're using here. Our guidance on the initSession call is to use the Activity context, but our `getAutoInstance` grabs the App context. If we just have a weak ref to the Activity, there's a risk that it could get GCed before the call is made to retrieve these values. Can you confirm that the init path is passing the App context?

Branch-SDK/src/io/branch/referral/SystemObserver.java
apeterson-branch: I can do this, but the change here was to separate out the Context (which seemed in the stack trace to be "dead") and the getting of the system service (which can return null) and finally the calling of that method.

To be really safe I can wrap the entire method, but we shouldn't be guessing here on what the original problem was that caused the exception -- that might just push the problem further down the path.

Let me know if you want me to wrap just the code indicated, or the whole method.

apeterson-branch: Reverted Weak Reference and wrapped this method with a try/catch

aaustin:
👍

PR #658, opened 2018-12-05
SDK-87 -- Android InstallListener exception when not on UI thread

apeterson-branch:
## Reference
SDK-87 -- Android InstallListener exception when not on UI thread

## Description
If an event is logged and the thread logging it is not the UI thread, there was an exception where the code is creating a new Handler() just to wait for 1500 milliseconds before reporting the install referrer.

This change uses a Timer instead.  Normally Timers should be avoided, but for this short of a duration it should not cause any issues.

### Note
Note that there were lint warnings as well, so I cleaned those up while in here.

## Testing Instructions
* Unit tests should all pass

I wrote a unit test (on a different branch) that proved the exception occurred before and is fixed afterwards.  Unfortunately that branch is now behind this one so unit tests were not checked in here.

## Risk Assessment [`LOW`]
Most of the changes here were around getting the unit tests in the other branch to prove the issue exists, and prove the issue was resolved.

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/io/branch/referral/InstallListener.java
+++ b/Branch-SDK/src/io/branch/referral/InstallListener.java
@@ -3,7 +3,6 @@ package io.branch.referral;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
-import android.os.Handler;
 import android.os.RemoteException;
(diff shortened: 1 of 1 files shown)


aaaronlopez:
👍 makes sense to me

PR #678, opened 2019-03-02
INTENG-4676 -- No deep link registered when push notification received while app in foreground

apeterson-branch:
## Reference
INTENG-4676 -- No deep link registered when push notification received while app in foreground

## Description
When an app is in the foreground, and receives a new push notification, the new notification is not handled by Branch correctly.  

This change requires the customer to make a new call into the SDK to reinitialize the session when its activity receives an `onNewIntent()` message.

    @Override
    protected void onNewIntent(Intent intent) {
        super.onNewIntent(intent);

        setIntent(intent);
        Branch.getInstance().reInitSession(this, myReferralInitListener);
    }

The lifecycle is such that onNewIntent() is called by the OS before onResume -- which will then pick up the new intent correctly.

## Testing Instructions
* Unit tests should all pass

## Notes
* If this PR is accepted, this will require documentation changes to follow.
* Note that the activity should still do the initial initialization of the session in `onStart()` per documentation.

## Risk Assessment [`HIGH`]
The code changes around session management should be reviewed with care.

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -1308,12 +1308,12 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
     }
 
     public boolean reInitSession(Activity activity, BranchUniversalReferralInitListener callback) {
-        getFinalIntent(activity, true);
+        handleActivityIntent(activity, true);
         return  initSession(callback);
     }
 
     public boolean reInitSession(Activity activity, BranchReferralInitListener callback) {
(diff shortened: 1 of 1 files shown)


code-crusher:
Hey @apeterson-branch I tested this and looks good. 

Minor, but I think we should change the function to `reInitSession(BranchUniversalReferralInitListener callback, Activity activity)` to keep it aligned with the initSession format of taking callback first and activity second.

aaustin:
Did you see my comment on the Jira? I don't think we should go this route?

apeterson-branch:
Closing for alternate solution

PR #681, opened 2019-03-07
[INTENG-4676][INTENG-4809]: Add support for push edge case

aaustin:
@code-crusher FYI


code-crusher:
Nice!!! Though I'm also seeing an edge case with this because now the callback gets triggered on every resume without a Branch link, such as opening a notification but without a Branch link or switching from a different activity with no intent data. Then the` branch.getLatestReferringParams()` have `{}` which gets back in the `deferredInitListener_`. We should add a check for empty JSON here https://github.com/BranchMetrics/android-branch-deep-linking/blob/36c5adc61c100433c4399611ab2ce4a3b0b923ee/Branch-SDK/src/io/branch/referral/ServerRequestRegisterOpen.java#L93 to ensure we do not fire unnecessary callbacks. Do you have any other thoughts to prevent this?

code-crusher:
Hey, @apeterson-branch I have cleaned up the Branch for you to merge and deploy the SDK.

code-crusher:
👍

apeterson-branch:
This solution didn't work in my sample app.

csalmi-branch:
Due to some of the changes in this PR, code compilation now fails. I'm reverting this change for further investigation.

PR #683, opened 2019-03-13
SDK-157 -- setDebug issues

apeterson-branch:
## Reference
SDK-157 -- setDebug() no longer works as expected.

## Description
Internal testing found that calling setDebug() before initSession() can be overridden by the setting of test mode in the manifest.

As part of this ticket, I unwound how the interactions between debug mode and test mode, and refactored quite a lot of the different ways that debugging works within the SDK.

## Testing Instructions
* Unit tests pass

### Detailed Testing Instructions
It turns out that the original issue was around the distinction between `setDebug()` and `enableLogging()`, and the interactions that happen between them.

**To demonstrate the problem**
1. add `setDebug()` in the TestBed app in `onStart()`.
2. change the manifest test flag to "false"

* OLD behavior:  You will not see logs until after you rotate the screen or background/foreground the app.
* NEW behavior:  You will see logs appropriately.

## Risk Assessment [`MEDIUM`]
I touched rather more code that I expected to.  There is not only a number of different ways to enable debugging and test mode, the keys are tied to how debugging is enabled as well.

- [x] JIRA Ticket is referenced in PR title
- Correctness & Style
    - [x] Conforms to [Style Guides](https://google.github.io/styleguide/javaguide.html)
    - [x] Mission critical pieces are documented in code and out of code as needed

### DIFF
--- a/Branch-SDK/androidTest/io/branch/referral/DebugTest.java
+++ b/Branch-SDK/androidTest/io/branch/referral/DebugTest.java
@@ -1,0 +1,67 @@
+package io.branch.referral;
+
+import android.support.test.runner.AndroidJUnit4;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(AndroidJUnit4.class)
+public class DebugTest extends BranchTest {
+
+    @Test
+    public void testDebugState() {
+        /*
+        DebugMode         TestMode          isTestModeEnabled     isDebugModeEnabled
+           0                  0                false                 false
+           0                  1                true                  true
+           1                  0                false                 true
+           1                  1                true                  true
+        */
+
+        Assert.assertFalse(BranchUtil.isDebugEnabled());
+        Assert.assertFalse(BranchUtil.isTestModeEnabled());
+
+        // Debug Mode
+        Branch.enableDebugMode();
+        Assert.assertTrue(BranchUtil.isDebugEnabled());
+        Assert.assertFalse(BranchUtil.isTestModeEnabled());
+
+        Branch.disableDebugMode();
+        Assert.assertFalse(BranchUtil.isDebugEnabled());
+        Assert.assertFalse(BranchUtil.isTestModeEnabled());
+
+        // Test Mode
+        Branch.enableTestMode();
+        Assert.assertTrue(BranchUtil.isDebugEnabled());
+        Assert.assertTrue(BranchUtil.isTestModeEnabled());
+
+        Branch.disableTestMode();
+        Assert.assertFalse(BranchUtil.isDebugEnabled());
+        Assert.assertFalse(BranchUtil.isTestModeEnabled());
+
+        // Both Debug and Test Mode (variation A, debug before test)
+        Branch.enableDebugMode();
+        Branch.enableTestMode();
+        Assert.assertTrue(BranchUtil.isDebugEnabled());
+        Assert.assertTrue(BranchUtil.isTestModeEnabled());
+
+        Branch.disableDebugMode();
+        Assert.assertTrue(BranchUtil.isDebugEnabled());     // Per Javadoc, debug is enabled if either debug or test is enabled.
+        Assert.assertTrue(BranchUtil.isTestModeEnabled());
+
+        Branch.disableTestMode();
+        Assert.assertFalse(BranchUtil.isDebugEnabled());
+        Assert.assertFalse(BranchUtil.isTestModeEnabled());
+
+        // Both Debug and Test Mode (variation B, test before debug)
+        Branch.enableTestMode();
+        Branch.enableDebugMode();
+        Assert.assertTrue(BranchUtil.isDebugEnabled());
+        Assert.assertTrue(BranchUtil.isTestModeEnabled());
+
+        Branch.disableTestMode();
+        Assert.assertFalse(BranchUtil.isTestModeEnabled());
+        Assert.assertFalse(BranchUtil.isDebugEnabled());    // Per Javadoc, turning off test mode also turns off debug mode.
+    }
+}
--- a/Branch-SDK/androidTest/io/branch/referral/DebugTest.java
+++ b/Branch-SDK/androidTest/io/branch/referral/DebugTest.java
@@ -8,17 +8,16 @@ import org.junit.runner.RunWith;
 
 @RunWith(AndroidJUnit4.class)
 public class DebugTest extends BranchTest {
+    /*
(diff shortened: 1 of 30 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -692,7 +715,7 @@ public static Branch getAutoInstance(@NonNull Context context) {
     public static Branch getAutoInstance(@NonNull Context context, boolean isReferrable) {
         isAutoSessionMode_ = true;
         customReferrableSettings_ = isReferrable ? CUSTOM_REFERRABLE_SETTINGS.REFERRABLE : CUSTOM_REFERRABLE_SETTINGS.NON_REFERRABLE;
-        boolean isDebug = BranchUtil.isTestModeEnabled(context);
+        boolean isDebug = BranchUtil.checkTestMode(context);
```
aaaronlopez: does this `isDebug` flag imply the test is being used? If so, can you rename it to `isTestMode`? just to distinguish debug and test

aaaronlopez: or should it be equal to `isDebugEnabled`?

Branch-SDK/src/io/branch/referral/InstantAppUtil.java
```diff
@@ -68,14 +68,10 @@ private static boolean isPreReleaseOBuild() {
     @SuppressWarnings("ConstantConditions")
     static boolean doShowInstallPrompt(@NonNull Activity activity, int requestCode, @Nullable String referrer) {
         if (activity == null) {
-            if (Branch.isLogging_ != null && Branch.isLogging_) {
-                Log.e("BranchSDK", "Unable to show install prompt. Activity is null");
-            }
+            Log.e("BranchSDK", "Unable to show install prompt. Activity is null");
```
aaaronlopez: Why are we logging without checking that debug mode is enabled?

Branch-SDK/src/io/branch/referral/InstantAppUtil.java
```diff
@@ -68,14 +68,10 @@ private static boolean isPreReleaseOBuild() {
...
             return false;
         } else if (!isInstantApp(activity)) {
-            if (Branch.isLogging_ != null && Branch.isLogging_) {
```
aaaronlopez: same here

Branch-SDK/src/io/branch/referral/validators/IntegrationValidator.java
```diff
@@ -195,11 +194,11 @@ private boolean checkIfIntentAddedForURIScheme(String uriScheme) {
         if (integrationModel.deeplinkUriScheme != null) {
             for (Iterator<String> it = integrationModel.deeplinkUriScheme.keys(); it.hasNext(); ) {
                 String key = it.next();
-                if (uriHost.equals(key)) {
+                if (uriHost != null && uriHost.equals(key)) {
```
aaaronlopez: nice null pointer check

Branch-SDK/androidTest/io/branch/referral/DebugTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    public void testDebugState() {
+        /*
+        DebugMode         TestMode          isTestModeEnabled     isDebugModeEnabled
+           0                  0                false                 false
+           0                  1                true                  true
+           1                  0                false                 true
+           1                  1                true                  true
+        */
```
aaaronlopez: this matrix looks in line with the other SDKs. thank you

Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
```diff
@@ -164,7 +164,7 @@ private Uri buildStrongMatchUrl(String matchDomain, DeviceInfo deviceInfo, PrefH
             String hardwareIDTypeVal = deviceInfo.isHardwareIDReal() ? Defines.Jsonkey.HardwareIDTypeVendor.getKey() : Defines.Jsonkey.HardwareIDTypeRandom.getKey();
             uriString += "&" + Defines.Jsonkey.HardwareIDType.getKey() + "=" + hardwareIDTypeVal;
             // Add GAID if available
-            if (systemObserver.GAIDString_ != null && !BranchUtil.isTestModeEnabled(context)) {
+            if (systemObserver.GAIDString_ != null && !BranchUtil.isTestModeEnabled()) {
```
aaaronlopez: Does this mean we send the GAID when `isDebugModeEnabled`? If so, I think this is a deviation

Branch-SDK/src/io/branch/referral/Branch.java
apeterson-branch: I named it `checkTestMode` because there is a side effect of saving a state variable.  `is` and `get` methods should be constant and have no side effects.

I couldn't think of any other descriptive prefaces though.  Semantically, this method's intention is to "check to see if test mode is enabled".   In this case it sounds better to just call it `isTestModeEnabled()`.

apeterson-branch: Note that there is also a weird distinction between `isTestModeEnabled(context)` and `isTestModeEnabled()`.    These have different behaviors depending on the order called.

Branch-SDK/src/io/branch/referral/InstantAppUtil.java
apeterson-branch: If it is truly an error, it should always be logged.   These two instances of logging were the only ones in the entire code base that went directly to `Log` instead of going through the Branch Log method.

Branch-SDK/src/io/branch/referral/Branch.java
apeterson-branch: Outdated

Branch-SDK/src/io/branch/referral/InstantAppUtil.java
apeterson-branch: Outdated.

Branch-SDK/src/io/branch/referral/InstantAppUtil.java
apeterson-branch: Outdated.  Using `PrefHelper.Debug()` now.

Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
apeterson-branch: Outdated.  Now using `BranchUtil.checkTestMode()` to determine if the SDK is in test mode or not.

aaaronlopez:
👍

PR #687, opened 2019-03-21
SDK-223 -- Release v3.1.0

apeterson-branch:
Release Android SDK v3.1.0


GeneShay:
Just ran it through my automation - everything passed with no issues! 👍 
Also looked through the JIRAs that you have in Confluence. Doesn't seem like anything that needs to be tested separately as long as you verified the functionality, but tested the .setDebug()/enableDebugMode() after install - it worked as expected now! Great work!

PR #696, opened 2019-03-28
SDK-230 -- Unwind SystemObserver

apeterson-branch:
## Reference
SDK-230 -- Unwind SystemObserver

## Description
SystemObserver.java is used to collect things like advertising IDs, Hardware IDs, etc. The usage of this class is a bit convoluted, and because of statics and copies of these classes it isn't always working as expected.

For example, calls through DeviceInfo.java will only use the very first SystemObserver (because of a static), where other calls to a SystemObserver can be made returning completely different results. "disableDeviceIDFetch_" is one such variable, that depending on which direction a call is made from, will set different values in a payload.

## Testing Instructions
* Unit tests should all pass

## Notes
Unwinding this was a little more than I had hoped when first undertaking this task.  I essentially removed the ability to construct a SystemObserver outside of DeviceInfo.

## Risk Assessment [`MEDIUM`]
The code changes affected a lot of areas, and a lot of internal method signatures.  While the behavior might be slightly different when moving in and out of debug, test, or simulation mode -- it should now be internally consistent regardless of where the call is coming from.

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
+++ b/Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
@@ -25,26 +25,26 @@ import java.util.List;
 class BranchStrongMatchHelper {
 
     private static BranchStrongMatchHelper branchStrongMatchHelper_;
-    Object mClient_ = null;
+    private Object mClient_ = null;
     private static final int STRONG_MATCH_CHECK_TIME_OUT = 500; // Time to wait for strong match check
-    public static final int STRONG_MATCH_URL_HIT_DELAY = 750; // Time default delay time in ms between a Strong match uri call and v1/install
+    private static final int STRONG_MATCH_URL_HIT_DELAY = 750; // Time default delay time in ms between a Strong match uri call and v1/install
     private static final long THIRTY_DAYS_EPOCH_MILLI_SEC = 30 * 24 * 60 * 60 * 1000L;
     private final Handler timeOutHandler_;
     private static int StrongMatchUrlHitDelay = BranchStrongMatchHelper.STRONG_MATCH_URL_HIT_DELAY;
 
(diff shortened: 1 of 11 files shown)


Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
```diff
@@ -25,26 +25,26 @@
...
-    boolean isStrongMatchUrlLaunched = false;
+    private boolean isStrongMatchUrlLaunched = false;
 
-    Class<?> CustomTabsClientClass;
-    Class<?> CustomServiceTabConnectionClass;
-    Class<?> CustomTabsCallbackClass;
-    Class<?> CustomTabsSessionClass;
-    Class<?> ICustomTabsServiceClass;
+    private Class<?> CustomTabsClientClass;
+    // private Class<?> CustomServiceTabConnectionClass;
```
aaaronlopez: remove

Branch-SDK/src/io/branch/referral/DeviceInfo.java
```diff
@@ -234,19 +194,28 @@ public String getPackageName() {
      * @return {@link String} with app version value
      */
     public String getAppVersion() {
-        return appVersion_;
+        return systemObserver_.getAppVersion(context_);
     }
 
-    public boolean isHardwareIDReal() {
```
aaaronlopez: When we remove this, let's ensure the key/value is still sent on outgoing events in request payload

Branch-SDK/src/io/branch/referral/BranchStrongMatchHelper.java
apeterson-branch: Removed.

aaaronlopez:
let's do it! 👍

PR #700, opened 2019-04-08
SDK-271 -- Customer wants to be able to create short links from the SDK even when they disable tracking

apeterson-branch:
## Reference
SDK-271 -- Customer wants to be able to create short links from the SDK even when they disable tracking

## Description
The SDK disables creating short links from the SDK when tracking is disabled. We want to change this so that you can still create short links.

## Testing Instructions
* Unit tests should all pass

The way that I tested this was to use the Branch-SDK-TestBed app and disable tracking, then refresh the URL.   I also verified that the URL can be updated with tracking enabled.

## Risk Assessment [`MEDIUM`]
There is a piece of logic that was a complicated piece of negative logic.   Please carefully review Branch.java around the new method `needsSession()`

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/io/branch/referral/Branch.java
@@ -2319,7 +2319,7 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
                             handleFailure(requestQueue_.getSize() - 1, BranchError.ERR_NO_SESSION);
                         }
                         //All request except open and install need a session to execute
-                        else if (!(req instanceof ServerRequestInitSession) && (!hasSession() || !hasDeviceFingerPrint())) {
+                        else if (needsSession(req)) {
                             networkCount_ = 0;
                             handleFailure(requestQueue_.getSize() - 1, BranchError.ERR_NO_SESSION);
                         } else {
@@ -2338,6 +2338,18 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
         } catch (Exception e) {
             e.printStackTrace();
         }
+    }
+
+    private boolean needsSession(ServerRequest request) {
+        // All request except open and install need a session to execute (TODO -- Incorrect comment)
+        if (request instanceof ServerRequestInitSession) {
+            return false;
+        } else if (request instanceof ServerRequestCreateUrl) {
+            return false;
+        }
+
+        boolean sessionAvailable = (hasSession() || hasDeviceFingerPrint());
+        return !sessionAvailable;
     }
     
     private void handleFailure(int index, int statusCode) {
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2339,6 +2339,17 @@ else if (!(req instanceof ServerRequestInitSession) && (!hasSession() || !hasDev
             e.printStackTrace();
         }
     }
+
+    private boolean needsSession(ServerRequest request) {
```
csalmi-branch: Can we walk through this tomorrow? I don't think this new function ends up with quite the same results as part of the if statement.

Branch-SDK/src/io/branch/referral/ServerRequestCreateUrl.java
```diff
@@ -273,4 +273,11 @@ private void updateShareEventToFabric(String url) {
     boolean isPersistable() {
         return false; // No need to retrieve create url request from previous session
     }
+
+    @Override
+    protected boolean prepareExecuteWithoutTracking() {
```
csalmi-branch: Will this ever be false?

apeterson-branch: This is an overriden method from the base class ServerRequest.  ServerRequest default behavior is in fact false, but for a CreateUrl request -- it is always true per the ticket.

Branch-SDK/src/io/branch/referral/Branch.java
apeterson-branch: Updated Logic

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -2318,8 +2318,8 @@ private void processNextQueueItem() {
                             networkCount_ = 0;
                             handleFailure(requestQueue_.getSize() - 1, BranchError.ERR_NO_SESSION);
                         }
-                        //All request except open and install need a session to execute
-                        else if (!(req instanceof ServerRequestInitSession) && (!hasSession() || !hasDeviceFingerPrint())) {
+                        // Determine if a session is needed to execute (SDK-271)
+                        else if (requestNeedsSession(req) && !isSessionAvailableForRequest()) {
```
Fabi755: This looks good. But is this reachable? In my case `if`-condition with check on `instance of ServerRequestRegisterInstall ` in [line 2316](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/700/files#diff-11fc8f6026de99b60216e9aca37d6d26R2316) catch the request and fail with not initialized session.

apeterson-branch: Are you dropping into the block from the check on line 2316 (that was unchanged in this PR) or are you dropping into the block from check on line 2322?

Fabi755: I'm dropping on unchanged lines on 2316. 
This line checks if request from type `ServerRequestRegisterInstall`. On link generation this is not and without tracking I haven't a user. Or did I miss something?

PR #721, opened 2019-07-16
[CORE-139][CORE-119] add wrapper + keys for preinstall analytics

code-crusher:
JIRA: https://branch.atlassian.net/browse/CORE-119

**Description:**
To pass extra keys for the pre-install analytics in the install request for the server to read the attribute the install to those campaign and ad partner added. 

**Impact:**
Updating the v1/install request


GeneShay:
:+1

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,105 @@
+package io.branch.referral;
+
+import android.content.Context;
+import io.branch.referral.Defines.PreinstallKey;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.HashMap;
```
apeterson-branch: Compiler Warnings.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Created by --vbajpai on --24/07/2019 at --13:44 for --android-branch-deep-linking-attribution
+ */
+class BranchPreinstall {
+
+  private String SYSTEM_PROPERTIES_CLASS_KEY = "android.os.SystemProperties";
```
apeterson-branch: Should these be `private static final`

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,105 @@
+package io.branch.referral;
```
apeterson-branch: Branch classes are formatted with 4 spaces.   Not a huge issue, but this would be an outlier.

Branch-SDK/src/io/branch/referral/Branch.java
```diff
@@ -696,6 +697,7 @@ public static Branch getAutoInstance(@NonNull Context context) {
         customReferrableSettings_ = CUSTOM_REFERRABLE_SETTINGS.USE_DEFAULT;
         boolean isTest = BranchUtil.checkTestMode(context);
         getBranchInstance(context, !isTest, null);
+        new BranchPreinstall().getPreinstallSystemData(context);
```
apeterson-branch: Do we need to actually `new()` this class, or can the getPreinstallSystemData() be a static method?   By inspection it doesn't look like we are saving state anywhere.

apeterson-branch: Note that this appears to be doing "File" APIs, and will be run on the UI thread... which may violate some STRICT preferences.  I can't verify this though.

apeterson-branch: Why not pass in the newly created `Branch` object?   Instead of having the class immediately call `Branch::getInstance()` 3 times

I suspect that if Branch is null for some reason that we are not checking.   May not be possible though.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Created by --vbajpai on --24/07/2019 at --13:44 for --android-branch-deep-linking-attribution
+ */
+class BranchPreinstall {
+
+  private String SYSTEM_PROPERTIES_CLASS_KEY = "android.os.SystemProperties";
+  private String BRANCH_PREINSTALL_PROP_KEY = "ro.branch.preinstall.apps.path";
+
+  public void getPreinstallSystemData(Context context) {
```
apeterson-branch: Again, this should be static, and probably take a `Branch` as an argument

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,105 @@
...
+      path = (String) Class.forName(SYSTEM_PROPERTIES_CLASS_KEY)
+          .getMethod("get", String.class).invoke(null, BRANCH_PREINSTALL_PROP_KEY);
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+
+    return path;
+  }
+
+  private JSONObject readBranchFile(String branchFilePath) {
```
apeterson-branch: Is there any way we can write a unit test here?   I'm worried that the only way we are going to find issues is after a customer does a release with us baked in.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,105 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Created by --vbajpai on --24/07/2019 at --13:44 for --android-branch-deep-linking-attribution
+ */
+class BranchPreinstall {
+
+  private String SYSTEM_PROPERTIES_CLASS_KEY = "android.os.SystemProperties";
+  private String BRANCH_PREINSTALL_PROP_KEY = "ro.branch.preinstall.apps.path";
```
apeterson-branch: `ro` -- is this correct?

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+          } catch (JSONException e) {
+            e.printStackTrace();
+          }
+        }
+      }
+    }
+  }
+
+  private String checkForBranchPreinstallInSystem() {
```
apeterson-branch: This does not return `null` -- the code that calls this only checks for null.

In fact, `readBranchFile()` is throwing an exception and logging a nasty File error in the Android logs.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: Done

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: Done

Branch-SDK/src/io/branch/referral/Branch.java
code-crusher: Nice catch. It was coming, moved it to the thread.

code-crusher: Done, checking for null also.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: Done

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: Should be `io`

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: Some stack traces were printing on the catch. Now removed.

Branch-SDK/src/io/branch/referral/Branch.java
code-crusher: moved to static

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
csalmi-branch: @apeterson-branch what compiler warnings are you seeing?

csalmi-branch:
Once unit tests have been added, I'm good with the change.

code-crusher:
> Once unit tests have been added, I'm good with the change.

Added the unit test of the cases the preinstall can have.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: I think it was related to unused imports, I don't seem them anymore. In any case, these files are being used.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: Yes, done.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static String checkForBranchPreinstallInSystem() {
+        String path = null;
+        try {
+            path = (String) Class.forName(SYSTEM_PROPERTIES_CLASS_KEY)
+                    .getMethod("get", String.class).invoke(null, BRANCH_PREINSTALL_PROP_KEY);
+        } catch (Exception e) {
+            return null;
+        }
+        return path;
```
apeterson-branch: This doesn't return null.  It returns `""` empty string.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,114 @@
...
+class BranchPreinstall {
+
+    private static final String SYSTEM_PROPERTIES_CLASS_KEY = "android.os.SystemProperties";
+    private static final String BRANCH_PREINSTALL_PROP_KEY = "io.branch.preinstall.apps.path";
+
+    public static void getPreinstallSystemData(Branch branchInstance, Context context) {
+        if (branchInstance != null) {
+            // check if the SystemProperties has the branch file path added
+            String branchFilePath = checkForBranchPreinstallInSystem();
+            if (branchFilePath != null) {
```
apeterson-branch: Because the check returns the empty string, this succeeds.   I think you should call `if (!TextUtils.isEmpty(branchFilePath))`

code-crusher:
> I still think we need to check for an empty file as well as a null one.

For these two cases, I added the unit tests.

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
code-crusher: Thanks for the suggestion! :)

Branch-SDK/src/io/branch/referral/BranchPreinstall.java
```diff
@@ -0,0 +1,117 @@
...
+                    BufferedReader br = new BufferedReader(new FileReader(branchFile));
+                    String line;
+
+                    while ((line = br.readLine()) != null) {
+                        branchFileContent.append(line);
+                    }
+                    br.close();
+                    branchFileContentJson = new JSONObject(branchFileContent.toString().trim());
+
+                    if (!TextUtils.isEmpty(branchFileContentJson.toString())) {
```
code-crusher: @apeterson-branch I'm checking the empty file data here.

apeterson-branch: This check isn't strictly necessary anymore because it is now handled at a higher level.  It's ok to leave in though as code changes over time and this will make sure another code path doesn't creep in.

PR #726, opened 2019-08-01
SDK-421 allow custom server url

echo-branch:
Allow custom base server url.


Branch-SDK/src/io/branch/referral/PrefHelper.java
```diff
@@ -185,6 +200,10 @@ static void shutDown() {
      * API uses.
      */
     public String getAPIBaseUrl() {
+        if (URLUtil.isHttpsUrl(customServerURL_)) {
+            return customServerURL_;
+        }
```
apeterson-branch: I suspect the following lines will cause unit tests to fail if the `Build.VERSION.SDK_INT` is < 20

echo-branch: It should still pass.  the < 20 is for TLS level supported by the url.  I can't know that for custom urls.

echo-branch: Oh wait, you might be referring to running the test on an old device.  Yea that will fail.

PR #730, opened 2019-08-22
[SDK-211]: add support for modules + imei

code-crusher:
**Description:**

1. Add support for Branch SDK to consume additional Branch Modules
2. Support for IMEI module

**Testing:**
- [x] Unit Test
- [X] Manual Test

**Reviewers:**
@apeterson-branch @BranchMetrics/core-team

### DIFF
--- a/Branch-SDK/src/io/branch/referral/DeviceInfo.java
+++ b/Branch-SDK/src/io/branch/referral/DeviceInfo.java
@@ -106,9 +106,11 @@ class DeviceInfo {
                 requestObj.put(Defines.Jsonkey.LocalIP.getKey(), localIpAddr);
             }
 
-            String imei = SystemObserver.getImei(context_);
-            if (!isNullOrEmptyOrBlank(imei)) {
-                requestObj.put(ModuleNameKeys.imei.getKey(), imei);
+            if (PrefHelper.getInstance(context_).shouldAddModules()) {
+                String imei = SystemObserver.getImei(context_);
+                if (!isNullOrEmptyOrBlank(imei)) {
+                    requestObj.put(ModuleNameKeys.imei.getKey(), imei);
+                }
             }
         } catch (JSONException ignore) {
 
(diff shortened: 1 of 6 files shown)


Branch-SDK/src/io/branch/referral/DeviceInfo.java
```diff
@@ -105,6 +106,10 @@ void updateRequestWithV1Params(JSONObject requestObj) {
                 requestObj.put(Defines.Jsonkey.LocalIP.getKey(), localIpAddr);
             }
 
+            String imei = SystemObserver.getImei(context_);
```
apeterson-branch: Question:  Isn't this expensive?   Except for the very very few customers where this might be used, every single request has to go down the path of calling `getSecondaryRequestMetaData`, and doing a JSON lookup for a string... and checking to see if that succeeded -- all of which most of the time will fail.

It's all on a background thread but ... again, seems expensive.

Branch-SDK/src/io/branch/referral/PrefHelper.java
```diff
@@ -152,6 +157,7 @@ private PrefHelper(Context context) {
         this.prefsEditor_ = this.appSharedPrefs_.edit();
         this.requestMetadata = new JSONObject();
         this.installMetadata = new JSONObject();
+        this.secondaryRequestMetadata = new JSONObject();
```
apeterson-branch: This is just a comment:

If`secondaryRequestMetadata` were null to start, you might be able to short circuit a whole ton of checks with a simple null check.

For example `getSecondaryRequestMetaData()` could check for null instead of doing the JSON lookup.  You would only need to create it in `addSecondaryRequestMetadata()` -- which again, is called very very seldom called.

code-crusher:
@csalmi-branch @apeterson-branch added unit tests

code-crusher:
@kdama-branch looks the CI is failing to init the AVD, can you check?

kdama-branch:
@code-crusher  fixed the ci runner. CI tests should be running fine now.

code-crusher:
awesome @kdama-branch, thanks!

@apeterson-branch CI is passed now.

PR #741, opened 2019-09-26
Inteng 7461

bklastaitis-branch:
Addresses the fact that GAID is not `null`(i.e. not omitted from the payload) when limit ad tracking (LAT) is turned on, which was the core bug described in INTENG-7461 (please also see my comments in the ticket regarding LAT).

In addition, I reset GAID and LAT values on every app launch since it is possible to end up in a situation where user changes, say, their LAT preference while keeping the app in the background, then when the app is launched again, the SDK will behave as if LAT never changed. Sorry, I realize now that I should have created a separate ticket for this and I can do it + split up this PR, if it helps to archive and keep track of things. Let me know!
*******************************
Note, there is one more crazy edge case that I know will fail to update LAT/GAID references within the SDK and that is when the user is in a split screen mode (i.e. Settings/BranchSKD-enabled-App):

- user starts app with LAT disabled,
- switched settings app and changes LAT value,
- clicks on the app again.

The reason this will fail is because the app will not call `Activity.onCreate()` (where I refresh GAID) only `Activity.onResume()`, however IMHO this is better than refreshing  GAID on every `Activity.onResume()` because refreshing GAID creates and AsyncTask, which is arguably costly.

### DIFF
--- a/Branch-SDK/src/io/branch/referral/ServerRequest.java
+++ b/Branch-SDK/src/io/branch/referral/ServerRequest.java
@@ -388,6 +388,7 @@ public abstract class ServerRequest {
         BRANCH_API_VERSION version = getBranchRemoteAPIVersion();
         int LATVal = DeviceInfo.getInstance().getSystemObserver().getLATVal();
         String gaid = DeviceInfo.getInstance().getSystemObserver().getGAID();
+        String keyToBeUsedInAdIdsObject = null;
         try {
             if (version == BRANCH_API_VERSION.V2 || version == BRANCH_API_VERSION.V1_CPID ||
                     version == BRANCH_API_VERSION.V1_LATD) {
@@ -398,7 +399,7 @@ public abstract class ServerRequest {
                         userDataObj.put(Defines.Jsonkey.AAID.getKey(), gaid);
                         userDataObj.remove(Defines.Jsonkey.UnidentifiedDevice.getKey());
                         // for future use
-                        addAdvertisingIdsObject(userDataObj, gaid);
+                        keyToBeUsedInAdIdsObject = getKeyToBeUsedInAdIdsObject(userDataObj);
                     } else if (payloadContainsNoDeviceIdentifiers(userDataObj)) {
                         userDataObj.put(Defines.Jsonkey.UnidentifiedDevice.getKey(), true);
                     }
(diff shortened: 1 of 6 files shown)


Branch-SDK/src/io/branch/referral/ServerRequest.java
```diff
@@ -388,41 +388,57 @@ private void updateGAdsParams() {
         BRANCH_API_VERSION version = getBranchRemoteAPIVersion();
         int LATVal = DeviceInfo.getInstance().getSystemObserver().getLATVal();
         String gaid = DeviceInfo.getInstance().getSystemObserver().getGAID();
-        if (!TextUtils.isEmpty(gaid)) {
-            try {
-                if (version == BRANCH_API_VERSION.V2 || version == BRANCH_API_VERSION.V1_CPID ||
-                        version == BRANCH_API_VERSION.V1_LATD) {
-                    JSONObject userDataObj = params_.optJSONObject(Defines.Jsonkey.UserData.getKey());
-                    if (userDataObj != null) {
+        String keyToBeUsedInAdIdsObject = null;
```
apeterson-branch: Let's put this back, and open up a new ticket to refactor this block of code.

Branch-SDK/src/io/branch/referral/ServerRequest.java
```diff
@@ -388,41 +388,57 @@ private void updateGAdsParams() {
...
+            if (keyToBeUsedInAdIdsObject == null) return;
+
+            JSONObject advertisingIdsObject = new JSONObject().put(keyToBeUsedInAdIdsObject, gaid);
+            params_.put(Defines.Jsonkey.AdvertisingIDs.getKey(), advertisingIdsObject);
+        } catch (JSONException e) {
+            e.printStackTrace();
         }
     }
+
+    private boolean payloadContainsNoDeviceIdentifiers(JSONObject payload) {
```
apeterson-branch: Can we reverse this logic to say `payloadContainsDeviceIdentifiers()` and do the "not" check just once?

Branch-SDK/src/io/branch/referral/ServerRequest.java
bklastaitis-branch: I couldn't reasonably make the required changes to this method without drastically changing it, even if I kept the existing duplicated code, so I made a gist that explains step by step by how the method morphed into this new version and what REAL changes were made in there vs. just the cleanup changes.
https://gist.github.com/bklastaitis-branch/57c73799d4c1f4144e8bae48122da11a

apeterson-branch: That's awesome!

PR #746, opened 2019-10-07
Release v4.1.1

bklastaitis-branch:
Includes changes for:
SDK-516
INTENG-7461

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,4 +1,10 @@
 # Branch Android SDK change log
+- v4.1.1
+  * _*Master Release*_ - October 7, 2019
+  * Support for Fire OS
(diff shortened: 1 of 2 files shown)


ChangeLog.md
```diff
@@ -1,4 +1,10 @@
 # Branch Android SDK change log
+- v4.1.1
+  * _*Master Release*_ - October 7, 2019
+  * Support for Fire OS
+  * Do not collect advertising ID when limit ad tracking is enabled
```
csalmi-branch: Can you update this to "Do not collect advertising ID when limit ad tracking is enabled after the app was initialized"? I want to be clear in the notes that it was set to not collect that before, but there was a bug with the feature caching that setting.

PR #750, opened 2019-10-10
SDK-434

bklastaitis-branch:
Carriers may want to whitelist Branch endpoints and "free-rate" them. To do that they need the whitelisted endpoints + a way to set Branch SDK to use those endpoints. There was already a way to set the API host name but we were missing a way to set the CDN host name, so this PR does just that.


bklastaitis-branch:
Hey @apeterson-branch, would you 👍 this again (assuming the description is sufficient)?

PR #751, opened 2019-10-14
SDK-561

bklastaitis-branch:
Share sheet buttons were not clickable on OTT devices.


bklastaitis-branch:
@apeterson-branch thanks so much for catching it! Somehow setting the child to be clickable made the parent not clickable.... (i.e. culprit was `listItemView.setClickable(true)`)

PR #754, opened 2019-10-17
SDK-581/close-field-gaps

bklastaitis-branch:
I'll post the below to Jira as well:

Added the following fields. It’s not the full list as per ticket description, please see notes below.
```
CPUType("cpu_type"),
DeviceBuildId("build"),
Locale("locale"),
ConnectionType("connection_type"),
DeviceCarrier("device_carrier"),
PluginType("plugin_type"),
PluginVersion("plugin_version"),
```

Note 1: `os_version_android` can be derived on the backend by reusing the value from `os_version` whenever the field `os` equals `Android`/`AMAZON_FIRE`/`AMAZON_FIRE_TV`. I think that is desirable over the alternative of duplicating the same value in two different fields `os_version` and `os_version_android` (we send a few less bits and that’s good for the footprint)

Note 2: `user_data_http_referrer` is part of the “Click” procedure, which is independent of the SDK, so I didn’t touch that. However, it would seem to me that `http_referrer` is already recorded because, upon a successful conversion (i.e. app install via branch link), the response to the install request contains a value like the following:
`”~referring_link": "https://branchster.app.link/41KBvaheQ0?_branch_match_id=637423416087744731&referrer=link_click_id%253D637423416087744731%2526utm_source%253DBranch%2526utm_medium%253Dmarketing%2526adjust_external_click_id%253D713115186409305202”,`

Note 3: `user_data_device_type` pretty much equals `user_data_platform`, which is not used on the SDK side and is only added on the traffic between the backend and db/postbacks/etc. The sets of values observed in the two fields are not the same but I think it's up to backend to make them the same. If SDK needs to provide any data to make it possible, please let me know!

Note 4: `store_install_begin_timestamp` is already sent up as `install_begin_ts` and `referrer_click_timestamp` is sent up as `clicked_referrer_ts`

Note 5: `customer_event_alias` was already added as a week or so ago, separate of my efforts here.


bklastaitis-branch:
@apeterson-branch I did not add any unit tests. I could make sure that only the install/open/event requests contain said values and that they in fact do contain them but I didn't see similar tests for all the other fields (besides GAID/LAT)... Let me know if those unit tests would be useful.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -2153,15 +2153,15 @@ public void userCompletedAction(final String action, BranchViewHandler.
...
+        ServerRequest req = new ServerRequestActionCompleted(context_, action, null, metadata, callback);
         if (!req.constructError_ && !req.handleErrors(context_)) {
             handleNewRequest(req);
         }
     }
     
     public void sendCommerceEvent(@NonNull CommerceEvent commerceEvent, JSONObject
             metadata, BranchViewHandler.IBranchViewEvents callback) {
-        ServerRequest req = new ServerRequestRActionCompleted(context_, commerceEvent, metadata, callback);
+        ServerRequest req = new ServerRequestActionCompleted(context_, "purchase", commerceEvent, metadata, callback);
```
apeterson-branch: Why isn't this using BRANCH_STANDARD_EVENT::PURCHASE?

In fact, why is there a CurrencyEvent at all?  Everything here except the List<Product> is 1-1 mapped in the base Event class.

Branch-SDK/src/main/java/io/branch/referral/DeviceInfo.java
```diff
@@ -91,6 +91,21 @@ void updateRequestWithV1Params(JSONObject requestObj) {
 
             requestObj.put(Defines.Jsonkey.OSVersion.getKey(), SystemObserver.getOSVersion());
 
+            String[] applicableRoutes = new String[]{"install", "open", "event"};
```
apeterson-branch: This is cut-paste.  Should this be pulled out?

Branch-SDK/src/main/java/io/branch/referral/DeviceInfo.java
```diff
@@ -153,6 +168,21 @@ void updateRequestWithV2Params(Context context, PrefHelper prefHelper, JSONObjec
...
```
apeterson-branch: This is cut-paste.  Should this be pulled out?

Branch-SDK/src/main/java/io/branch/referral/DeviceInfo.java
apeterson-branch: Having these as bare strings doesn't seem right.   There are other similar strings in ServerRequestInitSession.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestActionCompleted.java
```diff
@@ -39,17 +43,20 @@ public ServerRequestActionCompleted(Context context, String action, JSONObject m
...
+            }
             updateEnvironment(context, post);
             setPost(post);
         } catch (JSONException ex) {
             ex.printStackTrace();
             constructError_ = true;
         }
 
-        if (action != null && action.equalsIgnoreCase("purchase")) {
+        if (action != null && action.equalsIgnoreCase("purchase") && commerceEvent == null) {
```
apeterson-branch: Looking at a bare string -- maybe this should be a define somewhere.

Branch-SDK/src/main/java/io/branch/referral/SystemObserver.java
```diff
@@ -265,6 +267,68 @@ static int getOSVersion() {
         return android.os.Build.VERSION.SDK_INT;
     }
 
+    /**
```
apeterson-branch: Needs PII review

Branch-SDK/src/main/java/io/branch/referral/BranchUtil.java
```diff
@@ -36,6 +36,26 @@
...
+        Unity,
+        Corona,
+        Xamarin,
+        ReactNative,
+        Titanium,
+        mParticle,
+        AdobeLaunch,
+        AdobeLaunchMobileUI,
+        Segment,
+        Electron
```
csalmi-branch: Electron can be removed. Its actually a SDK (one we haven't released yet), not a plugin.

Cordova and Ionic are the same thing (remove Ionic, since "Ionic" is also coming out with Capacitor now)

Branch-SDK/src/main/java/io/branch/referral/Branch.java
bklastaitis-branch: You probably meant `CommerceEvent` not `CurrencyEvent` and I agree it would be simpler to have a single event class but it's not part of this PR. A separate refactoring PR would be probably be the best way to do it. (I did include some refactoring here but tried to keep it at a minimum so it doesn't interfere with the core thing of this PR, closing the field gaps)

PR #783, opened 2019-12-21
Sdk 691/init session forced

bklastaitis-branch:


Branch-SDK/src/main/java/io/branch/referral/Defines.java
```diff
@@ -100,9 +98,7 @@
         DeepLinkPath(Branch.DEEPLINK_PATH),
         
         AndroidAppLinkURL("android_app_link_url"),
-        AndroidPushNotificationKey("branch"),
```
apeterson-branch: I'm not opposed to these, but this will cause a major version bump.  Let's make sure we get sign off from PM.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestRegisterOpen.java
```diff
@@ -46,7 +46,9 @@ public void onPreExecute() {
         super.onPreExecute();
         // Instant Deep Link if possible
         if (Branch.getInstance().isInstantDeepLinkPossible()) {
-            callback_.onInitFinished(Branch.getInstance().getLatestReferringParams(), null);
+            if (hasCallBack()) {
```
apeterson-branch: Question:  Calling the method `hasCallback()` and then using the `callback_` member is non-optimal.   Can we instead have a stub (non-null) callback that checks to see if the caller passed one in, and call it if necessary?   It's a small thing...   but it turns out that every request class has the previous way and not all of them were changed.

Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
```diff
@@ -60,8 +60,9 @@ private void onTrackingDisabled(Context context) {
     }
     
     private void onTrackingEnabled() {
-        if (Branch.getInstance() != null) {
-            Branch.getInstance().registerAppInitWithoutIntent();
+        Branch b = Branch.getInstance();
+        if (b != null) {
```
apeterson-branch: We should avoid one letter variable names, even when it is obvious what is going on here.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestRegisterOpen.java
bklastaitis-branch: > but it turns out that every request class has the previous way and not all of them were changed.

Hmm.. every request I that see has my implementation, where we first check if callback is null (aka we call hasCallBack) and then we either invoke the callback or not.
See the implementation in:
 - ServerRequestRedeemRewards
 - ServerRequestRegisterInstall
 - ServerRequestGetRewards
 - ServerRequestIdentifyUserRequest
 - ServerRequestGetLATD
 - ServerRequestGetCPID (this one actually has the potential to throw an NPE, created SDK-756)

But if we were gonna go ahead with implementation you are describing, I fail to see how it is more optimal.. Do I understand you right (see below)?
```
Request {
    @NonNull StubCallback stubCallback = new StubCallback();
    Request(@Nullable BranchReferralInitListener c) {
        this.stubCallback = new StubCallback(c);
    }
    ...
    @Override public void onRequestSucceeded() {
        stubCallback.onInitFinished()
    }
    ...
    public static class StubCallback extends BranchReferralInitListener {
        private BranchReferralInitListener internalCallback;
        public StubCallback(@Nullable BranchReferralInitListener callback) {
            this.internalCallback = callback;
        }
       @Override public void onInitFinished() {
           if (internalCallback != null) {
               internalCallback.onInitFinished();
           }
       }
    }
}
```

Branch-SDK/src/main/java/io/branch/referral/Defines.java
bklastaitis-branch: @csalmi-branch @derrickstaten, these were basically publicly accessible (but NOT advertised) strings that we use as keys in an Android intent. They were incorrectly classified as `JsonKeys` instead of `IntentKeys`. Is it OK to go ahead with this change and make them `IntentKeys`? (if any of the users utilized these, then upon updating to the new SDK version, their builds would break at compile time)

Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
bklastaitis-branch: 👍

derrickstaten:
Is there any way to not break our users' code, if they were utilizing this? I'm strongly opposed to making changes that cause users' code to break.

Branch-SDK/src/main/java/io/branch/referral/Defines.java
derrickstaten: Is there any way to not break our users' code, if they were utilizing this? I'm generally opposed to making changes that cause users' code to break.

If not, please proceed.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestRegisterOpen.java
apeterson-branch: I am objecting only to the introduction of `hasCallback()` -- checking the local variable in that method, then using the local variable directly.   There is no advantage there (and is different than everywhere else) where you just say
```
if (callback_ != null) {
  callback_.doSomething();
}
```

However, if you *are* going to encapsulate the existence of a callback, then go all the way by making a stub and having the logic in there.   The advantage of doing that is that all of the other things that you do internally would be done in the stub as well.

apeterson-branch: Consider these two patterns:
```
if (callback_ != null) {
  callback_.doSomething();
}
```

and

```
if (hasCallback()) {
  callback_.doSomething();
}
```

I maintain the second one is not a good pattern.

bklastaitis-branch: Gotcha, that makes sense, I will revert to `callback_ != null`

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -1720,10 +1729,11 @@ public void getCrossPlatformIds(ServerRequestGetCPID.BranchCrossPlatformIdListen
      *                 to callback with last attributed touch data
      *
      */
-    public void getLastAttributedTouchData(BranchLastAttributedTouchDataListener callback) {
+    public void getLastAttributedTouchData(@NonNull BranchLastAttributedTouchDataListener callback) {
         if (context_ != null) {
             handleNewRequest(new ServerRequestGetLATD(context_, Defines.RequestPath.GetLATD.getPath(), callback));
         }
+        Defines.IntentKeys.ForceNewBranchSession.getKey();
```
apeterson-branch: This probably shouldn't be here.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -3650,6 +3670,10 @@ public void reInit() {
             init();
         }
     }
+
+    boolean isIDLSession() {
+        return "true".equals(instrumentationExtraData_.get(Defines.Jsonkey.InstantDeepLinkSession.getKey()));
+    }
```
apeterson-branch: instead of doing a string comparison here, can we do a Boolean parse test?   That way, just in case someone uses "True", "true", or "TRUE" it will still work.

Branch-SDK/src/main/java/io/branch/referral/BranchActivityLifecycleObserver.java
```diff
@@ -17,7 +19,10 @@
  * session.</p>
  */
 class BranchActivityLifecycleObserver implements Application.ActivityLifecycleCallbacks {
-    private int activityCnt_ = 0; //Keep the count of live  activities.
+    private int activityCnt_ = 0; //Keep the count of visible activities.
+
+    /* Set of activities observed in this session */
+    private Set<String> activitiesOnStack = new HashSet<>();
```
apeterson-branch: class variables in here appear to have trailing `_`

Branch-SDK/src/main/java/io/branch/referral/BranchActivityLifecycleObserver.java
```diff
@@ -66,6 +73,10 @@ public void onActivityResumed(@NonNull Activity activity) {
             // and the entry Activity is not the launcher Activity where user placed initSession themselves.
             Branch.sessionBuilder(activity).init();
         }
+
+        // must be called after session initialization, which relies on checking whether activity
+        // that is initializing the session is being launched from stack or anew
+        activitiesOnStack.add(activity.getLocalClassName());
```
apeterson-branch: Activities are added, but never removed.   Further, It's not obvious from the ticket why this is being addressed in this PR.

Branch-SDK/src/main/java/io/branch/referral/BranchActivityLifecycleObserver.java
```diff
@@ -123,4 +134,11 @@ private void maybeRefreshAdvertisingID(Context context) {
             branch.setGAParamsFetchInProgress(branch.getDeviceInfo().getSystemObserver().prefetchAdsParams(context, branch));
         }
     }
+
+    // default is true
+    boolean isCurrentActivityLaunchedFromStack() {
+        Branch branch = Branch.getInstance();
+        if (branch == null || branch.getCurrentActivity() == null) return true;
+        return activitiesOnStack.contains(branch.getCurrentActivity().getLocalClassName());
```
apeterson-branch: The semantics here are unclear to me.   If the current activity == null, why is it considered "launched from stack".

Branch-SDK/src/main/java/io/branch/referral/BranchActivityLifecycleObserver.java
```diff
@@ -123,4 +134,11 @@ private void maybeRefreshAdvertisingID(Context context) {
             branch.setGAParamsFetchInProgress(branch.getDeviceInfo().getSystemObserver().prefetchAdsParams(context, branch));
         }
     }
+
+    // default is true
+    boolean isCurrentActivityLaunchedFromStack() {
+        Branch branch = Branch.getInstance();
+        if (branch == null || branch.getCurrentActivity() == null) return true;
```
apeterson-branch: Braces required per coding guideline: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

Branch-SDK/src/main/java/io/branch/referral/ServerRequest.java
```diff
@@ -244,6 +245,10 @@ public JSONObject getPostWithInstrumentationValues(ConcurrentHashMap<String, Str
                 Set<String> keys = instrumentationData.keySet();
                 try {
                     for (String key : keys) {
+                        String value = instrumentationData.get(key);
+                        if (Defines.Jsonkey.InstantDeepLinkSession.getKey().equals(key) && "true".equals(value)) {
```
apeterson-branch: Can we use a Boolean comparison operator here instead of a string comparison

Branch-SDK/src/main/java/io/branch/referral/ServerRequestGetCPID.java
```diff
@@ -30,10 +32,10 @@ public boolean handleErrors(Context context) {
 
     @Override
     public void onRequestSucceeded(ServerResponse response, Branch branch) {
+        if (callback == null) return;
```
apeterson-branch: Braces required per coding guideline: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

Branch-SDK/src/main/java/io/branch/referral/ServerRequestGetCPID.java
```diff
@@ -43,6 +45,8 @@ public void onRequestSucceeded(ServerResponse response, Branch branch) {
 
     @Override
     public void handleFailure(int statusCode, String causeMsg) {
+        if (callback == null) return;
```
apeterson-branch: Braces required per coding guideline: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

Branch-SDK/src/main/java/io/branch/referral/ServerRequestGetLATD.java
```diff
@@ -41,10 +41,10 @@ public boolean handleErrors(Context context) {
 
     @Override
     public void onRequestSucceeded(ServerResponse response, Branch branch) {
+        if (callback == null) return;
```
apeterson-branch: Braces required per coding guideline: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

Branch-SDK/src/main/java/io/branch/referral/ServerRequestRegisterOpen.java
```diff
@@ -123,7 +125,7 @@ public void handleFailure(int statusCode, String causeMsg) {
     @Override
     public boolean handleErrors(Context context) {
         if (!super.doesAppHasInternetPermission(context)) {
-            if (hasCallBack()) {
+            if (callback_ != null && !Branch.getInstance().isIDLSession()) {
                 callback_.onInitFinished(null, new BranchError("Trouble initializing Branch.", BranchError.ERR_NO_INTERNET_PERMISSION));
             }
             return true;
```
apeterson-branch: the method `hasCallback()` is still  in this file and should be removed.

Branch-SDK/src/main/java/io/branch/referral/BranchActivityLifecycleObserver.java
```diff
@@ -35,6 +40,10 @@ public void onActivityStarted(@NonNull Activity activity) {
         Branch branch = Branch.getInstance();
         if (branch == null) return;
```
apeterson-branch: Braces required per coding guideline: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

Branch-SDK/src/main/java/io/branch/referral/BranchActivityLifecycleObserver.java
bklastaitis-branch: This is part of SDK-672, fixing the instant deeplinking, I linked the two issues on Jira. This was a good catch though because, when app is in background and being reopened via a deeplink or home screen icon, Activity does get recreated apparently, so I added removal of activities to onCreate.

Branch-SDK/src/main/java/io/branch/referral/BranchActivityLifecycleObserver.java
bklastaitis-branch: In reality, this should never happen but you are right, semantically, it should return false in that case. 👍

bklastaitis-branch:
I'm gonna close this PR because we decided with @derrickstaten to hold off from major changes like this until after we release a stable build with Play Store Referrer lib bundled in.

PR #784, opened 2020-01-03
SDK-723

bklastaitis-branch:
Added annotation to indicate returned object nullability in branch init session listeners because, if kotlin misspecifies them, the listener will not be invoked.


bklastaitis-branch:
I agree regarding both points. Will see if we can add Kotlin unit tests to the Branch-SDK-TestBed module (because I believe we can't add Kotlin support to the SDK module without breaking things for clients that don't use Kotlin)

PR #869, opened 2020-10-26
[SDK-1077]: add request when available in enableLogging

code-crusher:
## Reference
SDK-1077: [android] Update native SDK to report X-Branch-Request-ID when present

## Description
Print request-id in the logs when;
1. EnableLogging is set
2. X-Branch-Request-Id-Header is present in the response header

```
I/BranchSDK: returned request-id: [ 1231231__!32123123 ] with {"session_id":"849308671259384261","identity_id":"849308509983354059","link":"https://bnctestbed.test-app.link?%24identity_id=849308509983354059","data":"{\"+clicked_branch_link\":false,\"+is_first_session\":false}","device_fingerprint_id":"849308509959273715"}
```

## Testing Instructions
- [X] Manual

## Risk Assessment [`LOW`]

- Correctness & Style

cc @BranchMetrics/core-team

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,4 @@
-version: 2.1
-
-orbs:
-  android: circleci/android@0.2.1
+version: 2
 
 jobs:
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,8 @@
-version: 2
+version: 2.1
+
+orbs:
+  android: circleci/android@0.2.1
+
 jobs:
   build:
     working_directory: ~/code
@@ -6,12 +10,21 @@ jobs:
       - image: circleci/android:api-28
     steps:
       - checkout
+      - restore_cache:
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
       - run:
-          name: Update SDK Tools
-          command: $ANDROID_HOME/tools/bin/sdkmanager --update
+          name: Download Dependencies
+          command: ./gradlew androidDependencies
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
       - run:
           name: Run Lint
           command: ./gradlew lint
+      - run:
+          name: Run Test
+          command: ./gradlew test
       - run:
           name: Run Build
           command: ./gradlew build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,14 +11,14 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "Branch-SDK/build.gradle" }}
       - run:
           name: Download Dependencies
           command: ./gradlew androidDependencies
       - save_cache:
           paths:
             - ~/.gradle
-          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "Branch-SDK/build.gradle" }}
       - run:
           name: Run Lint
           command: ./gradlew lint
--- a/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
@@ -1,6 +1,5 @@
 package io.branch.referral;
 
-import androidx.annotation.Nullable;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
--- a/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
@@ -35,15 +35,21 @@ public class ServerResponse {
     private Object post_;
 
     /**
+     * Unique timestamp based request id provided from internal Branch servers for debugging
+     */
+    private String requestId_;
+
+    /**
      * <p>Main constructor method for the {@link ServerResponse} class that allows for the instantiation
      * of a server response object as a direct result of a server call.</p>
      *
      * @param tag        A {@link String} value of the <i>Tag</i> attribute of the current link.
      * @param statusCode {@link Integer} value of the HTTP status code.
      */
-    public ServerResponse(String tag, int statusCode) {
+    public ServerResponse(String tag, int statusCode, String requestId) {
         tag_ = tag;
         statusCode_ = statusCode;
+        requestId_ = requestId;
     }
 
     /**
--- a/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
@@ -1,5 +1,6 @@
 package io.branch.referral;
 
+import androidx.annotation.Nullable;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
@@ -33,6 +34,11 @@ public class ServerResponse {
      * cast to a type before use to allow the appropriate "get" methods to be used.
      */
     private Object post_;
+
+    /**
+     * Unique timestamp based request id provided from internal Branch servers for debugging
+     */
+    private String requestId_;
 
 
     /**
@@ -106,6 +112,14 @@ public class ServerResponse {
         return null;
     }
 
+    /**
+     * Set the passed request id
+     *
+     * @param requestId_ passed in from the response headers
+     */
+    public void setRequestId_(@Nullable String requestId_) {
+        this.requestId_ = requestId_;
+    }
 
     /**
      * Get the reason for failure if there any
--- a/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
@@ -36,12 +36,6 @@ public class ServerResponse {
     private Object post_;
 
     /**
-     * Unique timestamp based request id provided from internal Branch servers for debugging
-     */
-    private String requestId_;
-
-
-    /**
      * <p>Main constructor method for the {@link ServerResponse} class that allows for the instantiation
      * of a server response object as a direct result of a server call.</p>
      *
@@ -110,15 +104,6 @@ public class ServerResponse {
         }
 
         return null;
-    }
-
-    /**
-     * Set the passed request id
-     *
-     * @param requestId_ passed in from the response headers
-     */
-    public void setRequestId_(@Nullable String requestId_) {
-        this.requestId_ = requestId_;
     }
 
     /**
--- a/Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
@@ -8,6 +8,8 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.Locale;
+
 import io.branch.referral.Branch;
 import io.branch.referral.BranchError;
 import io.branch.referral.BuildConfig;
@@ -195,7 +197,7 @@ public abstract class BranchRemoteInterface {
 
         ServerResponse result = new ServerResponse(tag, statusCode, requestId);
         if(!TextUtils.isEmpty(requestId)){
-            PrefHelper.Debug(String.format("returned request-id: [ %s ] with %s", requestId, responseString));
+            PrefHelper.Debug(String.format(Locale.getDefault(), "Server returned: [%s] Status: [%d]; Data: %s", requestId, statusCode, responseString));
         } else {
             PrefHelper.Debug(String.format("returned %s", responseString));
         }
--- a/Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
@@ -90,20 +90,25 @@ public abstract class BranchRemoteInterface {
         if (addCommonParams(params, branchKey)) {
             modifiedUrl += this.convertJSONtoString(params);
         } else {
-            return new ServerResponse(tag, BranchError.ERR_BRANCH_KEY_INVALID);
+            return new ServerResponse(tag, BranchError.ERR_BRANCH_KEY_INVALID, "");
         }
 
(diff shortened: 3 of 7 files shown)


jdee:
So I started to do this sort of thing for iOS, adding a `requestId` field to a request object somewhere. But I ended up going with a two-line change that just dumps all response headers to the log. For the Web SDK, we'll probably make no change, since you can just see the header in the built-in browser tools. I wonder what the best approach is here. @bklastaitis-branch any opinions?

code-crusher:
Just dumping all the headers is not really UX friendly(and also we only need a single header to be printed), the above format will clearly specify what is the request ID connected with the response.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
```diff
@@ -116,6 +117,7 @@ private BranchResponse doRestfulGet(String url, int retryNumber) throws BranchRe
 
     private BranchResponse doRestfulPost(String url, JSONObject payload, int retryNumber) throws BranchRemoteException {
         HttpsURLConnection connection = null;
+        String requestId = null;
```
bklastaitis-branch, line 120: Variable declaration can be pulled into the try block (where value is assigned) for brevity.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
```diff
@@ -203,14 +207,23 @@ private BranchResponse doRestfulPost(String url, JSONObject payload, int retryNu
...
         } finally {
             if (connection != null) {
                 connection.disconnect();
             }
         }
     }
 
+    @Nullable
+    private String getRequestIdFromHeader(@NonNull HttpsURLConnection connection) {
+        if (!connection.getHeaderFields().isEmpty() && connection.getHeaderFields()
```
bklastaitis-branch, line 220: We don't need this method, it can be safely replaced with a oneliner: `connection.getHeaderField("X-Branch-Request-Id");` which does the same thing as the method. However, the string `"X-Branch-Request-Id"` should be pulled out and defined as one of the values in a new enum `Headers` in the `Defines` class.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
```diff
@@ -51,6 +49,7 @@ public BranchResponse doRestfulPost(String url, JSONObject payload) throws Branc
     ///-------------- private methods to implement RESTful GET / POST using HttpURLConnection ---------------//
     private BranchResponse doRestfulGet(String url, int retryNumber) throws BranchRemoteException {
         HttpsURLConnection connection = null;
+        String requestId = null;
```
bklastaitis-branch, line 52: Same as below: Variable declaration can be pulled into the try block (where value is assigned) for brevity.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
```diff
@@ -178,9 +179,14 @@ public static final BranchRemoteInterface getDefaultBranchRemoteInterface(Contex
      * response in Branch SDK terms.
      * see <a href="http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html">HTTP/1.1: Status Codes</a>
      */
-    private ServerResponse processEntityForJSON(String responseString, int statusCode, String tag) {
+    private ServerResponse processEntityForJSON(String responseString, int statusCode, String tag,@Nullable String requestId) {
```
bklastaitis-branch, line 182: Can we replace all arguments in `processEntityForJSON` with just `BranchResponse` and `tag`? This way it's clear that `processEntityForJSON` converts from the `BranchResponse` abstraction to json-formatted `ServerResponse`. Also, for brevity, let's get rid of `serverResponse.setRequestId_` and assign that value in the constructor, so it becomes a read-only variable, just like `ServerResponse.statusCode_`. In order to avoid empty strings we can use `Strings.emptyToNull()` when passing the request id to to the `ServerResponse` constructor

Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
```diff
@@ -106,6 +112,14 @@ public JSONArray getArray() {
         return null;
     }
 
+    /**
+     * Set the passed request id
+     *
+     * @param requestId_ passed in from the response headers
+     */
+    public void setRequestId_(@Nullable String requestId_) {
```
bklastaitis-branch, line 120: Let's get rid of this, see the comment on `processEntityForJSON` for more info

.circleci/config.yml
```diff
@@ -1,17 +1,27 @@
 version: 2
+
 jobs:
   build:
     working_directory: ~/code
     docker:
       - image: circleci/android:api-28
     steps:
       - checkout
+      - restore_cache:
```
bklastaitis-branch, line 10: I assume changes in this file were needed but I'm not really following what they. Would you describe why you added them? PR description/Jira ticket didn't mention anything.

bklastaitis-branch (changes requested):
@code-crusher I left some comments, requested mainly stylistic changes but I think they are important. It's not necessarily related to your code but the SDK is full of little design quirks like that, which make it harder to read and understand the code.

bklastaitis-branch:
@jdee I'd lean towards @code-crusher 's approach. UX wise, the reader of the logs is most likely only interested in the request id, not any other headers, so printing other headers just makes it harder to read/find the request id.

As for creating the request id entity in the first plcase - rather than just printing and forgetting - is fair game I think. We may one day use that request id, especially for things like fraud detection.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
code-crusher: Fair

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
code-crusher: 1. Yeah, let's do Object here, did not wanted change much but will update.
2. Already using `!TextUtils.isEmpty(requestId)` -- this checks for both null and empty. Not sure how using `Strings.emptyToNull()` will help since we have to check for null in any case, the first method does both.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
code-crusher: Yeah, missed the constant, will update

.circleci/config.yml
code-crusher: The CI was failing, we were using an old and incorrect config. This is just basic CI config for Android. Cache will make CI builds faster since the same dependecies will not be pulled again and again.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
code-crusher: I remember this now, this is being used outside of try-catch also, that's why put it here.

Branch-SDK/src/main/java/io/branch/referral/ServerResponse.java
code-crusher: I created this to keep usages in the future if someone want to pull thisi id apart from the logs, but I'm fine to remove it

code-crusher:
@bklastaitis-branch updated, if this looks good, can we merge?

ahmednawar:
Please check that Android and iOS are consistent in how they show this ID to developers. cc @jdee

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
bklastaitis-branch: My bad didn't notice it

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
bklastaitis-branch: @code-crusher this seems to be using a String literal still

bklastaitis-branch:
@code-crusher commented back, I think there are still two things left `ServerResponse.requestId` and `Defines.Headers.RequestId("X-Branch-Request-Id")`.

However, as per Ahmed's comment, it seems like we should get in line between iOS and Android in how we log this. So if we happen to go the iOS route, then my feedback would be irrelevant. I'll leave it to you two decide which route is better but I left my 2 cents earlier.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
code-crusher: Dunzo

code-crusher:
@jdee doe the iOS prints in the above format I have put in the description? If not can you share one?

code-crusher:
@bklastaitis-branch good to merge?

bklastaitis-branch:
@code-crusher I was waiting on @jdee response regarding the format, but besides that, I think you still haven't addressed the comment about `ServerResponse.requestId`, let me know if doesn't make sense to you, or otherwise, if you disagree with that comment, happy to chat.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
code-crusher: > @code-crusher I was waiting on @jdee response regarding the format, but besides that, I think you still haven't addressed the comment about `ServerResponse.requestId`, let me know if doesn't make sense to you, or otherwise, if you disagree with that comment, happy to chat.

Used `Strings.emptyToNull(requestId)` -- just doesn't feel okay to add another step of processing, but it's fine with me.

jdee:
Sorry I missed recent updates to this conversation. I've been updating the [tech spec](https://www.notion.so/brancheng/Request-ID-6c3c5c8525144a0cbae0e8da5567abc9) in Notion. Here's an example iOS log message.

```
2020-10-27 14:26:22.673178-0700 Branch-TestBed[61896:286008] [branch.io] BNCServerInterface.m(471) Debug: Server returned: [1325e434fa294d3bb7d461349118602d-2020102721] Status: 200; Data: {
    data = "{\"+clicked_branch_link\":false,\"+is_first_session\":false}";
    "device_fingerprint_id" = 848339731384240701;
    "identity_id" = 848339731459733646;
    link = "https://itrdp-jdee.branchbeta.link?%24identity_id=848339731459733646";
    "session_id" = 849748190046577385;
}.
```

code-crusher:
@bklastaitis-branch looks good to merge?

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
bklastaitis-branch: I think we are talking past each other @code-crusher 
> requestId should be a field within ServerResponse the same way it's a field in BranchResponse

I was saying that `ServerResponse` needs to have `requestId` passed in its constructor, as well as `BanchResponse`.. (I don't even know why we have `BanchResponse` in the first place, seems like we could just parse json directly to `ServerResponse`, but we can cross that bridge when we get there.)

bklastaitis-branch:
@code-crusher https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/869#discussion_r518247942

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterface.java
code-crusher: I don't see why that class needs to have this when BranchResponse is already passing it. What's the benefit of it that I'm missing?

bklastaitis-branch: > I still think that's the right thing to do because BranchResponse is just an intermediate class, if requestId was ever to be used in some other way besides logging, it would prolly would be referenced via ServerResponse instance which is further downstream.
...
> (I don't even know why we have BanchResponse in the first place, seems like we could just parse json directly to ServerResponse, but we can cross that bridge when we get there.)

code-crusher: But that's what I did earlier https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/869/commits/f8fa8e3724a88f47b73205a1c1cedf3bdc1fa772 and mentioned the same: https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/869#discussion_r513480024

code-crusher: Honestly, this is a too small change to think so much on this. I'll add the request id back into ServerResponse and update the PR

bklastaitis-branch: Sorry for appearing picky, I think we just keep talking past each other.

I know you did that earlier and I agreed that it was a good idea, my [comment](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/869#discussion_r513480024) that you linked to was specifically about removing the setter and instead having the value passed in the constructor of `ServerResponse`.
>get rid of serverResponse.setRequestId_ and assign that value in the constructor, so it becomes a read-only variable, just like ServerResponse.statusCode_

code-crusher:
@bklastaitis-branch added

bklastaitis-branch (approved):
@code-crusher thanks! FYI I added one more commit to make the print line identical to iOS (minus json formatting, which is present on iOS), hope that's fine. This is good to merge now but I'll leave that to you to make sure you saw my last message about the extra commit :)

code-crusher:
Did you remove the request ID keyword too? I'm not really in favour of that since I don't see how any developer outside of Branch really know about it. But in any case, let's merge this and get it out.

echo-branch:
I know this is a bit late but in the future I would strongly prefer to print raw server responses in verbose logging. Any effort to "pretty print" what we think is most useful is of limited value over raw logs and just complicates our code.

1. The raw response requires less code and less maintenance.  Every new header or field in responses will be logged automatically.
2. By the time verbose logging is enabled, the client is just going to copy paste the logs to us.  Client readability matters a lot less at that point.  While having more data is always useful to us.
3. Raw logs can be quickly tested via curl or postman.  This is of great value when debugging.  Pretty printed versions may omit data that is required to repro issues.

PR #876, opened 2020-11-19
Core 1355/reduce calling close api

bklastaitis-branch:
## Reference
[CORE-1355](https://branch.atlassian.net/browse/CORE-1355) -- Reduce /v1/close calls from Android SDK.

## Description
1. Backend will be sending a header X-Branch-Send-Close-Request to support the v1/close decision.
2. By default, SDK will not get any header, so if it's not present, SDK should assume to not send v1/close requests.
3. If you get the header X-Branch-Send-Close-Request in the response, it will most likely be true only, but still, we should add a check for true just to keep it future proof. If the value is true, send the v1/close requests.

## Testing Instructions
One can whitelist the app in the Flipper here: http://flipper.prod.branch.io/?env=beta&namespace=gateway -> enable_send_v1_close_request.
@code-crusher @jdee or @echo-branch, whoever reviews this, would you help me test this? I could access flipper but haven't used it before and need a refresher on how to test using staging/beta environments.

LOW, but I haven't tested this with flipper myself, would appreciate help from the reviewer if that person has experience.

- Correctness & Style


Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
```diff
@@ -65,7 +66,8 @@ private BranchResponse doRestfulGet(String url, int retryNumber) throws BranchRe
             connection.setConnectTimeout(timeout);
             connection.setReadTimeout(timeout);
 
-            String requestId = getRequestIdFromHeader(connection);
+            String requestId = connection.getHeaderField(Defines.HeaderKey.RequestId.getKey());
+            Branch.getInstance().setCloseRequestNeeded(Boolean.parseBoolean(connection.getHeaderField(Defines.HeaderKey.SendCloseRequest.getKey())));
```
jdee, line 70: What happens if the header doesn't exist? Or if the field isn't parseable as a Boolean?

bklastaitis-branch: if header doesn't exist we get a null, if we try to parse the null, we get false, which is the default :)

jdee (approved):
:+1: LGTM

Issue #605, opened 2018-08-22
Branch not respected the timeout


bayrontq:
Hello!

The branch.initSession has not respected the timeout of 3000 ms and nor the timeout defined with branch.setNetworkTimeout.

Depending on the quality of the internet it takes longer than 15 seconds.

I use Banch SDK 2.19.0

StuStirling:
I am also seeing this issue running Branch SDK 2.18.0. It can run indefinitely when there is no internet.

sojanpr:
This is a known issue with URLConnection class bug in Java. If you check the URLConnection time out implementation, the time out is repeated per each DNS resolved IP addresses.

Branch is using URLConnection by default for networking.

Luckily Branch SDK provides support for your own network implementation in case you want to have more control or you want to use any other libraries like Volley OKHttp etc.

Here is how you add a custom network interface
https://github.com/BranchMetrics/android-branch-deep-linking/blob/master/Branch-SDK/src/io/branch/referral/Branch.java#L449

Here is the abstract interface
https://github.com/BranchMetrics/android-branch-deep-linking/blob/master/Branch-SDK/src/io/branch/referral/network/BranchRemoteInterface.java

StuStirling:
I have created a simple OkHttp replacement to be able to manage the timeouts. Seems to work ok.

```

class OkhttpBranchNetworkInterface : BranchRemoteInterface() {
    private val okHttpClient = OkHttpClient.Builder()
            .readTimeout(3, TimeUnit.SECONDS)
            .writeTimeout(3, TimeUnit.SECONDS)
            .connectTimeout(3, TimeUnit.SECONDS)
            .build()

    override fun doRestfulGet(url: String): BranchResponse {
        val request = Request.Builder()
                .url(url)
                .build()

        return handleNetworkRequest(request)
    }

    override fun doRestfulPost(url: String, payload: JSONObject): BranchResponse {
        val request = Request.Builder()
                .url(url)
                .post(RequestBody.create(MediaType.parse("application/json"),payload.toString()))
                .build()

        return handleNetworkRequest(request)
    }

    private fun handleNetworkRequest(request: Request) : BranchResponse {
        return try {
            val response = okHttpClient.newCall(request).execute()
            BranchResponse(response.body().toString(),response.code())
        } catch (exception: Exception) {
            if (exception is UnknownHostException) throw BranchRemoteException(BranchError.ERR_BRANCH_NO_CONNECTIVITY)
[...]

PR #882, opened 2020-12-08
Sdk 1090 fix timeout

bklastaitis-branch:
## Reference
https://branch.atlassian.net/browse/SDK-1090

## Description
Currently the SDK does not actually respect the value passed in the API, `setNetworkTimeout`, because of the `HttpsURLConnection.setConnectTimeout(...)` implementation that applies the passed in timeout to all IP addresses that an endpoint resolves to. We host `v1/open` endpoint on 10+ machines, so in a typical case with our default timeout of 5.5. seconds, we will not timeout for like a full minute.

Applied solution:
Artificially cap the allocated time given to the request to complete, and cancel the request otherwise. This is the solution I'm planning to implement for right now.

## Testing Instructions
Run the tester app, try to set the network timeout to various values and observe timeouts happen at the expected time. Here is a command to run emulator with super high latency to repro timeouts:
`emulator -avd NameOfYouEmulator -netspeed 1:1 -netdelay 500`

MEDIUM

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/Branch.java
@@ -1,5 +1,6 @@
 package io.branch.referral;
 
+import static io.branch.referral.BranchError.ERR_BRANCH_REQ_TIMED_OUT;
 import static io.branch.referral.BranchPreinstall.getPreinstallSystemData;
 import static io.branch.referral.BranchUtil.isTestModeEnabled;
 
@@ -590,24 +591,19 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
      */
     public static Branch getInstance(@NonNull Context context, @NonNull String branchKey) {
         if (branchReferral_ == null) {
-            branchReferral_ = Branch.initInstance(context);
-        }
-        if (branchReferral_.prefHelper_.isValidBranchKey(branchKey)) {
-            boolean isNewBranchKeySet = branchReferral_.prefHelper_.setBranchKey(branchKey);
-            //on setting a new key clear link cache and pending requests
-            if (isNewBranchKeySet) {
-                branchReferral_.linkCache_.clear();
-                branchReferral_.requestQueue_.clear();
+            branchReferral_ = new Branch(context.getApplicationContext());
+            if (branchReferral_.prefHelper_.isValidBranchKey(branchKey)) {
+                branchReferral_.prefHelper_.setBranchKey(branchKey);
+            } else {
+                PrefHelper.Debug("Branch Key is invalid. Please check your BranchKey");
             }
-        } else {
-            PrefHelper.Debug("Branch Key is invalid. Please check your BranchKey");
         }
         return branchReferral_;
     }
     
     private static Branch getBranchInstance(@NonNull Context context, boolean isLive, String branchKey) {
         if (branchReferral_ == null) {
-            branchReferral_ = Branch.initInstance(context);
+            branchReferral_ = new Branch(context.getApplicationContext());
 
             // Configure live or test mode
             boolean testModeAvailable = BranchUtil.checkTestMode(context);
@@ -766,16 +762,6 @@ public class Branch implements BranchViewHandler.IBranchViewEvents, SystemObserv
         branchReferral_.setIsReferrable(isReferrable);
         return branchReferral_;
     }
-    
-    /**
-     * <p>Initialises an instance of the Branch object.</p>
-     *
-     * @param context A {@link Context} from which this call was made.
-     * @return An initialised {@link Branch} object.
-     */
-    private static Branch initInstance(@NonNull Context context) {
-        return new Branch(context.getApplicationContext());
-    }
 
     // Package Private
     // For Unit Testing, we need to reset the Branch state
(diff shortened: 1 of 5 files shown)


Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -590,24 +591,19 @@ public static Branch getInstance() {
      */
     public static Branch getInstance(@NonNull Context context, @NonNull String branchKey) {
         if (branchReferral_ == null) {
-            branchReferral_ = Branch.initInstance(context);
-        }
-        if (branchReferral_.prefHelper_.isValidBranchKey(branchKey)) {
```
bklastaitis-branch, line 595: just some cleanup, the method was potentially resetting branch key at run time which the method was not meant to do.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -766,16 +762,6 @@ public static Branch getAutoTestInstance(@NonNull Context context, boolean isRef
...
         return branchReferral_;
     }
-    
-    /**
-     * <p>Initialises an instance of the Branch object.</p>
-     *
-     * @param context A {@link Context} from which this call was made.
-     * @return An initialised {@link Branch} object.
-     */
-    private static Branch initInstance(@NonNull Context context) {
```
bklastaitis-branch, line 776: also cleanup

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -2421,144 +2424,152 @@ protected ServerResponse doInBackground(Void... voids) {
         @Override
         protected void onPostExecute(ServerResponse serverResponse) {
             super.onPostExecute(serverResponse);
-            if (serverResponse != null) {
```
bklastaitis-branch, line 2424: The only change in this method is changing the `if` statement to reduce indentation layers, the change was from `if (serverResponse != null) {} ` to `if (serverResponse == null || isCancelled()) return`.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestRegisterOpen.java
```diff
@@ -103,12 +103,6 @@ public void onRequestSucceeded(ServerResponse resp, Branch branch) {
         onInitSessionCompleted(resp, branch);
     }
     
-    void setInitFinishedCallback(Branch.BranchReferralInitListener callback) {
```
bklastaitis-branch, line 106: cleanup, unused method

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
```diff
@@ -57,9 +56,6 @@ private BranchResponse doRestfulGet(String url, int retryNumber) throws BranchRe
         PrefHelper prefHelper = PrefHelper.getInstance(branch.getApplicationContext());
         try {
             int timeout = prefHelper.getTimeout();
-            if (timeout <= 0) {
```
bklastaitis-branch, line 60: `prefHelper.getTimeout` has a non-zero default value, so this would always fail/have no effect unless the client used `setNetworkTimeout` API and passed a negative value, which makes no sense.

Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java
```diff
@@ -81,17 +76,20 @@ private BranchResponse doRestfulGet(String url, int retryNumber) throws BranchRe
...
-                        return new BranchResponse(getResponseString(connection.getInputStream()), responseCode, Strings.emptyToNull(requestId));
+                        result = new BranchResponse(getResponseString(connection.getInputStream()), responseCode);
                     }
                 } catch (FileNotFoundException ex) {
                     // In case of Resource conflict getInputStream will throw FileNotFoundException. Handle it here in order to send the right status code
                     PrefHelper.Debug("A resource conflict occurred with this request " + url);
-                    return new BranchResponse(null, responseCode, Strings.emptyToNull(requestId));
+                    result = new BranchResponse(null, responseCode);
                 }
+                result.requestId = Strings.emptyToNull(requestId);
```
bklastaitis-branch, line 91: I removed `requestId` from `BranchResponse` because it's a public class that clients sometimes need to implement, e.g. https://github.com/BranchMetrics/android-branch-deep-linking-attribution/issues/605#issuecomment-431330448

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -2421,144 +2426,152 @@ protected ServerResponse doInBackground(Void... voids) {
...
+                                }
+
+                                if (updateRequestsInQueue) {
+                                    updateAllRequestsInQueue();
+                                }
+
+                                if (thisReq_ instanceof ServerRequestInitSession) {
+                                    setInitState(SESSION_STATE.INITIALISED);
+                                    thisReq_.onRequestSucceeded(serverResponse, branchReferral_);
+                                    if (!((ServerRequestInitSession) thisReq_).handleBranchViewIfAvailable((serverResponse))) {
```
jdee, line 2539: I realize you're just shifting/rearranging existing code, but I hadn't seen/noticed some of this before. Do these doubled parens around serverResponse do anything? Is this the same as:

```java
if (!((ServerRequestInitSession)thisReq_).handleBranchViewIfAvailable(serverResponse)) {
```

It would be more readable as

```java
if (thisReq_ instanceof ServeRequestInitSession) {
    ServerRequestInitSession initReq = (ServerRequestInitSession)thisReq_;
    // ...
    if (!initReq.handleBranchViewIfAvailable(serverResponse)) {
```

And for that matter when you have code like this checking `instanceof SomeClass`, it might be time to refactor. But that's obviously out of scope here.

jdee (approved):
:+1: Aside from clean-up on a number of fronts and adjusting the default timeout, the main part of the change seems to be the CountDownLatch used with the timeout, which is a good addition to avoid race conditions.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
bklastaitis-branch: There is so much to refactor in this SDK it makes me cringe. In fact I started and doing a big refactor and realized that I'll just make the PR impossible to follow.

Hopefully I can do the reactor before the next release :)

PR #887, opened 2021-02-04
Core 1645 deprecate old sdk initializers

bklastaitis-branch:
**Reviewing tips:
I tried to clean this as much as possible but couldn't get it perfect still.. This branch was started off of `origin/CORE-1588-hashed-aprtner-params`, which is the last branch to be merged to master. That commit is 6f8a5110, and it appears in the middle of the list of commits in this PR.. All commits following that one will have distinct changes with a reasonable commit message, so it will probably be easiest to review this PR one commit at a time starting from 6f8a5110.** 

## Description
We don't support `is_referrable` on the server side anymore
https://branch.slack.com/archives/CKTKKK40H/p1600730008057200

so we can deprecate sdk initializer overloads that use `isReferrable` i.e. `Branch.get(Auto)Instance(..., boolean isReferrable)`

While we are at it, let's deprecate undocumented initializers `Branch.getInstance(Context, ...)`, so we are left with just one initializer Branch.getAutoInstance(Context) and the overload that takes in a branch key. These initializers were meant get SDK singleton or initialize it, just like getAutoInstance, but they were mean to be called from outside of Application context and didn’t attach activity lifecycle listener. To that end, clients can still use getAutoInstance and the code will be smart enough to see when it’s called from Application or not and will attach the lifecycle listener accordingly. Not only were these not mentioned anywhere in our documentation but were also introducing bugs wrt getPreinstallSystemData, which was only added to getAutoInstance.

## Testing Instructions
N/A

LOW

- Correctness & Style


echo-branch (approved):
Looks good, works in my test app.

PR #888, opened 2021-02-04
Core 1667 fix android tests

bklastaitis-branch:
To be merged after 
https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/887

## Reference
CORE-1667 -- Fix Android Tests.
https://branch.atlassian.net/browse/CORE-1667

## Description
Last couple releases had broken Android testshttp://gitlab.prod.branch.io/branchmetricspublic/android-branch-deep-linking-attribution/commits/master

Upon taking a closer look I saw a few things:

we use outdated and deprecated junit libs, never libs allow us to control lifecycle events easier

many android tests are simply broken

BranchEventTest was misused as a base class because it had it's own `@Test` methods. What this means is that for every `@Test` method  inside child class of `BranchEventTest`, first `setup` is called, then all parent class' `@Test` methods are run, then the actual child class' `@Test` method ir run, and finally parent class' `tearDown` is run - this unnecessarily prolonged test execution and potentially introduced test co-dependencies leading to test instability.

network request tests, BranchSDKTests, lived in testbed app because they would actually execute the requests instead of simply mocking them. Mocking the execution of the request would speed up execution of tests besides not costing us money every time the test suite is run.

## Testing Instructions
Run the test suite `./gradlew Branch-SDK:connectedAndroidTest`

HIGH (because of SDK code changes that were needed to address the failing tests)

- Correctness & Style


Branch-SDK/src/androidTest/java/io/branch/referral/BranchSDKTests.java
```diff
@@ -234,32 +197,32 @@ public void onStateChanged(boolean changed, BranchError error) {
...
     @Test
     public void testZLoad() throws InterruptedException {
-        initSession();
-
+        initSessionResumeActivity();
         final CountDownLatch signalFinal = new CountDownLatch(1);
         final CountDownLatch signal = new CountDownLatch(1);
-        for (int i = 0; i < 100; i++) {
-            final int idx = i;
+        final int reps = 20;
```
bklastaitis-branch, line 208: this used to be a much larger number, like 100, which just bogged up resources as each request
needs it's own AsyncTask, and the test never actually tested that all 100 requests succeeded because `signal.await(10, TimeUnit.SECONDS);` did not have `Assert.assertTrue` in front of it

echo-branch (approved):

Branch-SDK/src/main/java/io/branch/referral/BranchAsyncTask.java
```diff
@@ -16,14 +16,11 @@
      * @param params Params for executing this Async task
      * @return This object for method chaining
      */
-    public AsyncTask<Params, Progress, Result> executeTask(Params... params) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
```
bklastaitis-branch, line 20: min android api version set on the sdk ensures that this is statement is always true

Branch-SDK/src/main/java/io/branch/referral/BranchPostTask.java
```diff
@@ -0,0 +1,181 @@
...
+
+import static io.branch.referral.BranchError.ERR_BRANCH_REQ_TIMED_OUT;
+
+/**
+ * Asynchronous task handling execution of server requests. Execute the network task on background
+ * thread and request are  executed in sequential manner. Handles the request execution in
+ * Synchronous-Asynchronous pattern. Should be invoked only form main thread and  the results are
+ * published in the main thread.
+ */
+public class BranchPostTask extends BranchAsyncTask<Void, Void, ServerResponse> {
```
bklastaitis-branch, line 20: this is a massive class that used to live inside the Branch.java singleton. I was able to pass the SDK singleton as constructor argument and make the relevant methods & parameters package private, so moving this class out to its own file should not have an effect.

However, the `onPostExecute` method changed a little bit, it would be best to look at the original and compare the two. In a nutshell, I could not understand why the requests (once completed) were being removed from the request queue **by index** and not **by reference** to the completed request, which is just so much safer. In addition, we used to remove ALL requests from the queue once we encounter the first failed request whose status code is not 400/409. I disagree with the assumption that all subsequent requests will fail. With the flaky connections on mobile phones being so common, the next request could succeed just fine... To that end, I now only remove the recently completed request from the queue and leave all other requests on the queue.

Branch-SDK/src/main/java/io/branch/referral/BranchShortLinkBuilder.java
```diff
@@ -137,10 +137,6 @@ public String getShortUrl() {
      * @param callback A {@link Branch.BranchLinkCreateListener} callback instance that will trigger
      */
     public void generateShortUrl(Branch.BranchLinkCreateListener callback) {
-        super.generateUrl(callback);
-    }
-
-    void generateShortUrlInternal(Branch.BranchLinkCreateListener callback, boolean isFromShareSheet) {
```
bklastaitis-branch, line 143: I'm removing only internal methods that overload url generation with the argument `isFromShareSheet`, which apparently isn't used in any shape or form.

Branch-SDK/src/main/java/io/branch/indexing/BranchUniversalObject.java
```diff
@@ -620,7 +622,7 @@ public String getShortUrl(@NonNull Context context, @NonNull LinkProperties link
      * @param callback       An instance of {@link io.branch.referral.Branch.BranchLinkCreateListener} to receive the results
      */
     public void generateShortUrl(@NonNull Context context, @NonNull LinkProperties linkProperties, @Nullable Branch.BranchLinkCreateListener callback) {
-        if (Branch.getInstance().isTrackingDisabled()) {
```
bklastaitis-branch, line 623: potential NPE

Branch-SDK/src/main/java/io/branch/referral/BranchLinkData.java
```diff
@@ -373,7 +368,14 @@ public boolean equals(Object obj) {
         if (tags == null) {
             if (other.tags != null)
                 return false;
-        } else if (!tags.toString().equals(other.tags.toString()))
+        } else if (!tags.toString().equals(other.tags != null ? other.tags.toString() : null))
```
bklastaitis-branch, line 371: adds null safety

Branch-SDK/src/main/java/io/branch/referral/BranchLinkData.java
```diff
@@ -373,7 +368,14 @@ public boolean equals(Object obj) {
...
+            return false;
+
+        // todo JSONObject's order of keys is not consistent, so this equality can lead to false positives
+        if (params == null) {
+            if (other.params != null)
+                return false;
+        } else if (!params.toString().equals(other.params != null ? other.params.toString() : null))
```
bklastaitis-branch, line 378: This equality never worked before (see the comment why), now it will work most of the time but still not always. It's the best we can do though without a serialization library or a custom, recursive JSONObject equals method which would be easy to mess up and I doubt it would add much value given that this never worked and no one noticed.

Branch-SDK/src/main/java/io/branch/referral/ServerRequest.java
```diff
@@ -645,11 +645,6 @@ public BRANCH_API_VERSION getBranchRemoteAPIVersion() {
         return BRANCH_API_VERSION.V1;  // Default is v1
     }
     
-    public void reportTrackingDisabledError() {
```
bklastaitis-branch, line 648: public method for internal use... I just copy pasted the implementation of the method to places where the method was invoked.

Branch-SDK/src/main/java/io/branch/referral/ServerRequest.java
```diff
@@ -668,17 +663,4 @@ boolean isInitializationOrEventRequest() {
         }
         return false;
     }
-
-    protected static void addPartnerParams(JSONObject body, BranchPartnerParameters partnerParams) throws JSONException {
```
bklastaitis-branch, line 672: not used anywhere, should have been cleaned up after implementing hashed Facebook param integration.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestCreateUrl.java
```diff
@@ -22,7 +22,6 @@
     private Branch.BranchLinkCreateListener callback_;
     /* Default long link base url*/
     private static final String DEF_BASE_URL = "https://bnc.lt/a/";
-    private boolean isReqStartedFromBranchShareSheet_;
```
bklastaitis-branch, line 25: As mentioned in the previous comment, this is not used anywhere, plus we have a few overloaded methods that `set` this value but, again, we never `get` it, so don't need it and we don't need the overloaded methods.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestGetCPID.java
```diff
@@ -51,9 +51,7 @@ public void handleFailure(int statusCode, String causeMsg) {
             return;
         }
 
-        callback.onDataFetched(null,
```
bklastaitis-branch, line 54: I don't why we had this implemented the way it was in the first place

Branch-SDK/src/main/java/io/branch/referral/ServerRequestGetLATD.java
```diff
@@ -55,8 +55,7 @@ public void onRequestSucceeded(ServerResponse response, Branch branch) {
     @Override
     public void handleFailure(int statusCode, String causeMsg) {
         if (callback != null) {
-            callback.onDataFetched(null, new BranchError("Failed to get last attributed touch data",
```
bklastaitis-branch, line 58: I don't why we had this implemented the way it was in the first place

Branch-SDK/src/main/java/io/branch/referral/ServerRequestRegisterInstall.java
```diff
@@ -102,13 +102,6 @@ public void onRequestSucceeded(ServerResponse resp, Branch branch) {
         onInitSessionCompleted(resp, branch);
     }
     
-    
-    void setInitFinishedCallback(Branch.BranchReferralInitListener callback) {
```
bklastaitis-branch, line 106: Not used anywhere anymore (forgot to cleanup the last time)

Branch-SDK/src/main/java/io/branch/referral/ServerRequestQueue.java
```diff
@@ -71,29 +71,24 @@ private ServerRequestQueue(Context c) {
...
+                    if (aQueue.isPersistable()) {
+                        JSONObject json = aQueue.toJSON();
+                        if (json != null) {
+                            jsonArr.put(json);
                         }
                     }
                 }
-
-                try {
-                    editor.putString(PREF_KEY, jsonArr.toString()).commit();
```
bklastaitis-branch, line 90: This changes from writing directly to disk to changing the shared prefs singleton + setting up an async operation to write the new value to disk. Previous implementation was fine because it was running in a separate thread, but to improve performance, I removed the creation of the thread. Not only will this improve performance, I think might also prevent concurrency bugs because thread creation takes time. To avoid any unexpected bugs that expanded the scope of the existing try catch block to include the entire method implementation.

Branch-SDK/src/main/java/io/branch/referral/BranchContentUrlBuilder.java
```diff
@@ -31,6 +31,6 @@ public String getContentUrl() {
      * @param callback A {@link Branch.BranchLinkCreateListener} callback instance that will trigger
      */
     public void generateContentUrl(Branch.BranchLinkCreateListener callback) {
-        super.generateUrl(callback);
```
bklastaitis-branch, line 34: part of https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/888/files#r577363727

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -1641,10 +1638,6 @@ private String generateShortLinkSync(ServerRequestCreateUrl req) {
         return null;
     }
     
-    private void generateShortLinkAsync(final ServerRequest req) {
```
bklastaitis-branch, line 1644: bloat

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -1734,23 +1726,7 @@ private boolean isSessionAvailableForRequest() {
...
-        ServerRequest req;
-        if (index >= requestQueue_.getSize()) {
-            req = requestQueue_.peekAt(requestQueue_.getSize() - 1);
-        } else {
-            req = requestQueue_.peekAt(index);
-        }
-        handleFailure(req, statusCode);
-    }
-    
-    private void handleFailure(final ServerRequest req, int statusCode) {
```
bklastaitis-branch, line 1747: bloat

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -1734,23 +1726,7 @@ private boolean isSessionAvailableForRequest() {
         return (hasSession() && hasDeviceFingerPrint());
     }
     
-    private void handleFailure(int index, int statusCode) {
```
bklastaitis-branch, line 1737: bloat + see comment in BranchPostTask wrt unnecessary use of indexes (i.e. the places where this is called all have a reference to the request that needs to handle the error)

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -2257,200 +2234,12 @@ private boolean checkIntentForUnusedBranchLink(Intent intent) {
...
-                                // cache the link
-                                linkCache_.put(((ServerRequestCreateUrl) thisReq_).getLinkPost(), url);
-                            }
-                        }
-                        //On Logout clear the link cache and all pending requests
-                        else if (thisReq_ instanceof ServerRequestLogout) {
-                            linkCache_.clear();
-                            requestQueue_.clear();
-                        }
-                        requestQueue_.dequeue();
```
bklastaitis-branch, line 2380: I could not understand why the successful requests were being removed from the request queue **by index** and not **by reference** to the successfully completed request, which is just so much safer.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -2257,200 +2234,12 @@ private boolean checkIntentForUnusedBranchLink(Intent intent) {
...
-                            hasNetwork_ = false;
-                            //Collect all request from the queue which need to be failed.
-                            ArrayList<ServerRequest> requestToFail = new ArrayList<>();
-                            for (int i = 0; i < requestQueue_.getSize(); i++) {
-                                requestToFail.add(requestQueue_.peekAt(i));
-                            }
-                            //Remove the requests from the request queue first
-                            for (ServerRequest req : requestToFail) {
-                                if (req == null || !req.shouldRetryOnFail()) { // Should remove any nullified request object also from queue
-                                    requestQueue_.remove(req);
```
bklastaitis-branch, line 2347: we used to remove ALL (except `v1/event` and `v1/profile`) requests from the queue once we encounter the first failed request whose status code is not 400/409. I disagree with the assumption that all subsequent requests will fail. With the flaky connections on mobile phones being so common, the next request could succeed just fine... To that end, I now only remove the recently completed request from the queue (unless it failed and it's a `v1/event` or `v1/profile` event) but I leave all other requests on the queue.

Branch-SDK/src/main/java/io/branch/referral/BranchPostTask.java
```diff
@@ -0,0 +1,194 @@
...
+            //for request which need to be replayed on failure.
+            branch.networkCount_ = 0;
+            thisReq_.handleFailure(status, serverResponse.getFailReason());
+        }
+
+        if (!thisReq_.shouldRetryOnFail()) {
+            branch.requestQueue_.remove(thisReq_);
+        } else {
+            // failure has already been handled
+            // todo does it make sense to retry the request without a callback? (e.g. CPID, LATD)
```
bklastaitis-branch, line 184: @echo-branch the current implementation of CPID and LATD requests was such that they will retry on failure indefinitely but their clearCallbacks method was a no op, so potentially the user would just keep getting errors in the callback indefinitely. I fixed the `clearCallbacks` methods of those requests so they do what the method name says but wanted to double check with you and see what the iOS does.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestQueue.java
```diff
@@ -148,26 +143,8 @@ void enqueue(ServerRequest request) {
     }
     
     /**
-     * <p>Removes the queued {@link ServerRequest} object at position with index 0 within the queue,
-     * and returns it as a result.</p>
-     *
-     * @return The {@link ServerRequest} object at position with index 0 within the queue.
-     */
-    ServerRequest dequeue() {
```
bklastaitis-branch, line 156: This is part of the change made to BranchPostTask, onPostExecute, I now use queue.remove(request).

Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
```diff
@@ -11,7 +13,7 @@
  * </p>
  */
 
-class TrackingController {
+public class TrackingController {
```
bklastaitis-branch, line 16: I don't think making this public makes any difference because only `isTrackingDisabled` method is public, which is public in `Branch.java` as well. The method `isTrackingDisabled` was needed to avoid a potential NPE.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -293,20 +292,18 @@
...
-    
+
     private final Semaphore serverSema_ = new Semaphore(1);
-    
+
     final ServerRequestQueue requestQueue_;
     
-    private int networkCount_ = 0;
-    
-    private boolean hasNetwork_ = true;
```
bklastaitis-branch, line 308: Not needed anymore after making change to BranchPostTask (this flag was used to block subsequent requests when previous request failed)

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -293,20 +292,18 @@
      * the class during application runtime.</p>
      */
     private static Branch branchReferral_;
-    
+
     private BranchRemoteInterface branchRemoteInterface_;
-    private final PrefHelper prefHelper_;
+    final PrefHelper prefHelper_;
```
bklastaitis-branch, line 297: all this moving of methods and parameters in this class from `private` to package private is meant to accommodate moving out the BranchPostTask to its own file.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -876,19 +873,11 @@ public static void enableCookieBasedMatching(String cookieMatchDomain, int delay
      */
     private void executeClose() {
         if (initState_ != SESSION_STATE.UNINITIALISED) {
-            if (!hasNetwork_) {
```
bklastaitis-branch, line 879: The `if` block can only happen when init session request is in flight, in which case the removing it from the queue has no effect, the request will complete either way + it will result in a bug because the `else` block won't be executed which would clean up the last session params (the way this owuld manifest is that during the next session, if user queries last session params before actually initializing the session, they will get valid values instead of "BNC_NO_VALUE".)

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -1672,22 +1656,20 @@ private void processNextQueueItem() {
                 
                 serverSema_.release();
                 if (req != null) {
+                    PrefHelper.Debug("processNextQueueItem, req " + req.getClass().getSimpleName());
                     if (!req.isWaitingOnProcessToFinish()) {
                         // All request except Install request need a valid IdentityID
                         if (!(req instanceof ServerRequestRegisterInstall) && !hasUser()) {
                             PrefHelper.Debug("Branch Error: User session has not been initialized!");
                             networkCount_ = 0;
-                            handleFailure(requestQueue_.getSize() - 1, BranchError.ERR_NO_SESSION);
```
bklastaitis-branch, line 1680: Unnecessary setup for a race condition bug

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -2263,194 +2231,6 @@ private boolean checkIntentForUnusedBranchLink(Intent intent) {
...
-                                }
-                            }
-                        } else {
-                            //Publish success to listeners
-                            thisReq_.onRequestSucceeded(serverResponse, branchReferral_);
-                        }
-                    }
-                }
-                networkCount_ = 0;
-                if (hasNetwork_ && initState_ != SESSION_STATE.UNINITIALISED) {
```
bklastaitis-branch, line 2439: I removed both conditions here because I don't either one makes sense, e.g. if SESSION_STATE.UNINITIALISED, then we should execute the request and return an error indicating that SESSION_STATE.UNINITIALISED.

wrt `hasNetwork_` I left a few comments on that already, the reasoning is the same though, shouldn't execute requests right away and let the client know why they are failing rather than holding back on executing those requests.

Branch-SDK/src/main/java/io/branch/referral/BranchLinkData.java
echo-branch: If that check didn't work before, could we just remove it?

Branch-SDK/src/main/java/io/branch/referral/BranchPostTask.java
echo-branch: iOS doesn't retry them.

jdee (approved):
:100: This is nice. I've seen that misused base class pattern elsewhere with tests embedded in the base class. At a minimum, the same tests are run repeatedly for each subclass, and testing takes much longer.

Branch-SDK/src/main/java/io/branch/referral/BranchLinkData.java
bklastaitis-branch: So `BranchLinkData.java` is used as a key to cache urls that the client creates, if we removed the equality of `params`, then url link builders with different `params` would yield the same url each time which is worse I think.

bklastaitis-branch: We use "stringified" json as keys in a map to track impression on the disco SDK, it's most likely going to be changed eventually because it isn't stable but so far it seems to work reasonably well, so I figured it's a good enough improvement *for now*, I do agree that eventually it should be fixed properly.

Branch-SDK/src/main/java/io/branch/referral/BranchPostTask.java
bklastaitis-branch: Perfect!

PR #932, opened 2021-09-20
INTENG-13463 -- Android app crashing when running in the background

agerstner-branch:
## Reference
INTENG-13463 -- Android app crashing when running in the background

## Description
Client is reporting that their android app is crashing when it is in the background and they are using mParticle Branch kit. After investigations, this is due to a change in the Branch SDK where BranchPostTask was moved out of the Branch singleton class into a stand alone class. To accomplish the same functionality, a reference of the Branch object was passed in and in some rare situations can be null while being garbage collected, causing NPE on Amazon Fire OS devices. This change reverts the task back to being an inner class, removing references and accessing the Branch object directly.

Related Jira Tickets:
INTENG-13463
INTENG-13402
CORE-2388
CORE-2385
CORE-2030

## Testing Instructions
We have not been able to reproduce the issue internally, but we worked directly with the customer, supplying test AAR packages and having them verify the change solved the issue in their application.

## Risk Assessment LOW
Since this change is invisible to the end user, the risk is low.

1 unrelated test fails before the change. Verified that all remaining tests passed before and after the change.
Verified locally that the class was being called.

- [x] JIRA Ticket is referenced in PR title.
- Correctness & Style
    - [x] Conforms to [AOSP Style Guides](https://source.android.com/setup/contribute/code-style)
    - [x] Mission critical pieces are documented in code and out of code as needed.
- [x] Unit Tests reviewed and test issue sufficiently.


bklastaitis-branch (approved):
LGTM

PR #938, opened 2021-10-01
Release v5.0.13

echo-branch:
- v5.0.13
  * _*Master Release*_ - Oct 1, 2021
  * Always include Google Play Store referrer in install
  * IntegrationValidator now recoginizes String resources
  * Fix edge case that could cause a stack overflow


agerstner-branch (approved):
LGTM

PR #964, opened 2022-02-04
Resolve Timeout Issues

gdeluna-branch:
## Reference
INTENG-14315 -- Resolve timeout issues

## Description

The problem appears to be the first network request (open/install) is prematurely cancelled and not retried. The exception that is being thrown is InterruptedIOException which is not handled but falls through to the IOException catch block, and is not retried. They are not placed back on the request queue because they do not override shouldRetryOnFail to true. Thus the sdk is in a state of no recovery.

I verified that

Setting the[ connect timeout](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/f8966bc81f910171ccc4e29798e150066ef5a07a/Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java#L138) to 100ms throws SocketTimeoutException
	- Already handled with retries in the caught exception block

Setting the [read timeout ](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/f8966bc81f910171ccc4e29798e150066ef5a07a/Branch-SDK/src/main/java/io/branch/referral/network/BranchRemoteInterfaceUrlConnection.java#L139)to 100ms throws SocketTimeoutException
	- Also handled as above

Setting the countdown latch timeout in [executeTimedBranchPostTask(req, prefHelper_.getTimeout());](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/de0667f6988ba720040e1ed3125418d2b9352d97/Branch-SDK/src/main/java/io/branch/referral/Branch.java#L1669) to 100ms throws InterruptedIOException

Possible root cause:
	The countdown latch is timed against the entire task of constructing and interfacing with the HttpsURLConnection object and other calls. It seems likely that an older device with less resources would take more time executing before getting to the network IO, which also may take longer.

Proposed changes:

- 	Split out ConnectTimeout, ReadTimeout, TaskTimeout (for countdown latch)
         - These values are each used for different purposes and should be configurable to the customers’ needs
         - For back compat, I could leave ReadTimeout as “Timeout” but that might be confusing
	

- Catch InterruptedIOException explicitly and retry
         - If there were a thread timeout, or a failure not due to network, a subsequent retry proved to be much more likely to succeed because the [underlying connection may be reused ](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html)and the task is overall faster to complete.
         
- Override shouldRetryOnFail for ServerRequestRegisterInstall and ServerRequestRegisterOpen to be true
        - Unless other requests imply an initialization request, this seems necessary for recovery

Follow up
	- When shouldRetryOnFail == true; it will forever be placed back on the queue if it did not fail with an unretryableErrorCode. Should there be a limit or an extended retry interval?
	- Should the request’s onPostExecuteInner be calling latch_.countDown() when doInBackground already does so before returning?

Added unit tests, modified existing tests to catch task timeout.

After feedback, task timeout is now computed from the connect and read timeouts

## Testing Instructions
Manual testing with test app using locally built package referenced in old device(KFMUWI) and emulators.

## Risk Assessment [`LOW`]

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
@@ -54,7 +54,8 @@ public class PrefHelper {
      */
     private static final int MAX_RETRIES = 3; // Default retry count is 3
 
-    static final int TIMEOUT = 5500; // Default timeout id 5.5 sec
+    static final int TIMEOUT = 5500; // Default timeout is 5.5 sec
+    static final int CONNECT_TIMEOUT = 1000; // Default timeout is 10 seconds
 
     private static final String SHARED_PREF_FILE = "branch_referral_shared_pref";
     
@@ -87,7 +88,8 @@ public class PrefHelper {
     private static final String KEY_RETRY_COUNT = "bnc_retry_count";
     private static final String KEY_RETRY_INTERVAL = "bnc_retry_interval";
     private static final String KEY_TIMEOUT = "bnc_timeout";
-    
+    private static final String KEY_CONNECT_TIMEOUT = "bnc_connect_timeout";
+
     private static final String KEY_LAST_READ_SYSTEM = "bnc_system_read_date";
     
     private static final String KEY_EXTERNAL_INTENT_URI = "bnc_external_intent_uri";
@@ -265,6 +267,27 @@ public class PrefHelper {
      */
     public int getTimeout() {
         return getInteger(KEY_TIMEOUT, TIMEOUT);
+    }
+
+    /**
+     * <p>Sets the duration in milliseconds to override the timeout value for initiating requests.</p>
+     *
+     * @param connectTimeout The {@link Integer} value of the connect timeout setting in milliseconds.
+     */
+    public void setConnectTimeout(int connectTimeout) {
+        setInteger(KEY_CONNECT_TIMEOUT, connectTimeout);
+    }
+
+
+    /**
+     * <p>Returns the currently set timeout value for opening a communication channel with a remote
+     * resource. This may take longer on older devices with lower memory and threading capabilities.</p>
+     *
+     * @return An {@link Integer} value containing the currently set timeout value in
+     * milliseconds.
+     */
+    public int getConnectTimeout() {
+        return getInteger(KEY_CONNECT_TIMEOUT, CONNECT_TIMEOUT);
     }
     
     /**
--- a/Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
@@ -272,24 +272,13 @@ public class PrefHelper {
     }
 
     /**
-     * <p>Sets the duration in milliseconds to override the timeout value for tasks</p>
-     *
-     * @param taskTimeout The {@link Integer} value of the timeout setting in milliseconds.
-     */
-    public void setTaskTimeout(int taskTimeout) {
-        setInteger(KEY_TASK_TIMEOUT, taskTimeout);
-    }
-
-    /**
-     * <p>Returns the currently set timeout value for tasks. This will be the default
-     * SDK setting unless it has been overridden manually between Branch object instantiation and
-     * this call.</p>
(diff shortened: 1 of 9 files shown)


Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -54,7 +54,9 @@
      */
     private static final int MAX_RETRIES = 3; // Default retry count is 3
 
-    static final int TIMEOUT = 5500; // Default timeout id 5.5 sec
+    static final int TIMEOUT = 5500; // Default timeout is 5.5 sec
+    static final int CONNECT_TIMEOUT = 10000; // Default timeout is 10 seconds
+    static final int TASK_TIMEOUT = TIMEOUT+CONNECT_TIMEOUT; // Default timeout is 15.5 seconds
```
agerstner-branch, line 59: Since this is computed as a constant, I would do the same for the setters.

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -54,7 +54,9 @@
      */
     private static final int MAX_RETRIES = 3; // Default retry count is 3
 
-    static final int TIMEOUT = 5500; // Default timeout id 5.5 sec
+    static final int TIMEOUT = 5500; // Default timeout is 5.5 sec
+    static final int CONNECT_TIMEOUT = 10000; // Default timeout is 10 seconds
```
agerstner-branch, line 58: 10sec seems excessive as a default? However, I have no basis for this statement. 🤣

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -266,6 +270,48 @@ public void setTimeout(int timeout) {
     public int getTimeout() {
         return getInteger(KEY_TIMEOUT, TIMEOUT);
     }
+
+    /**
+     * <p>Sets the duration in milliseconds to override the timeout value for tasks</p>
+     *
+     * @param taskTimeout The {@link Integer} value of the timeout setting in milliseconds.
+     */
+    public void setTaskTimeout(int taskTimeout) {
```
agerstner-branch, line 279: I would drop this setter and have the getter return a computed value.

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -266,6 +270,48 @@ public void setTimeout(int timeout) {
...
+    /**
+     * <p>Returns the currently set timeout value for tasks. This will be the default
+     * SDK setting unless it has been overridden manually between Branch object instantiation and
+     * this call.</p>
+     *
+     * @return An {@link Integer} value containing the currently set timeout value in
+     * milliseconds.
+     */
+    public int getTaskTimeout() {
+        return getInteger(KEY_TASK_TIMEOUT, TASK_TIMEOUT);
```
agerstner-branch, line 292: return computed value instead.

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
gdeluna-branch: Done

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
gdeluna-branch: Done

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
gdeluna-branch: Discussed internally, these values are approx equal to the ios implementation

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
gdeluna-branch: done

agerstner-branch (approved):
LGTM

echo-branch (approved):

PR #1087, opened 2023-07-26
[SDK-2034] Remove unnecessary fields from v1/url requests

nsingh-branch:
## Reference
SDK-2034 -- Remove unnecessary fields from v1/url requests

## Summary
Added a section of code to `ServerRequestCreateUrl()` in `ServerRequestCrealUrl.java` to remove 6 fields from `v1/url` requests which are no longer necessary.

## Motivation

The motivation for this change is to reduce the number of fields in v1/url requests down to only what's required or used. 

## Type Of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing Instructions

Create short url's and observe the outgoing requests. They should not contain any of the 6 removed fields. 
```
RandomizedBundleToken,
RandomizedDeviceToken,
SessionID,
AnonID,
IsHardwareIDReal,
HardwareID
```


gdeluna-branch:
The summary states 6 removed fields, the testing instructions say 10, and lists 6. So it those 6 right?

Branch-SDK/src/main/java/io/branch/referral/ServerRequestCreateUrl.java
```diff
@@ -88,11 +85,15 @@ public ServerRequestCreateUrl(Context context, final String alias, final int typ
 
             setPost(linkPost_);
 
+            //Remove unnecessary fields
+            linkPost_.remove("anon_id");
```
gdeluna-branch, line 89: Where are these added from originally? I think i'd rather have them omitted to begin with than exist on the object indeterminately. 

I don't want to create feature scope but it's a good idea to be additive than subtractive wrt only requiring we need to begin with.

gdeluna-branch:
also pull master and check test results.

nsingh-branch:
> The summary states 6 removed fields, the testing instructions say 10, and lists 6. So it those 6 right?

Ah yeah 6 is the correct number. I copied that over from iOS where there were 10 fields.

Branch-SDK/src/main/java/io/branch/referral/ServerRequestCreateUrl.java
nsingh-branch: They're added in the setPost method above, which calls DeviceInfo's updateRequestWithV1Params. Since that function just adds all of the fields to each request, the alternative would be to add some checks in that function for specific request types.

gdeluna-branch: Yeah we'll want to refactor that so that each type is explicit about its requirements. (in another pr)

gdeluna-branch (approved):

PR #1094, opened 2023-07-31
Convert gaid, oaid, faid, to coroutine

gdeluna-branch:
## Reference
SDK-2018 -- [Android] Convert GAID AsyncTask to coroutine

## Description
Reintroducing the change here https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/1084
>Converts the original GAID AsyncTask to a coroutine. Originally the GAID fetch task was managed with a countdown latch and a manually spun thread. Because the ordering of sub tasks in the queue is managed by manual locks, the coroutine must still post to the callback passed in to signal the original wait lock to be lifted.

>The coroutine itself is a suspend function calls into the AdvertisingIdClient by compiled code, and returns the info object containing the lat value and id if allowed. We still maintain the old logic to read/write to the DeviceInfo/SystemObserver class but may soon be refactored.

>Note: The resumeWith function executes on a DefaultDispatcher-worker thread. Soon this pool will grow with other coroutines being converted from AsyncTasks.

>Additionally, removing the call in the Application.OnCreate as that breaks the separation of Application and Activity lifecycles. Because [we have to check for the GAID on every re-entry into the app regardless](https://developer.android.com/training/articles/ad-id), it is an unnecessary and buggy optimization with little value.

>[More reading](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html)

Additional changes include
- Converting the Huawei OAID and Amazon Fire Ad Id to coroutines
- Removal of old unit tests that are obsoleted with threading changes and call ordering.
- Minor fixes for other tests and adding Huawei mobile services dependency for tests.

## Testing Instructions
Same as before, ensure the advertising id is populated when allowed on your device/emulator, and not present when disallowed.

## Risk Assessment [`MEDIUM`]

- [✅] I, the PR creator, have tested — integration, unit, or otherwise — this code.

- Correctness & Style


nsingh-branch (approved):
I was able to integrate this version into a test app and see the advertising_id.

nsingh-branch (approved):

PR #1180, opened 2024-03-21
[SDK-2325] Added callback to disableTracking()

nsingh-branch:
## Reference
SDK-2325 -- Add callback to disableTracking()
## Description
This change adds a new callback to the `disabledTracking()` method that returns the new tracking state, the referring params, and an error. If tracking is disabled, only the tracking state will be returned. If tracking is enabled, it will also return the referringParams and any potential errors from the Branch session initialization.

This change is useful for delayed initialization situations where tracking is initially disabled and the re-enabling of tracking is the first Branch initialization call. 

## Testing Instructions

Call the `disableTracking()` method without and with the callback and observe that all of the returned variables are correct.

```
//Example code with callback
Branch.getInstance().disableTracking(true, new Branch.TrackingStateCallback() {
    @Override
    public void onTrackingStateChanged(boolean trackingDisabled, JSONObject referringParams, BranchError error) {
        if (trackingDisabled) {
            Log.d("BranchSDK", "Tracking is now disabled.");
        } else {
            Log.d("BranchSDK", "Tracking is now enabled. Referring params: " + referringParams);
        }
    }
});
```

## Risk Assessment [`LOW`]

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
@@ -25,6 +25,7 @@ public class TrackingController {
     }
 
     void disableTracking(Context context, boolean disableTracking, @Nullable Branch.TrackingStateCallback callback) {
+        // If the tracking state is already set to the desired state, then return instantly
         if (trackingDisabled == disableTracking) {
             if (callback != null) {
                 callback.onTrackingStateChanged(trackingDisabled, Branch.getInstance().getFirstReferringParams(), null);
--- a/Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
@@ -3,6 +3,7 @@ package io.branch.referral;
 import android.content.Context;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import org.json.JSONObject;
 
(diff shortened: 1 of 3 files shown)


Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
```diff
@@ -22,19 +23,32 @@ public class TrackingController {
...
-    void disableTracking(Context context, boolean disableTracking) {
-        if (trackingDisabled != disableTracking) {
-            trackingDisabled = disableTracking;
-            if (disableTracking) {
-                onTrackingDisabled(context);
-            } else {
-                onTrackingEnabled();
+
+    void disableTracking(Context context, boolean disableTracking, @Nullable Branch.TrackingStateCallback callback) {
+        if (trackingDisabled == disableTracking) {
```
gdeluna-branch, line 28: Can you add a comment why this check is necessary?

Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
```diff
@@ -22,19 +23,32 @@ public class TrackingController {
...
+            }
+            return;
+        }
+
+        trackingDisabled = disableTracking;
+        PrefHelper.getInstance(context).setBool(PrefHelper.KEY_TRACKING_STATE, disableTracking);
+
+        if (disableTracking) {
+            onTrackingDisabled(context);
+            if (callback != null) {
```
gdeluna-branch, line 40: Does tracking disabled also return a callback? If not I'd suggest adding, so that there's mirrored control for both enable and disable tracking.

Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
nsingh-branch: Ah yeah that naming is confusing. Just added this comment above to clarify.        
`// If the tracking state is already set to the desired state, then return instantly`

Branch-SDK/src/main/java/io/branch/referral/TrackingController.java
nsingh-branch: `onTrackingDisabled` does not have a callback since it just clears PrefHelper values. And `onTrackingEnabled` actually uses the BranchReferralInitListener callback, which it passes a few functions down to `ServerRequestRegisterOpen`. 

Does it still make sense to have a callback for `onTrackingDisabled`?

gdeluna-branch: My thinking is if tracking is explicitly disabled, the user should know when that completes. But if values are only getting set, then it should be all synchronous.

nsingh-branch: Yeah thats what i'm thinking. It does all appear synchronous so this should technically return the TrackingStateCallback with the state when its done.

gdeluna-branch (approved):

PR #1199, opened 2024-06-03
Reinstallation workflow Fix

NidhiDixit09:
## Reference
QSS-112 - Reinstallation Scenario
https://branch.atlassian.net/browse/QSS-112

## Description
Re-installation test was failing, when ran using Github action, because of extra quotes in envirnoment variables and missing check for signing.

## Testing Instructions
Run appium tests and validate that Re-installation test runs. However, with current branch test will fail, because SDK and Automation Test Bed binaries are exactly same to the one in master branch. So browserstack will not load binary from this branch. Its a duplicate binary. In order to pass test, make some changes in SDK and then run. 

Example : I had added a variable and initalized it. So that binary in this branch is not duplicate to the one in master branch.

```
private final int test;

DeviceInfo(Context context) {
        context_ = context;
        systemObserver_ = new SystemObserverInstance();
        test = 123;
    }

```

LOW

- [ X] I, the PR creator, have tested — integration, unit, or otherwise — this code.

### DIFF
--- a/.github/workflows/appium-automation.yml
+++ b/.github/workflows/appium-automation.yml
@@ -43,20 +43,21 @@ jobs:
           java-version: ${{ env.JAVA_VERSION }}
       # to reduce complexity, testbed app uses sdk as a project reference
       - name: Decode Keystore
-        if: ${{ !( github.ref == 'refs/heads/Reinstallationtest-workflow-fix-')}}
         env:
             ENCODED_STRING: ${{ secrets.KEYSTORE }}
         run: |
             cd sdk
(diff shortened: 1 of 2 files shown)


gdeluna-branch:
So browserstack will not accept duplicate uploads? What is the error you get from the upload response?

NidhiDixit09:
Browserstack doesnt return any error - it simply returns same APP_URL. So NEW_APP_URL = OLD_APP_URL. And when test runs, it re-installs app with same app_url (which is already installed). So instead of a re-install, its an app open.

gdeluna-branch (approved):

NidhiDixit09:
@gdeluna-branch I have also updated description. Updated line` test will fail` to `test is ineffective,`.

PR #1213, opened 2024-07-31
add nix

allen-branch:
Adds Nix flake for easy setup of an Android development environment ensuring consistent versions. See https://github.com/tadfisher/android-nixpkgs/tree/main for more info.

### DIFF
--- a/devshell.nix
+++ b/devshell.nix
@@ -1,0 +1,40 @@
+{ pkgs }:
+
+with pkgs;
+
+let
+  jdk = pkgs.jdk17;
+in
+with pkgs;
+
+# Configure your development environment.
+#
+# Documentation: https://github.com/numtide/devshell
+devshell.mkShell {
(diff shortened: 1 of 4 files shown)


devshell.nix
```diff
@@ -0,0 +1,40 @@
+{ pkgs }:
+
+with pkgs;
```
rkim-branch, line 3: I recommend you fold this into the flake.nix file, unless, you also want to support devshell without nix's flake.
It's just one more file people have to worry about

rkim-branch (approved):

devshell.nix
allen-branch: Yeah good point. Let me give this a shot with copilot. I'm sure it's straightforward. There's not really anything special I'm using from devshell.

gdeluna-branch (approved):

PR #1214, opened 2024-08-01
fix CI

allen-branch:
run tests using reactivecircus emulator

SDK-2403


.github/workflows/unit-and-instrumented-tests-action.yml
```diff
@@ -3,167 +3,118 @@ on:
...
-    runs-on: macos-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-      # repo's gradle is configured to run on java 17
-      - name: Setup java 17 for gradle
-        uses: actions/setup-java@v3
+      - name: create AVD and generate snapshot for caching
+        if: steps.avd-cache.outputs.cache-hit != 'true'
+        uses: reactivecircus/android-emulator-runner@v2
```
gdeluna-branch, line 69: We'll need to maintain using verified actions only.

ahmednawar:
@allen-branch @gdeluna-branch what's the next step here?

allen-branch:
> @allen-branch @gdeluna-branch what's the next step here?

@rkim-branch mentioned that the requirements for verified actions only, so we would need to most likely need to our own version of the github action that I attempted to use in this PR.

PR #1219, opened 2024-09-05
[INTENG-20861] v1/install is triggered before Install Referrer Data is fully fetched.

nsingh-branch:
## Reference
INTENG-20861 -- v1/install is triggered before Meta Install Referral Data is fully fetched.  

## Description

When using `deferInitForPluginRuntime` with our React Native plugin, we encountered an issue where not all install referrers were fetched in time to be included in the `v1/install` request. This happened despite having a lock that should have ensured the fetch completed before continuing.

The issue was primarily observed when using `deferInitForPluginRuntime` and React Native, but did not occur consistently in native Android apps or when `deferInitForPluginRuntime` was not used.

The problem stemmed from the `callback.onInstallReferrersFinished()` being called in the finally block, which executed before the coroutine in the try block could finish. This caused the install referrer lock to be added and then instantly removed, leading to incomplete referrer data.

To resolve this, the `callback.onInstallReferrersFinished()` was moved from the finally block into the `resumeWith` and catch blocks. This ensures the callback is only triggered once the coroutine has either successfully completed or an exception has been caught, preventing premature lock removal.

## Testing Instructions

Use the React Native sample app provided in the Jira ticket. Simulate an install with any install referrer and observe the logs to confirm that the lock is added and removed only after the referrers are fully fetched.

## Risk Assessment [`MEDIUM`]

Modifying the initialization code.

- Correctness & Style

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/SystemObserver.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/SystemObserver.java
@@ -600,9 +600,13 @@ abstract class SystemObserver {
 
                 @Override
                 public void resumeWith(@NonNull Object o) {
-                    BranchLogger.v("fetchInstallReferrer resumeWith got result: " + o);
-                    InstallReferrerResult latestReferrer = (InstallReferrerResult) o;
-                    AppStoreReferrer.processReferrerInfo(context_, latestReferrer.getLatestRawReferrer(), latestReferrer.getLatestClickTimestamp(), latestReferrer.getLatestInstallTimestamp(), latestReferrer.getAppStore(), latestReferrer.isClickThrough());
+                    if (o != null) {
+                        BranchLogger.v("fetchInstallReferrer resumeWith got result: " + o);
+                        InstallReferrerResult latestReferrer = (InstallReferrerResult) o;
+                        AppStoreReferrer.processReferrerInfo(context_, latestReferrer.getLatestRawReferrer(), latestReferrer.getLatestClickTimestamp(), latestReferrer.getLatestInstallTimestamp(), latestReferrer.getAppStore(), latestReferrer.isClickThrough());
+                    } else {
+                        BranchLogger.v("fetchInstallReferrer resumeWith got null result");
+                    }
 
                     if (callback != null) {
                         callback.onInstallReferrersFinished();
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/main/java/io/branch/referral/SystemObserver.java
```diff
@@ -600,18 +600,18 @@ public CoroutineContext getContext() {
 
                 @Override
                 public void resumeWith(@NonNull Object o) {
-                    if (o != null) {
```
gdeluna-branch, line 603: While this is annotated with `NonNull`, it is still possible for a null to be passed through at runtime. I've intentionally ignored the warning to remove the redundant null check.

Branch-SDK/src/main/java/io/branch/referral/SystemObserver.java
```diff
@@ -600,18 +600,18 @@ public CoroutineContext getContext() {
...
                 }
             });
-        }
-        catch(Exception e){
+        } catch(Exception e) {
             BranchLogger.e("Caught Exception SystemObserver fetchInstallReferrer " + e.getMessage());
-        }
-        finally {
-            if(callback != null){
+            if (callback != null) {
```
gdeluna-branch, line 614: While this should work, what was the thread order of 
`fetchLatestInstallReferrer`
`resumeWith`
and the `finally` block?

nsingh-branch: I believe the order goes:
1. `fetchLatestInstallReferrer` is called as a coroutine
2. `resumeWith` is called when the coroutine finishes
3. `finally` usually gets called when `fetchLatestInstallReferrer` finishes, but i was seeing it get called right after the coroutine starts in this case.

nsingh-branch: @gdeluna-branch After printing out the threads I can see the order is
1. `initTasks` on mqt_native_modules
2. `fetchInstallReferrer` on mqt_native_modules
3. `fetchInstallReferrer finally` block on mqt_native_modules
4. `fetchInstallReferrer` `resumeWith` on DefaultDispatcher-worker-1
5. `initTasks onInstallReferrersFinished` on DefaultDispatcher-worker-1

gdeluna-branch (approved):

PR #1259, opened 2025-04-15
Remove user agent string query from start up

gdeluna-branch:
Originally this code replicated behavior to fetch the user agent string after opens. However, even though it is dispatched, it still requires main thread https://issuetracker.google.com/issues/289118199#comment9. For integrations with heavy start up sequences, this still has a performance impact of about 50ms to 200ms. 
After consideration, this technique does not offer a favorable trade off between impacting start up over invoking only when necessary.

This change only queries when needed by v2 event. The code to use either the `WebSettings` static or the `WebView` instance remains unchanged as that is needed to workaround certain external Chromium package crashes.

## Reference
SDK-XXX -- <TITLE>.

## Testing Instructions
Run the test app to validate that the user agent string is still present on v2 events. 
- Include queueing events prior to open.
- User agent string query only called when a v2 event is enqueued.

- Correctness & Style


NidhiDixit09 (approved):
Looks Good to me. Tested both scenarios. In both cases v1/open request had no user agent field whereas v2/event had this string.

PR #1269, opened 2025-05-21
INTENG-22685 - Integration validator protected endpoint bugfix

rob-gioia-branch:
## Reference
INTENG-22685

## Description
If a customer is using Advanced Compliance and then runs the Integration Validator code, the Integration Validator will not work. This is because the protected-api endpoint gets used for the requests, and there is no app-settings protected equivalent for that endpoint. Since the Integration Validator code is never released into the wild, but only used for testing, we can safely keep that endpoint going to the regular Branch endpoint, with all other calls like opens, installs, etc... still going through the protected endpoint for Advanced Compliance customers. To address this, I've added a boolean parameter to the getAPIBaseUrl() function, which can take in false in the case of any test code, like the integration validator, so that the regular endpoint (instead of protected-api) is used.  

## Risk Assessment [`LOW`]

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
@@ -239,8 +239,8 @@ public class PrefHelper {
      * @return A {@link String} variable containing the hard-coded base URL that the Branch
      * API uses.
      */
-    public String getAPIBaseUrl() {
-        if (URLUtil.isHttpsUrl(customServerURL_)) {
+    public String getAPIBaseUrl(boolean useCustom) {
+        if (useCustom && URLUtil.isHttpsUrl(customServerURL_)) {
             return customServerURL_;
         }
 
--- a/Branch-SDK-TestBed/src/main/java/io/branch/branchandroidtestbed/SettingsActivity.java
+++ b/Branch-SDK-TestBed/src/main/java/io/branch/branchandroidtestbed/SettingsActivity.java
@@ -54,7 +54,7 @@ public class SettingsActivity extends Activity {
     void setupApiUrlText() {
         final EditText apiUrlText = findViewById(R.id.api_url_text);
         final PrefHelper prefHelper = PrefHelper.getInstance(this);
-        String currentApiUrl = prefHelper.getAPIBaseUrl();
+        String currentApiUrl = prefHelper.getAPIBaseUrl(true);
 
         apiUrlText.setText(currentApiUrl);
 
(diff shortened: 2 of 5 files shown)


Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -239,8 +239,8 @@ static void setAPIUrl(String url) {
      * @return A {@link String} variable containing the hard-coded base URL that the Branch
      * API uses.
      */
-    public String getAPIBaseUrl() {
-        if (URLUtil.isHttpsUrl(customServerURL_)) {
+    public String getAPIBaseUrl(boolean useCustom) {
+        if (useCustom && URLUtil.isHttpsUrl(customServerURL_)) {
```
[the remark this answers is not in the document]
rob-gioia-branch: Omitting this change since that is something outside of this PR (pre-existing code not related to the change)

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -239,8 +239,8 @@ static void setAPIUrl(String url) {
      * @return A {@link String} variable containing the hard-coded base URL that the Branch
      * API uses.
      */
-    public String getAPIBaseUrl() {
```
gdeluna-branch, line 242: Why not keep this function and add the value into the overloaded function?

rob-gioia-branch: Refactored per your suggestion @gdeluna-branch !

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -255,6 +256,26 @@ public String getAPIBaseUrl() {
...
+    /**
+    * Overloaded version of the getAPIBaseUrl() function that allows specifying if the custom URL should be used
+     * Important for endpoints that do not have a protected version, such as the App Settings endpoint used by the Integration Validator
+    */
+    public String getAPIBaseUrl(boolean useCustom) {
+        if (useCustom && URLUtil.isHttpsUrl(customServerURL_)) {
+            return customServerURL_;
+        }
+
+        if (useEUEndpoint_) {
```
gdeluna-branch, line 268: I think you can call `getAPIBaseUrl` after the first `if`
```
       if (useCustom && URLUtil.isHttpsUrl(customServerURL_)) {
             return customServerURL_;
        }
       else {
             return getAPIBaseUrl();
       }
```

to reduce duplicate code, this make sense?

gdeluna-branch: Ah no actually because `getAPIBaseUrl` will return the custom url. Let's leave this as you've done it here.

gdeluna-branch (approved):

PR #1353, opened 2026-06-23
Fix androidTest compilation (canClearInitData port + setInitState signature)

wpinho-branch:
## Summary

The `androidTest` source set has not compiled since [5fab3857](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/commit/5fab3857) (EMT-2260), due to two migration breakages in the test harness. This blocks the entire instrumented suite from building or running.

## Changes

1. **`canClearInitData` parity** — `BranchMigrationTest` calls `adapter.canClearInitData()`, which existed only on the legacy `ServerRequestQueue` and was never ported to `BranchRequestQueueAdapter`. Added `canClearInitData()` to `BranchRequestQueue` (counts queued `ServerRequestInitSession`, true when ≤1 remains) and delegated from the adapter.
2. **`setInitState` signature** — `BranchTest.tearDown()` passed the legacy `Branch.SESSION_STATE.UNINITIALISED` enum to `setInitState`, which now takes a `BranchSessionState`. Use `BranchSessionState.Uninitialized`.

## Verification

`./gradlew :Branch-SDK:compileDebugAndroidTestJavaWithJavac` → BUILD SUCCESSFUL (previously failed on `Unresolved reference: canClearInitData` then `SESSION_STATE cannot be converted to BranchSessionState`).

## Why this matters

The instrumented suite being uncompilable since 5fab3857, together with the `unit-and-instrumented-tests` workflow being disabled, means the cold-start regressions tracked in EMT-3859 / EMT-3860 shipped to beta with no instrumented coverage running. This unblocks that coverage and is a prerequisite for the EMT-3860 cold-start regression test.

### DIFF
--- a/Branch-SDK/src/androidTest/java/io/branch/referral/BranchTest.java
+++ b/Branch-SDK/src/androidTest/java/io/branch/referral/BranchTest.java
@@ -48,7 +48,7 @@ abstract public class BranchTest extends BranchTestRequestUtil {
         }
 
         if (branch != null) {
-            branch.setInitState(Branch.SESSION_STATE.UNINITIALISED);
+            branch.setInitState(BranchSessionState.Uninitialized.INSTANCE);
             Branch.shutDown();
             branch = null;
         }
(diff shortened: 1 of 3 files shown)


wpinho-branch:
Removed the leftover `DEBUG:` log in `canClearInitData()` (b0946130). Behavior is unchanged and the unit module stays green.

gdeluna-branch (approved):

PR #1357, opened 2026-06-24
EMT-3863 -- Restore a deprecated source-compat alias for the relocated LATD listener

wpinho-branch:
A beta tester hit a compile break upgrading 5.16.3 to the beta: the 6.0 refactor moved the nested `ServerRequestGetLATD.BranchLastAttributedTouchDataListener` to `Branch.BranchLastAttributedTouchDataListener` and left nothing at the old location, so their `new ServerRequestGetLATD.BranchLastAttributedTouchDataListener(){...}` stopped compiling.

The move itself is correct — that listener never belonged on an internal request class — so I keep it and re-add the old nested type as a `@Deprecated` alias that extends the new one. Same single method (`onDataFetched`), so existing anonymous implementations satisfy the relocated callsites with no cast. I also repointed the two `getLastAttributedTouchData` javadoc `@param` links, which still named the old location.

This restores **source** compatibility only. A prebuilt app that isn't recompiled would still hit `NoSuchMethodError`, because the callsite's method descriptor changed when the parameter type moved — the javadoc states this. One caveat worth calling out for integrators: the alias emits a deprecation warning, so a team building with `-Werror` / warnings-as-errors must suppress it or migrate to `Branch.BranchLastAttributedTouchDataListener`.

Test compiles the reported 5.x usage and asserts `@Deprecated`: RED (compile error "cannot find symbol") before, GREEN after. Full unit module green (414/0).

Merge order: independent of the other beta fixes; base `6.0.0-beta.0`.

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/Branch.java
@@ -2585,7 +2585,7 @@ public class Branch {
      * saved via PreferenceHelper.setLATDAttributionWindow(). If no value has been saved, Branch
      * defaults to a 30 day attribution window (SDK sends -1 to request the default from the server).
      *
-     * @param callback An instance of {@link io.branch.referral.ServerRequestGetLATD.BranchLastAttributedTouchDataListener}
+     * @param callback An instance of {@link io.branch.referral.Branch.BranchLastAttributedTouchDataListener}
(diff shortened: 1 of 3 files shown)


gdeluna-branch (approved):
Please merge into beta, not alpha since alpha is not supported.

PR #1358, opened 2026-06-24
EMT-3861 -- Restore the no-argument Branch.logout() overload

wpinho-branch:
A beta tester flagged that `branch.logout()` no longer compiles — the beta exposes only `logout(LogoutStatusListener)`. 5.x shipped both, with the no-arg form delegating (`logout(){ logout(null); }`); it was dropped during the EMT-2260 queue cleanup, and since the behavior it provided still exists, the removal looks accidental.

I restore exactly that overload. It's behavior-neutral: `QueueOperationLogout` types its callback as nullable and guards every dereference, so `logout(null)` is safe, and there's no overload ambiguity (distinct arity).

Test is a reflection check that `logout()` exists and `logout(LogoutStatusListener)` is unchanged: RED (`NoSuchMethodException`) before, GREEN after. Full unit module green (414/0).

Merge order: independent of the other beta fixes; base `6.0.0-beta.0`.

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/Branch.java
@@ -1011,6 +1011,15 @@ public class Branch {
      * <p>This method should be called if you know that a different person is about to use the app. For example,
      * if you allow users to log out and let their friend use the app, you should call this to notify Branch
      * to create a new user for this device. This will clear the first and latest params, as a new session is created.</p>
+     */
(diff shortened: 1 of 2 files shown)


gdeluna-branch:
Why not point this to beta? 6.0.0-beta.0 is the current. Otherwise code looks good. We might want to update the comment language later.

gdeluna-branch (approved):
Please merge to beta.

PR #1359, opened 2026-06-24
EMT-3864 -- Gate queue/lock trace behind an opt-in flag so enableLogging() shows traffic

wpinho-branch:
A beta tester reported `enableLogging()` "no longer displays network requests." They weren't gone: the request/response lines log at VERBOSE, and the 6.0 queue rewrite added ~187 queue/lock trace lines that also printed at VERBOSE and buried the real ones.

## Approach (revised per review)

The first revision gated the trace behind a custom `BranchLogger.t()` channel plus a separate `traceLoggingEnabled` flag. Per review feedback (two gates and a non-standard log method are friction the SDK should not add), that was dropped. Every touched statement is now classified to a standard `android.util.Log` severity, so the log level alone separates the output, per https://source.android.com/docs/core/tests/debug/understanding-logging:

- **DEBUG** = network request/response traffic (`getting` / `posting to` / `Server returned`) — what a developer wants when they turn logging on.
- **VERBOSE** = the ~187 internal request queue and lock trace lines (constructor, enqueue, lock acquire/release, state transitions, `WAIT_LOCK_DEBUG`, `MODERNIZATION_TRACE`), matching the platform guidance that VERBOSE is development-only, most-detailed output.
- `enableLogging()` now defaults to **DEBUG**, so it shows clean traffic again. A developer diagnosing a stuck request uses the existing `enableLogging(BranchLogger.BranchLogLevel.VERBOSE)` overload to add the queue/lock detail.

One gate (the log level), standard severities, no new public API and no custom log method.

**Removed:** `BranchLogger.t()`, `traceLoggingEnabled`, `Branch.setTraceLogging()`.

## Scope

All `.t()` reclassified to `.v()` across `Branch.java`, `ServerRequestInitSession`, `BranchRequestQueue`, `BranchRequestQueueAdapter`, `ModernLinkGenerator`. The traffic lines in `BranchRemoteInterface` promoted `.v()` -> `.d()`. Full request/response payloads (`Post value`, the full `returned` body) stay at VERBOSE, to keep DEBUG readable and avoid dumping request bodies by default. The STUCK_LOCK/queue diagnostics are new in the 6.0 queue (the file does not exist on 5.x), so nothing here is a regression from a shipped release.

## Verification

`./gradlew :Branch-SDK:testDebugUnitTest --rerun-tasks` -> **409 tests, 0 failures, 0 errors**. `TraceLoggingTest` asserts: DEBUG shows traffic and hides the queue trace, VERBOSE shows both, nothing emits while logging is disabled, and `enableLogging()` resolves to DEBUG.

## Changelog (6.0)

`enableLogging()` now defaults to DEBUG and shows request/response traffic. The queue/lock diagnostics moved to `enableLogging(BranchLogger.BranchLogLevel.VERBOSE)`.

Base `6.0.0-beta.0`.


Branch-SDK/src/main/java/io/branch/referral/BranchLogger.kt
```diff
@@ -89,6 +98,24 @@ object BranchLogger {
...
+    /**
+     * <p>Creates an internal <b>Trace</b> message (request queue / lock diagnostics). Gated by
+     * [traceLoggingEnabled] independently of [loggingLevel], so it stays out of normal logging
+     * until a developer opts in via Branch.setTraceLogging(true). Requires logging to be enabled.</p>
+     *
+     * @param message A {@link String} value containing the trace message to record.
+     */
+    @JvmStatic
+    fun t(message: String?) {
+        if (loggingEnabled && traceLoggingEnabled && message?.isNotEmpty() == true) {
```
gdeluna-branch: Having two gates is more friction and currently the logging statements are not standardized. `.t` is not standard and I don't want the SDK to have custom practices. 

I think the solution is to audit all logs in a given scenario and organize them by severity that conform to the platform standard. https://source.android.com/docs/core/tests/debug/understanding-logging

gdeluna-branch (changes requested):

Branch-SDK/src/main/java/io/branch/referral/BranchLogger.kt
wpinho-branch: Agreed on both points. Dropping the custom .t() channel and the separate traceLoggingEnabled gate, and standardizing every statement to the platform severities instead.

gdeluna-branch (approved):

Branch-SDK/src/main/java/io/branch/referral/BranchLogger.kt
wpinho-branch: Done in the last push. The custom .t() channel and the traceLoggingEnabled gate are both gone, and every statement now maps to a platform severity: DEBUG for request/response traffic, VERBOSE for queue internals.

PR #1360, opened 2026-06-24
SDK-2463 -- Proposed Solution: Skip the Open on configuration-change activity recreation

wpinho-branch:
Folding/unfolding a foldable logs a spurious `Open` even though the app never leaves the foreground. Same thing happens on rotation: any configuration change that recreates the foregrounded activity destroys + recreates it, so the visible-activity count momentarily hits 0 and the recreated activity logs an Open.

`BranchOpenObserver` (the observer registered in 6.0) fires `sendOpen()` whenever the count goes 0 → 1. The fix records, on the count → 0 stop, whether the activity is being destroyed for a configuration change (`activity.isChangingConfigurations`), and skips `sendOpen()` on the matching 0 → 1 start. Real foregrounds still fire the Open. This mirrors how AndroidX `ProcessLifecycleOwner` uses `isChangingConfigurations()` in `onStop` to avoid emitting an event on a config-change recreation.

A note on how I got here: I first patched `BranchActivityLifecycleObserver`, but an adversarial review caught that that class isn't the one registered in 6.0 — `setActivityLifeCycleObserver` registers `BranchOpenObserver`. Verified with a grep before redoing the fix on the live observer. Worth knowing for reviewers.

**Scope — this is also on 5.x, but via a different mechanism.** `master` registers `BranchActivityLifecycleObserver`, which *closes the session* on count 0 and re-opens on the recreated resume (a different code path). That needs its own guard on `closeSessionInternal()`; I have that draft ready. This PR fixes the 6.0 `BranchOpenObserver` path only — the 5.x patch should be a separate ticket/PR.

Heads-up for reviewers:
- This shifts Open/session counts on foldable- and rotation-heavy cohorts (it stops the spurious churn). Correct behavior, but worth a line in the 6.0 changelog since it's a measurable change for customers.
- Known edge: if a config-change recreation never restarts the activity (app backgrounded mid-recreate), the flag could suppress one later Open. Rare — config-change recreations complete in the same loop.

Tests: a config-change recreation fires no second Open; a real background → foreground still does (RED with the guard removed, GREEN after). Full unit module green.

Merge order: independent; base `6.0.0-alpha.1`.

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/observers/BranchOpenObserver.kt
+++ b/Branch-SDK/src/main/java/io/branch/observers/BranchOpenObserver.kt
@@ -11,13 +11,25 @@ internal class BranchOpenObserver(private val branchInstance: Branch) : Applicat
 
     private var activityCount = 0
 
+    // SDK-2463: set when the last visible activity stops because it is being recreated for a
+    // configuration change (folding/unfolding a foldable, rotation, etc.). It tells the matching
+    // onActivityStarted that the foreground->foreground recreation is not a real app open.
+    private var recreatingForConfigChange = false
+
     override fun onActivityStarted(activity: Activity) {
         activityCount++
         BranchLogger.v("BranchOpenObserver onActivityStarted: $activity activityCount incremented to: $activityCount")
 
         if (activityCount == 1) {
-            branchInstance.sendOpen()
+            if (recreatingForConfigChange) {
+                // The single visible activity was just recreated for a configuration change; the app
+                // never left the foreground, so logging an Open here would be a spurious open.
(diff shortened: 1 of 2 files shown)


wpinho-branch:
Superseded by #1364 (same SDK-2463 fix, refined) and by an incoming ProcessLifecycleOwner prototype. Suggest closing this so we keep a single SDK-2463 PR open.

wpinho-branch:
Closing in favor of #1366 (ProcessLifecycleOwner approach).

This approach was superseded by the guard (#1364) and then by ProcessLifecycleOwner (#1366). We're proceeding with #1366 as the production solution for SDK-2463.

PR #1362, opened 2026-06-24
EMT-3870 -- Track request retry/timeout on the monotonic clock (elapsedRealtime)

wpinho-branch:
`RequestRetryInfo` measured elapsed time for retry/timeout tracking with `System.currentTimeMillis()` (the wall clock). A device clock change (NTP correction, user edit, timezone/DST) can move that clock arbitrarily, so a request that hasn't really been waiting can look timed out, or one that has can be masked.

Measure elapsed time with `SystemClock.elapsedRealtime()` (monotonic, unaffected by wall-clock changes) across `firstAttemptTime`, `lastAttemptTime`, `firstWaitLockTime`, `hasExceededTimeout`, and `hasExceededWaitLockTimeout`.

The test captures `firstAttemptTime` on the monotonic clock and asserts a fresh request is not timed out. It FAILS on the wall-clock version (epoch-vs-uptime mismatch) and PASSES after the fix — proven by stashing the production change. Full unit module green.

**Base:** `6.0.0-beta.0` — the `currentTimeMillis` tracking predates EMT-3859, so this is independent of the other queue PRs.


Branch-SDK/src/main/java/io/branch/referral/BranchRequestQueue.kt
```diff
@@ -21,30 +22,34 @@ import java.util.concurrent.atomic.AtomicInteger
...
+    fun hasExceededTimeout(timeoutMs: Long): Boolean =
+        (SystemClock.elapsedRealtime() - firstAttemptTime) > timeoutMs
+
+    fun hasExceededWaitLockTimeout(timeoutMs: Long): Boolean =
+        firstWaitLockTime > 0 && (SystemClock.elapsedRealtime() - firstWaitLockTime) > timeoutMs
+
     fun incrementRetry() {
         retryCount++
-        lastAttemptTime = System.currentTimeMillis()
+        lastAttemptTime = SystemClock.elapsedRealtime()
```
gdeluna-branch, line 47: When is `lastAttemptTime` used?

gdeluna-branch (approved):
Please remove any unused code, otherwise looks good.

Branch-SDK/src/main/java/io/branch/referral/BranchRequestQueue.kt
wpinho-branch: It isn't. Nothing reads it.

The field predates this PR. I only switched its clock source along with the other three, so the struct doesn't mix wall clock and monotonic time. It's dead either way.

Removing it felt out of scope here, so I'm tracking it as a follow-up. Happy to drop it in this PR if you'd rather.

wpinho-branch: Tracked as EMT-3970.

PR #1363, opened 2026-06-24
EMT-3877 -- Add a public-API diff gate against the last 5.x release

wpinho-branch:
The beta removed/moved several public APIs without anyone noticing until testers reported it (no-arg `logout()`, `BranchUniversalObject.showShareSheet()`, the synchronous deep-link param getters, the relocated `BranchLastAttributedTouchDataListener`). This adds an automated gate so accidental public-API removals in 6.0 are caught at build time.

## What it does
A `verification` task **`apiCompatibilityReport`** that:
1. Resolves the last stable 5.x release AAR (`io.branch.sdk.android:library:5.21.1`) as a baseline.
2. Builds the current AAR and extracts `classes.jar` from both.
3. Runs japicmp (CLI fat jar via `JavaExec`) scoped to `io.branch`, public API, binary-incompatible changes, and writes `Branch-SDK/build/reports/api-diff/api-diff.{txt,html}` plus a console summary.

**Report mode by default — it does not fail the build**, because 6.0 intentionally changes some APIs. Pass `-PapiDiffStrict` to make it fail on binary-incompatible removals; flip the default once the 6.0 public API is frozen.

## Evidence
`./gradlew :Branch-SDK:apiCompatibilityReport` runs green and the report flags the real beta removals against 5.21.1 — `showShareSheet(...)`, `getFirstReferringParamsSync()`/`getLatestReferringParamsSync()`, `ServerRequestGetLATD$BranchLastAttributedTouchDataListener`, and `void logout()` — exactly the regressions this gate exists to catch. `-PapiDiffStrict` fails the build on those.

## Notes
- japicmp CLI fat jar (0.23.1) over the Gradle plugin: runs in its own JVM via `JavaExec`, no plugin-classpath friction on AGP 8.12.2 / Gradle 8.13.
- `--ignore-missing-classes` is intentional (the AAR `classes.jar` has no transitive deps); it only silences absent framework supertypes, all SDK removals are still reported.
- Self-contained: only `Branch-SDK/build.gradle.kts` changes; no source or version touched.

**Base:** `6.0.0-beta.0` (standalone build tooling, independent of the other PRs).

### DIFF
--- a/Branch-SDK/build.gradle.kts
+++ b/Branch-SDK/build.gradle.kts
@@ -1,4 +1,14 @@
 import org.gradle.api.tasks.testing.logging.*
+import java.io.OutputStream
+
+// EMT-3877 — tees an OutputStream to two sinks (console + report file) so the
+// japicmp diff is both printed and persisted. Used by `apiCompatibilityReport`.
+class TeeOutputStream(private val a: OutputStream, private val b: OutputStream) : OutputStream() {
+    override fun write(byte: Int) { a.write(byte); b.write(byte) }
+    override fun write(bytes: ByteArray, off: Int, len: Int) { a.write(bytes, off, len); b.write(bytes, off, len) }
+    override fun flush() { a.flush(); b.flush() }
+    override fun close() { b.close() } // never close System.out
+}
 
 plugins {
     id("com.android.library")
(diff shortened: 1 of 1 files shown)


wpinho-branch:
Addressed the unclosed-stream finding: `TeeOutputStream` is now closed in `doLast` (it closes only the file stream, never `System.out`), so the japicmp text-report handle no longer leaks in the Gradle daemon.

gdeluna-branch (approved):

PR #1364, opened 2026-06-30
SDK-2463 -- Guard against duplicate OPEN on foldable configuration changes

wpinho-branch:
## Why

On a foldable, folding or unfolding fires a configuration change that destroys and recreates the foregrounded Activity. The observer counts the stop/start as a fresh foreground and emits a second OPEN, even though the app never actually re-opened. This silently inflates open metrics for customers as foldables become more common.

## What

Track `Activity.isChangingConfigurations()` on stop and skip the OPEN on the immediately following start. A genuine background-to-foreground still emits OPEN.

## Test

Added a unit test on the observer:
- fold/unfold while foregrounded emits a single OPEN (fails without the guard)
- a real background-to-foreground emits OPEN again

## Scope

This is the minimal fix for the duplicate OPEN. Making the SDK fully fold-aware (dedicated Fold/Unfold events via WindowManager / FoldingFeature) is out of scope here.

## Notes

Only reproduces when the app lets the Activity be recreated on fold. Apps that declare screenSize/screenLayout in configChanges already avoid it.

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/observers/BranchOpenObserver.kt
+++ b/Branch-SDK/src/main/java/io/branch/observers/BranchOpenObserver.kt
@@ -10,17 +10,23 @@ import android.os.Bundle
 internal class BranchOpenObserver(private val branchInstance: Branch) : Application.ActivityLifecycleCallbacks {
 
     private var activityCount = 0
+    private var isChangingConfiguration = false
 
     override fun onActivityStarted(activity: Activity) {
         activityCount++
         BranchLogger.v("BranchOpenObserver onActivityStarted: $activity activityCount incremented to: $activityCount")
 
-        if (activityCount == 1) {
+        if (activityCount == 1 && !isChangingConfiguration) {
             branchInstance.sendOpen()
         }
+        isChangingConfiguration = false
     }
 
     override fun onActivityStopped(activity: Activity) {
+        // A configuration change (e.g. folding/unfolding a foldable) destroys and recreates the
+        // foregrounded Activity. Remember it so the following onActivityStarted does not emit a
+        // duplicate OPEN for what is still the same session (SDK-2463).
+        isChangingConfiguration = activity.isChangingConfigurations
         activityCount--
         BranchLogger.v("BranchOpenObserver onActivityStopped: $activity activityCount decremented to: $activityCount")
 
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/main/java/io/branch/observers/BranchOpenObserver.kt
```diff
@@ -10,17 +10,23 @@ import android.os.Bundle
...
 
     private var activityCount = 0
+    private var isChangingConfiguration = false
 
     override fun onActivityStarted(activity: Activity) {
         activityCount++
         BranchLogger.v("BranchOpenObserver onActivityStarted: $activity activityCount incremented to: $activityCount")
 
-        if (activityCount == 1) {
+        if (activityCount == 1 && !isChangingConfiguration) {
```
gdeluna-branch, line 19: Please look into `ProcessLifecycleOwner`, this observer should handle our use cases. Will provide details offline.

gdeluna-branch (changes requested):

wpinho-branch:
Alternative approach prototyped in #1366 using `ProcessLifecycleOwner` (process-level foreground detection). On-device benchmark (Pixel 8 Pro, API 34): both this guard and #1366 fix the duplicate OPEN (configuration change: 0 emitted vs baseline 1). #1366 trades one androidx dependency for handling multi-window/PIP/multi-Activity by construction; this guard is the smaller, lower-risk change. One of the three SDK-2463 PRs should win.

wpinho-branch:
@gdeluna-branch 
I implemented the `ProcessLifecycleOwner` approach in #1366 and validated it on-device. `ON_START` fires only on a real process foreground, so a config-change recreation never emits a second OPEN, and it also removes the stale-flag edge case this guard carries. The only trade-off is the `androidx.lifecycle:lifecycle-process` dependency.

My suggestion is to merge #1366 and close this one and #1360. If we'd rather not take the dependency into the beta, #1364 stays as the zero-dependency fallback. Let me know which you prefer.

gdeluna-branch:
Closing in favor of #1366

wpinho-branch:
Closing in favor of #1366 (ProcessLifecycleOwner approach).

After evaluation, we're proceeding with the ProcessLifecycleOwner solution which:
- Eliminates the stale-flag OPEN suppression bug by construction
- Provides 700ms debounce for quick app switches
- Is the idiomatic Android approach for process-level lifecycle detection

The guard approach (#1364) was a good minimal fix for the immediate beta, but ProcessLifecycleOwner is the right long-term solution for 6.0.

The lifecycle-process dependency (~30KB) is acceptable given the correctness improvements.

PR #1365, opened 2026-06-30
EMT-3882 -- Restore synchronous deep-link param getters

wpinho-branch:
## Why

5.x exposed `getFirstReferringParamsSync()` and `getLatestReferringParamsSync()`. The beta dropped them, leaving only async access (`requestDeepLinkData`). Both synchronous getters are restored so existing 5.x integrations keep working.

## What

- `getFirstReferringParamsSync()` and `getLatestReferringParamsSync()` restored with the 5.x latch semantics: block up to `LATCH_WAIT_UNTIL` (2500ms) for the data, then return the locally stored params.
- The latches are released from `ServerRequestInitSession.onInitSessionCompleted()`, the beta equivalent of the old queue's init-complete hook (5.x released them in `ServerRequestQueue`).
- Both delegate to the existing `getFirstReferringParams()` / `getLatestReferringParams()` for the actual read, so behavior matches the async getters once data is available.

## Test

`SyncReferringParamsApiTest` pins the public contract (both methods exist, public, no-arg, return `JSONObject`) and guards against the API being removed again. It fails with `NoSuchMethodException` without the restoration. Full `:Branch-SDK:testDebugUnitTest` is green.

Note on coverage: the suite mocks `Branch` everywhere and has no real-instance harness, so behavioral latch-release (block then release on init) is exercised by the init flow rather than a unit test.

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/referral/Branch.java
+++ b/Branch-SDK/src/main/java/io/branch/referral/Branch.java
@@ -38,6 +38,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import io.branch.coroutines.RequestDeepLink;
 import io.branch.indexing.BranchUniversalObject;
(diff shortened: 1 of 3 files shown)


gdeluna-branch (approved):
Please merge to beta.

PR #1366, opened 2026-07-02
SDK-2463 -- ProcessLifecycleOwner for foreground detection (alternative to #1364, validated)

wpinho-branch:
## Summary

Fixes SDK-2463 (duplicate OPEN on foldable configuration changes) by detecting app foreground/background at the **process level** with AndroidX `ProcessLifecycleOwner`, replacing the per-Activity start/stop counting in `BranchOpenObserver`.

This is an alternative to the surgical guard in #1364 (and the earlier #1360). All three fix the reported bug; this PR additionally removes a class of edge-case bugs the guard carries and is the idiomatic Android approach, at the cost of one AndroidX dependency.

> **This description is the single, current source of truth for the guard-vs-PLO comparison. Earlier inline comments are kept as history — one of them ("debounce not reproducible") was later disproven; see the on-device proof below.**

## What this PR does

- New `BranchProcessLifecycleObserver` (`DefaultLifecycleObserver`): `ON_START` sends the OPEN, `ON_STOP` is a no-op (matches current beta behaviour, which does not close the session on background).
- Registered during init in place of `BranchOpenObserver`, which is deleted.
- Adds `androidx.lifecycle:lifecycle-process:2.5.1` (pinned for Kotlin 1.6.21).
- Unit tests pin the observer contract and the no-suppression property.

`ProcessLifecycleOwner` ON_START fires only on a real process foreground (cold start / return from background), never on a config-change recreation, inter-Activity navigation, multi-window, or PIP. It also debounces ON_STOP by 700ms.

## Guard (#1364) vs ProcessLifecycleOwner (this PR) — evidence

| Property                                       | Guard #1364        | PLO (this PR)               | How verified                                                                     |
| ---------------------------------------------- | ------------------ | --------------------------- | -------------------------------------------------------------------------------- |
| Fixes the foldable duplicate OPEN              | ✅ 0 (baseline 1)  | ✅ 0                        | On-device, dark-mode config-change trigger                                       |
| Cold start / real background→foreground        | ✅ 1 / 1           | ✅ 1 / 1                    | On-device                                                                        |
| Stale-flag suppresses a legitimate OPEN        | ❌ **has the bug** | ✅ immune by construction   | Deterministic unit test (guard expects 2 OPENs, gets 1)                          |
| 700ms quick-switch debounce (no spurious OPEN) | ❌ none            | ✅ yes                      | Instrumented test on-device (quick 300ms → 1, slow 1500ms → 2; stable ×2, ~7.9s) |
| Multi-window / PIP                             | ❓ unproven        | ❓ unproven (architectural) | Needs a multi-activity harness — tracked in EMT-3900                             |

Honesty note on the debounce: it is a proven `ProcessLifecycleOwner` behaviour, but its practical guard-vs-PLO impact is confined to a narrow window, since the activity's own stop timing (~750ms after backgrounding) is close to the 700ms debounce. The mechanism difference is real; the everyday difference is small.

## Static comparison

|                             | Guard #1364                                           | PLO (this PR)                                             |
| --------------------------- | ----------------------------------------------------- | --------------------------------------------------------- |
| Production diff             | 8 lines in `BranchOpenObserver`                       | new observer (+47) / delete old (−48) / swap registration |
| New dependency              | none                                                  | `lifecycle-process` (~30KB, near-ubiquitous)              |
| Config change               | `isChangingConfigurations` flag                       | handled by construction (process-level)                   |
| Multi-Activity interleaving | shared flag (edge cases)                              | no per-Activity state                                     |
| Known edge-case bug         | yes (stale-flag OPEN suppression)                     | none of that class                                        |
| Mechanism change            | minimal (same counting)                               | replaces the detection mechanism                          |
| Unit test                   | mock Activity stop/start + `isChangingConfigurations` | direct ON_START/ON_STOP contract + no-suppression         |

## Recommendation

- **Close SDK-2463 in the beta with minimum risk this week → #1364** (8 lines, no new dependency).
- **Do it the right way for 6.0, if the AndroidX dependency is acceptable → this PR**, and delete the guard. It fixes the bug, eliminates the guard's OPEN-suppression edge case, and adds the debounce, in idiomatic form.

Net: trading one ubiquitous AndroidX dependency for eliminating a real bug + gaining the debounce + idiomatic code is a good long-term deal. The dependency is the only real trade-off.

## Open items

- **Decision needed** on accepting the `lifecycle-process` dependency (the only trade-off).
- **Multi-window / PIP** not yet proven — tracked in **EMT-3900**, which also carries the on-device instrumented debounce test. That test can only be committed once **EMT-3875** makes the `androidTest` set compile again, and the SDK sets an explicit `targetSdk` (today only `minSdk` is set, which makes the test APK target 21 and hit the legacy permission-review screen).
- Only one SDK-2463 PR should merge; close **#1360** and whichever of #1364 / this PR is not chosen.

## Notes

- `onStop` is intentionally a no-op (the current beta observer does not close the session on background); a future session-close would go there.


wpinho-branch:
## Extended validation (follow-up)

**Emulator (Pixel 8 Pro, API 34) — every reliably reproducible scenario shows this PR and the #1364 guard behaving identically:**

| Scenario | Baseline beta.0 | #1364 guard | This PR (PLO) |
|---|---|---|---|
| Config-change recreation (dark-mode toggle) | 1 (bug) | 0 | 0 |
| Quick app switch ~0.5s (pause/resume, no stop) | 0 | 0 | 0 |
| Real background → foreground | 1 | 1 | 1 |

The ProcessLifecycleOwner-specific edges (700ms ON_STOP debounce on a fast switch, multi-window/PIP) require the activity to fully stop inside a narrow window, or a multi-activity/multi-window harness — not cleanly scriptable via adb on a single-activity TestBed. **So on the common paths, adopting PLO is low behavioral risk: it matches the guard's observable behavior exactly.**

**The one difference I could prove deterministically is a correctness bug in the guard, not in PLO:**

- A config-change stop sets `BranchOpenObserver`'s shared `isChangingConfigurations` flag. If the recreation's start is deferred (app backgrounded mid-recreate) and a genuine foreground follows, the stale flag **suppresses that legitimate OPEN**. A unit test on the guard path (expected 2 OPENs: cold + real foreground) **fails with 1**.
- PLO holds no per-Activity state — `repeatedForegrounds_neverSuppressAnOpen` passes (3 foregrounds → 3 OPENs). ProcessLifecycleOwner never dispatches ON_START for a config change, so the whole stale-flag class cannot occur.

**Net:** both fix the reported foldable duplicate OPEN identically; PLO additionally eliminates a (rare) OPEN-suppression bug the guard carries, for the cost of one androidx dependency. Also hardened here: dropped the now-unused `Application` param from the init path and added the regression test. 409 unit tests green, lint clean.

wpinho-branch:
> **⚠️ UPDATE / SUPERSEDED:** the debounce WAS reproducible — my adb method was wrong. It is now proven on-device with an instrumented test (see the later correction comment and the PR description). The text below is kept for history.

---

## Deeper validation of the remaining claims (700ms debounce, multi-window)

Chased the two scenarios the first pass could not reproduce. Findings:

**700ms ON_STOP debounce — real in design, not cleanly reproducible on this emulator.** Measured the lifecycle timing: after HOME the activity hits `onPause` immediately but `onActivityStopped` only ~750ms later, and ProcessLifecycleOwner's pause-debounce is ~700ms from `onPause`. Those two windows nearly coincide, so:
- Return < ~0.7s: the activity only pauses (never stops) -> **both** guard and PLO emit 0 (nothing to react to).
- Return 0.8-1.1s: the process ON_STOP had already fired in every run -> **both** emit 1.

So on this device the distinguishable window collapses to near-zero; I could not produce a run where PLO suppressed an OPEN the guard emitted. The debounce advantage is genuine ProcessLifecycleOwner behaviour but marginal in practice here.

**Multi-window** — freeform toggle via adb did not produce a real `onMultiWindowModeChanged` transition on this emulator, so it is not cleanly scriptable. PLO emitted no spurious OPEN, but the transition likely did not truly fire, so I am calling this inconclusive rather than proven.

**Net:** the one hard, reproducible differentiator remains the guard's stale-flag OPEN-suppression bug (deterministic unit test), which PLO avoids by construction. The debounce / multi-window / PIP advantages are real architecturally but need a multi-activity + real multi-window instrumented harness to demonstrate — the SDK androidTest set does not currently compile (canClearInitData / EMT-3875), and Robolectric cannot drive ProcessLifecycleOwner (needs the manifest InitializationProvider, stripped by Config.NONE). Recommend that harness as the follow-up if we want those proven.

wpinho-branch:
## Correction + on-device proof of the 700ms debounce

My earlier note said the debounce was "not cleanly reproducible" — that was a **method** problem (open-loop adb timing), not reality. Proved it deterministically with an **instrumented test** driving the real ProcessLifecycleOwner, timing controlled in-process, backgrounding via UiAutomation HOME:

```
BranchProcessLifecycleDebounceInstrumentedTest:
  quickReturn_underDebounce_emitsNoExtraOpen  (300ms background)  -> 1 ON_START  PASS
  slowReturn_pastDebounce_emitsExtraOpen      (1500ms background) -> 2 ON_STARTs PASS
OK (2 tests)   [ran twice, stable, ~7.9s each]
```

It counts ProcessLifecycleOwner ON_START dispatches (which this observer forwards to `sendOpen()` 1:1). Because the two cases give **distinct** values (1 vs 2), the test is discriminating, not tautological. So: a quick app-switch (< 700ms) emits **no** extra OPEN; a real background (> 700ms) does. The activity-count guard has no such debounce (emits on every stop -> start).

### Why it could only run locally with hacks (and how CI fixes that)
The test is NOT in this PR — it needs three things the beta branch lacks today:
1. **EMT-3875**: the `androidTest` source set does not compile (`canClearInitData`, `setInitState`). I stubbed those two pre-existing lines locally to run; that must land properly first.
2. **A `targetSdk`**: `defaultConfig` sets `minSdk` but no `targetSdk`, so the test APK targets 21 → API 34 shows the legacy permission-review screen that intercepts ActivityScenario, and blocks install. Setting `targetSdk = 34` fixes both (worth doing in the SDK config regardless).
3. A clean emulator — locally the dev emulator ran out of binder buffer (GSA/YouTube background apps); a headless CI emulator with those disabled is stable.

### CI recipe (once EMT-3875 lands)
The repo already runs instrumented tests on push. Add this test and run it via the standard emulator job:
```yaml
- uses: reactivecircus/android-emulator-runner@v2
  with:
    api-level: 34
    target: google_apis
    arch: arm64-v8a   # or x86_64 on Intel runners
    emulator-options: -no-window -gpu swiftshader_indirect -no-snapshot -noaudio -no-boot-anim
    disable-animations: true
    script: ./gradlew :Branch-SDK:connectedDebugAndroidTest
```
CI emulators are dedicated VMs without the background-app binder pressure that killed the local run, so it runs deterministically there.

Branch-SDK/src/main/java/io/branch/observers/BranchProcessLifecycleObserver.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+internal class BranchProcessLifecycleObserver(private val branchInstance: Branch) : DefaultLifecycleObserver {
+
+    override fun onStart(owner: LifecycleOwner) {
+        BranchLogger.v("BranchProcessLifecycleObserver onStart: process foregrounded, sending OPEN")
+        branchInstance.sendOpen()
+    }
+
+    override fun onStop(owner: LifecycleOwner) {
+        BranchLogger.v("BranchProcessLifecycleObserver onStop: process backgrounded")
```
gdeluna-branch, line 26: Let's add a call for `branchInstance.closeSessionInternal` so that previous session parameters are not incorrectly referenced.

Branch-SDK/build.gradle.kts
```diff
@@ -20,6 +20,8 @@ dependencies {
     implementation("androidx.annotation:annotation:1.4.0")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutinesVersion")
     implementation("androidx.constraintlayout:constraintlayout:2.1.4")
+    // App foreground/background detection at the process level (SDK-2463): ProcessLifecycleOwner.
+    implementation("androidx.lifecycle:lifecycle-process:2.5.1")
```
gdeluna-branch: What's the lowest version we should/can support?

wpinho-branch:
Confirmed. The observer was not being removed from ProcessLifecycleOwner on shutDown(), causing Branch instances to leak in test cycles.

Fixed by adding `unregister()` method to BranchProcessLifecycleObserver companion object and calling it from `Branch.shutDown()`. The cleanup is now symmetric with registration.

Additionally hardened the implementation for production:
- Added `@Volatile` to the static instance field for cross-thread visibility
- Made register/unregister execute synchronously when already on main thread (eliminates unnecessary posting overhead for standard Application.onCreate init path)
- Added `shutDownForTesting()` method that blocks with CountDownLatch to ensure deterministic cleanup in test cycles, preventing static pollution between tests
- Branch.shutDown() now detects test mode and uses the blocking variant automatically

Verified with new test suite (BranchProcessLifecycleObserverStaticIsolationTest) that validates multiple init/shutdown cycles without memory leaks.

wpinho-branch:
The `onStop()` method doesn't call `closeSessionInternal` because BranchActivityLifecycleObserver (separate, existing observer) handles session close when activities stop.

This PR only changes foreground detection for OPEN events. The Process observer handles OPEN; the Activity observer handles close. Added a comment in `onStop()` documenting this architecture.

wpinho-branch:
The architecture uses two observers in parallel:
- **ProcessLifecycleObserver** (this new one) → handles OPEN when app enters foreground
- **BranchActivityLifecycleObserver** (existing) → still handles close on activity stop

So `closeSessionInternal` is already covered by the Activity observer. This PR only changes how we detect app foreground for the OPEN event - the session close logic stays where it was.

Added a comment in `onStop()` to document this since it's not obvious at first glance.

Branch-SDK/src/main/java/io/branch/observers/BranchProcessLifecycleObserver.kt
wpinho-branch: Fixed in f1797a9. The comment was referencing `BranchActivityLifecycleObserver`, which does not exist in the codebase. Corrected to accurately describe the behavior: session close on background is an intentional no-op here, matching what the old `BranchOpenObserver` also did (it had no `closeSessionInternal` call on `onActivityStopped` either). Comment now reads: *"Session close on background is intentionally not triggered here. The pre-existing BranchOpenObserver also did not call closeSessionInternal on background — no regression. When the beta session model is finalized, session-close logic belongs here."*

Branch-SDK/build.gradle.kts
wpinho-branch: Lowered to `2.4.1` in f1797a9. `DefaultLifecycleObserver` is available since `2.4.0`, and `2.4.1` is compatible with our Kotlin 1.6.21 without forcing a higher stdlib floor on SDK consumers. `2.5.1` pulls in `kotlin-stdlib:1.7.10` transitively — a higher floor than our own toolchain, which is a bad precedent for a library SDK. `2.4.1` is the right floor here.

PR #1371, opened 2026-07-08
EMT-3899 -- Harden synchronous referring-params latch release (volatile + TOCTOU)

wpinho-branch:
## EMT-3899 — Harden synchronous referring-params latch release

Follow-up from EMT-3882 / #1365 (merged). Fixes two issues in the synchronous getter latch release flagged by the automated review.

### Changes
- **TOCTOU NPE fix** (`ServerRequestInitSession.onInitSessionCompleted`): capture each latch in a local variable before the null-check, so the check and `countDown()` act on the same reference. Prevents an NPE if a timing-out `*Sync` caller nulls the field between the check and the release.
- **Visibility fix** (`Branch.java`): mark `getFirstReferringParamsLatch` / `getLatestReferringParamsLatch` `volatile`, so a release on the init thread is seen by the awaiting sync getter (releases early instead of waiting the full 2.5s `LATCH_WAIT_UNTIL`).
- **Test** (`LatchReleaseConcurrencyTest`): release-on-init and timeout-fallback on a real `Branch` instance (Robolectric). The existing suite mocks `Branch`, so this path had no coverage.

### Verification
- `./gradlew :Branch-SDK:testDebugUnitTest` → 415 passed, 0 failed (new test 2/2).
- `./gradlew :Branch-SDK:assembleDebug` → BUILD SUCCESSFUL.

### Acceptance criteria
- [x] Latch release captures a local reference before dereferencing.
- [x] Latch fields are volatile.
- [x] Test covers release-on-init and timeout-fallback on a real instance.

Jira: EMT-3899


Branch-SDK/src/test/java/io/branch/referral/LatchReleaseConcurrencyTest.kt
```diff
@@ -0,0 +1,106 @@
...
+ *    concurrent timeout+null-assign on the *Sync side cannot interleave between the check and
+ *    countDown().
+ */
+class LatchReleaseConcurrencyTest : BranchTestBase() {
+
+    private lateinit var branch: Branch
+
+    @Before
+    fun setUp() {
+        super.setUpBase()
```
gdeluna-branch: BranchTestBase.setUpBase() is itself `@Before`, so JUnit 4 already calls it. This explicit super call runs it twice. You can remove this.

Branch-SDK/src/test/java/io/branch/referral/LatchReleaseConcurrencyTest.kt
```diff
@@ -0,0 +1,106 @@
...
+        while (branch.getLatestReferringParamsLatch == null && System.currentTimeMillis() < deadline) {
+            Thread.sleep(10)
+        }
+
+        val latch = branch.getLatestReferringParamsLatch
+        assertNotNull("latch should be non-null once getLatestReferringParamsSync is in-flight", latch)
+
+        // Simulate what onInitSessionCompleted does after the EMT-3899 TOCTOU fix:
+        // capture a local reference first, then count down.
+        latch!!.countDown()
```
gdeluna-branch, line 78: latch!!.countDown() throws NPE (not assertion failure) if the poll on lines 67–70 times out. The assertNotNull on
  line 73 doesn't protect it because latch is re-read from the field on line 72. Capture once and assert before using: 
val latch = branch.getLatestReferringParamsLatch; 
assertNotNull(..., latch); 
latch!!.countDown()

Branch-SDK/src/test/java/io/branch/referral/LatchReleaseConcurrencyTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    fun getFirstReferringParamsSync_returns_after_timeout_without_init() {
+        // PrefHelper returns NO_STRING_VALUE ("bnc_no_value") for installParams on a fresh context,
+        // so getFirstReferringParamsSync will enter the await block and block for LATCH_WAIT_UNTIL.
+        val future = CompletableFuture<org.json.JSONObject>()
+        Thread {
+            future.complete(branch.getFirstReferringParamsSync())
+        }.start()
+
+        // Allow LATCH_WAIT_UNTIL (2500 ms) plus a generous 1500 ms buffer for slow CI machines.
+        val result = future.get(4000L, TimeUnit.MILLISECONDS)
```
gdeluna-branch, line 104: Slow test (unconditional 2.5 s) with no CI marker. Add a `@Category(SlowTest::class)` or equivalent so it can be excluded from the default test run.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -1032,8 +1032,8 @@ public void logout(LogoutStatusListener callback) {
 
     private static final int LATCH_WAIT_UNTIL = 2500; // ms; the *Sync getters give up waiting after this.
 
-    CountDownLatch getFirstReferringParamsLatch = null;
-    CountDownLatch getLatestReferringParamsLatch = null;
+    volatile CountDownLatch getFirstReferringParamsLatch = null;
+    volatile CountDownLatch getLatestReferringParamsLatch = null;
```
gdeluna-branch: Implicit package-private is inconsistent with private static final two lines above. Add /* package */ or make them private with a package-visible accessor.

Branch-SDK/src/test/java/io/branch/referral/LatchReleaseConcurrencyTest.kt
wpinho-branch: Good catch, removed. Worth knowing it's not just here: 7 test classes call super.setUpBase() the same way, so the base setup runs twice across most of the suite. I kept this PR to its own file rather than touching the other six.

Branch-SDK/src/test/java/io/branch/referral/LatchReleaseConcurrencyTest.kt
wpinho-branch: Added. There was no category in the repo, so this PR introduces `SlowTest` (Branch-SDK/src/test/java/io/branch/referral/SlowTest.kt) and marks the timeout test with it.

I wired it as opt-in rather than excluded by default: `-PexcludeSlowTests` skips it, a normal run still covers it. Excluding by default would have quietly dropped the only test of the real timeout path. Verified: 438 tests by default, 437 with the flag, and the class goes from 2.586s to 0.072s.

Happy to flip it to excluded-by-default if you'd rather.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
wpinho-branch: Done, both fields are now `/* package */ volatile`. They stay package-private because LatchReleaseConcurrencyTest reads them from the same package, so the marker just makes that intent explicit.

Branch-SDK/src/test/java/io/branch/referral/LatchReleaseConcurrencyTest.kt
wpinho-branch: I think this one's already covered, unless I'm misreading you.

The field is read once into a local on line 72, and line 77 uses that local:

```kotlin
val latch = branch.getLatestReferringParamsLatch   // 72
assertNotNull("latch should be non-null ...", latch)  // 73
latch!!.countDown()                                // 77
```

That's the shape you suggested. If the poll times out, `latch` is null and line 73 throws AssertionError, so line 77 never runs — the failure is an assertion, not an NPE.

Left it as is. If you were pointing at something else, say the word and I'll fix it.

gdeluna-branch (approved):

PR #1375, opened 2026-07-15
EMT-3881: restore BranchUniversalObject.showShareSheet() delegating to native share

wpinho-branch:
## EMT-3881: restore BranchUniversalObject.showShareSheet()

Restores the public `showShareSheet()` overload that 5.x exposed on `BranchUniversalObject`, which was removed in the 6.0 beta and left integrations that share from a BUO with no way to show the share sheet.

### Approach

Per product direction (restore for API continuity), the overload is added back but delegates to the native share, `Branch.share(Activity, BranchUniversalObject, LinkProperties, BranchNativeLinkShareListener, String, String)`, rather than resurrecting the removed in-SDK share dialog.

```java
@Deprecated
public void showShareSheet(@NonNull Activity activity, @NonNull LinkProperties linkProperties,
                           @NonNull ShareSheetStyle style, @Nullable Branch.BranchLinkShareListener callback)
```

- `ShareSheetStyle.getMessageTitle()` becomes the chooser title (`EXTRA_TITLE`), `getMessageBody()` the subject (`EXTRA_SUBJECT`).
- `BranchLinkShareListener` is adapted to `BranchNativeLinkShareListener`; the app reported to `onChannelSelected` is carried into `onLinkShareResponse` as the channel.
- Below API 22, where `Branch.share` is unavailable, the callback receives `ERR_BRANCH_NO_SHARE_OPTION` instead of the SDK crashing.
- Marked `@Deprecated` pointing to `Branch.share(...)`.
- `IChannelProperties` was removed in the beta, so only the 4-arg overload is restored.

### Migration notes from 5.x

- `onShareLinkDialogLaunched()` and `onShareLinkDialogDismissed()` do not fire; the OS share sheet exposes no such events.
- `ShareSheetStyle` styling for the old custom dialog (copy-url icon, more-option icon, fonts) is ignored; only the message title and body are used.
- The channel passed to `onChannelSelected` and `onLinkShareResponse` is the chosen app's flattened `ComponentName`, not a display label such as `WhatsApp`, and is null if the share fails before an app is chosen.
- The `~channel` link parameter is not taken from the user's choice, because the link is built before the sheet opens; set it up front with `LinkProperties.setChannel()`.
- The share sheet requires API 22.

### Verification

- `./gradlew :Branch-SDK:testDebugUnitTest` passes. `NativeShareListenerAdapterTest` covers the channel handoff; `ShowShareSheetApiCompatTest` fails if the overload is removed again, since nothing in the SDK, TestBed or androidTest calls it.
- `./gradlew :Branch-SDK:lintDebug`: no NewApi finding on the `Branch.share` call.

### Changes after review

- Javadoc cut to a summary and the parameters; the migration notes moved here.
- Compat test trimmed to the overload check.
- API 22 guard added.

### Out of scope

- Lint still reports a pre-existing NewApi in `ApiUsageAnalytics.kt` (`ConcurrentHashMap.newKeySet`, API 24).

### DIFF
--- a/Branch-SDK/src/main/java/io/branch/indexing/BranchUniversalObject.java
+++ b/Branch-SDK/src/main/java/io/branch/indexing/BranchUniversalObject.java
@@ -1,5 +1,6 @@
 package io.branch.indexing;
 
+import android.app.Activity;
 import android.content.Context;
 import android.os.Parcel;
 import android.os.Parcelable;
@@ -18,6 +19,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 
 import io.branch.referral.Branch;
+import io.branch.referral.BranchError;
 import io.branch.referral.BranchLogger;
 import io.branch.referral.BranchShortLinkBuilder;
 import io.branch.referral.BranchUtil;
@@ -27,6 +29,7 @@ import io.branch.referral.util.BranchEvent;
 import io.branch.referral.util.ContentMetadata;
 import io.branch.referral.util.CurrencyType;
 import io.branch.referral.util.LinkProperties;
+import io.branch.referral.util.ShareSheetStyle;
 
 /**
  * <p>Class represents a single piece of content within your app, as well as any associated metadata.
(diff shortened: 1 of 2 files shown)


Branch-SDK/src/main/java/io/branch/indexing/BranchUniversalObject.java
```diff
@@ -374,8 +377,81 @@ public void generateShortUrl(@NonNull Context context, @NonNull LinkProperties l
     
     //------------------ Share sheet -------------------------------------//
 
+    /**
+     * <p>Shows a share sheet for this {@link BranchUniversalObject} so users can share a Branch link
+     * to other applications.</p>
+     *
+     * <p>Restored for source compatibility with 5.x integrations that shared directly from a
```
gdeluna-branch, line 384: These comments are very verbose and not valuable when displayed as part of the public api. Please minimize commentary, only an explanation of the function and parameters are needed.

Branch-SDK/src/test/java/io/branch/referral/ShowShareSheetApiCompatTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.robolectric.RobolectricTestRunner;
+
+import java.lang.reflect.Method;
+
+import io.branch.indexing.BranchUniversalObject;
+import io.branch.referral.util.LinkProperties;
+import io.branch.referral.util.ShareSheetStyle;
+
+/**
+ * EMT-3881: the public BranchUniversalObject.showShareSheet() overload that 5.x exposed was removed
```
gdeluna-branch, line 19: Can you explain how this test will help? Won't existing tests validate the same thing?

Branch-SDK/src/main/java/io/branch/indexing/BranchUniversalObject.java
wpinho-branch: Condensed to a one-line summary and the parameters; the 5.x migration notes moved to the PR description.

Branch-SDK/src/test/java/io/branch/referral/ShowShareSheetApiCompatTest.java
wpinho-branch: Nothing in the SDK, TestBed, or androidTest calls showShareSheet, so removing it again would leave every build and test green except this one. I trimmed it to that single check and dropped the other two assertions.

wpinho-branch:
Added an API 22 guard; below it the callback now receives ERR_BRANCH_NO_SHARE_OPTION instead of the SDK crashing.

PR #1377, opened 2026-07-17
update v2/event to v3/events

gdeluna-branch:
## Reference
SDK-XXX -- <TITLE>.

## Summary

  - Migrates standard and custom event endpoints from `v2/event/standard|custom` → `v3/events/standard|custom` in `Defines.RequestPath`
  - Consolidates the repeated V1/V2 body-routing conditional duplicated across 6+ mutation methods on `ServerRequest` into a single private
  `deviceDataParamsObject()` helper — returns `params_` for V1 (flat body) or the `user_data` sub-object for V2 (nested body)
  - First step toward extracting body-decoration logic out of `ServerRequest` entirely

  ## Changes

  - **`Defines.java`** — update `TrackStandardEvent` and `TrackCustomEvent` path strings
  - **`ServerRequest.java`** — add `deviceDataParamsObject()`; wire it into `addDMAParams`, `addConsumerProtectionAttributionLevel`,
  `updateLimitFacebookTracking`, `updateDisableAdNetworkCallouts`, `updateGAdsParams`, `updateDeviceInfo`, `updateEnvironment`
  - **`DeviceDataParamsRoutingTest.kt`** — 16 Robolectric tests covering V1/V2 routing and no-op guards for each affected method
  - **Testbed** — add "Send Custom Event" button mirroring the existing standard event buttons

  ## Test plan

  - [ ] All 16 `DeviceDataParamsRoutingTest` cases pass
  - [ ] Build and run testbed; confirm standard event, custom event, and commerce event all reach `v3/events/*` in logcat
  - [ ] Confirm V1 requests (install, open) still write device fields at the top level of the body
  - [ ] Confirm V2 requests (log event) still write device fields inside `user_data`

- Correctness & Style


Branch-SDK/src/main/java/io/branch/referral/Defines.java
```diff
@@ -267,8 +267,8 @@ public enum RequestPath {
         RegisterInstall("v1/install"),
         RegisterOpen("v1/open"),
         ContentEvent("v1/content-events"),
-        TrackStandardEvent("v2/event/standard"),
-        TrackCustomEvent("v2/event/custom"),
+        TrackStandardEvent("v3/events/standard"),
+        TrackCustomEvent("v3/events/custom"),
```
wpinho-branch, line 271: This moves every standard and custom event onto v3, and it lands in a beta line. Neither the PR nor the ticket says anything about whether v3 is live for all keys, whether the response shape matches what the SDK parses today, or how this gets rolled back if it isn't.

What's the backend story here? If v3 is already serving, saying so in the description is enough. If it's gated, this probably needs to wait or sit behind something the SDK can flip.

Branch-SDK/src/main/java/io/branch/referral/ServerRequest.java
```diff
@@ -476,38 +462,24 @@ void updateGAdsParams() {
...
+                Defines.Jsonkey latKey = isV1 ? Defines.Jsonkey.LATVal : Defines.Jsonkey.LimitedAdTracking;
+                Defines.Jsonkey adIdKey = isV1 ? Defines.Jsonkey.GoogleAdvertisingID : Defines.Jsonkey.AAID;
+                userDataObj.put(latKey.getKey(), LATVal);
                 if (!TextUtils.isEmpty(gaid)) {
                     if (!SystemObserver.isHuaweiMobileServicesAvailable(context_)) {
                         // Fire OS overloads ad id (representing it as Google ad id at the top level),
                         // HUAWEI only reports ad id in the advertising_ids object
                         if (prefHelper_.getConsumerProtectionAttributionLevel() == Defines.BranchAttributionLevel.FULL || !prefHelper_.isAttributionLevelInitialized()) {
-                            params_.put(Defines.Jsonkey.GoogleAdvertisingID.getKey(), gaid);
+                            userDataObj.put(adIdKey.getKey(), gaid);
```
wpinho-branch, line 476: I traced this one carefully since it's the most sensitive method in the diff, and the consumer-protection guard is intact: the ad ID still only gets written when the level is FULL or the level was never initialized. No regression from the V1/V2 merge.

The gap is on the test side. Both new cases for this method mock a FULL attribution level, so the path that matters most, the one where the ad ID must not be written, never runs. A third case with the level set to something below FULL, asserting the key is absent, would lock that in.

wpinho-branch (commented):
The consolidation into a single body-routing helper is a real improvement over the conditional repeated across those methods, and I like that it comes with tests.

Two things about how this is packaged rather than the code itself. Only the code review check has run here, so there's been no build and no unit run, which means the new routing tests have never executed in CI. A trivial commit would trigger the workflows.

And the ticket is the two path strings in Defines. The helper extraction across seven mutation methods is the larger and riskier half of the diff, riding on a one-line ticket. If those were separate PRs the endpoint change could be reviewed and reverted on its own, which matters more than usual given where it points.

I read the diff carefully but didn't build this branch.

gdeluna-branch:
Will re-open this when other code refactorings are done.

PR #1381, opened 2026-07-21
BranchConfiguration Builder

gdeluna-branch:
## Reference
EMT-3912 -- [Android][Beta] Configuration Builder Class

<html><head></head><body><p>Now removing the manifest and string-resource tiers, keeping branch.json.</p>
<p>Now removing the orphaned constants and the stale <code>enableTestMode</code> javadoc link.</p>
<p>Two leftovers. Let me check the validator one before deciding.</p>
<p><code>BranchIntegrationModel</code> reads the manifest keys for a dashboard-comparison diagnostic, not as a config source — separate concern, leaving it. Let me pull the exact mapping for the description.</p>
<p>Done. Two clean full-suite runs, TestBed assembles, instrumented sources compile.</p>
<p><strong>What changed for <code>readBranchKey</code>:</strong> it now resolves from <code>branch.json</code> only. <code>readBranchKeyFromManifest</code>, <code>readBranchKeyFromMetaData</code>, <code>handleTestModeFallback</code>, <code>readBranchKeyFromStringResources</code>, and the <code>BRANCH_KEY_LIVE</code>/<code>BRANCH_KEY_TEST</code>/<code>SOURCE_MANIFEST</code>/<code>SOURCE_STRINGS</code> constants are deleted — 107 lines out of <code>BranchUtil</code>.</p>
<p>Its three callers, per the audit:</p>
<ul>
<li><code>applyBranchKey</code> — dropped the vestigial comparison; key source is now unconditionally <code>"init_function"</code></li>
<li><code>BranchConfigurationManager.validateConfiguration</code> — deleted, zero callers</li>
<li><code>BranchKeysValidatorCheck</code> — now reads the resolved key from <code>PrefHelper</code> instead of the manifest, fixing the false failure for builder-supplied keys</li>
</ul>
<p><code>BranchIntegrationModel</code> still reads the manifest keys, but only to compare local setup against the dashboard in the Integration Validator — a diagnostic, not a config source. Left alone; flagging in case you want it in the branch.json PR.</p>
<hr>
<h2>Description</h2>
<p>Replaces the scattered pre-init setters on <code>Branch</code> with a single immutable configuration object and one initialization entry point.</p>
<h3>Added</h3>
<ul>
<li><strong><code>BranchConfiguration</code></strong> + nested <code>Builder</code> — all pre-init settings, validated at <code>build()</code>, immutable afterward.</li>
<li><strong><code>Branch.initialize(Context, BranchConfiguration)</code></strong> — the sole init entry point.</li>
<li><strong><code>DMAParameters</code></strong> — named fields replacing three positional booleans.</li>
<li><strong><code>BranchLogLevel.NONE</code></strong> — fully silences SDK output, including <code>logAlways</code>.</li>
</ul>
<h3>Setting mappings</h3>

Removed from Branch | Builder equivalent
-- | --
enableTestMode() / disableTestMode() | setTestMode(Boolean)
setAPIUrl(String) | setApiUrl(String)
setCDNBaseUrl(String) | setCdnBaseUrl(String)
useEUEndpoint() | setEUEndpoint(Boolean)
setBranchRemoteInterface(...) | setRemoteInterface(...)
enableLogging() / enableLogging(level) | setLogLevel(BranchLogLevel)
enableLogging(callback) | setLoggingCallback(...)
disableLogging() | setLogLevel(NONE)
setCallbackForTracingRequests(...) | setRequestTracingCallback(...)
setNetworkTimeout(int) | setNetworkTimeout(Int)
setNetworkConnectTimeout(int) | setNetworkConnectTimeout(Int)
setRetryCount(int) | setRetryCount(Int)
setRetryInterval(int) | setRetryInterval(Int)
setNoConnectionRetryMax(int) | setNoConnectionRetryMax(Int)
setDMAParamsForEEA(b, b, b) | setDMAParameters(DMAParameters)
setLimitFacebookTracking(boolean) | setLimitFacebookAttribution(Boolean)
disableAdNetworkCallouts(boolean) | setAdNetworkCalloutsDisabled(Boolean)
setFBAppID(String) | setFacebookAppId(String)
addInstallMetadata(k, v) | addInstallMetadata(String, String)
setPreinstallCampaign(String) | setPreinstallCampaign(String)
setPreinstallPartner(String) | setPreinstallPartner(String)
setReferringLinkAttributionForPreinstalledAppsEnabled() | setReferringLinkAttributionForPreinstalledApps(Boolean)
addWhiteListedScheme(String) | addWhitelistedScheme(String)
setWhiteListedSchemes(List) | addWhitelistedScheme(String) per scheme
addUriHostsToSkip(String) | addUriHostToSkip(String)
setIsUserAgentSync(boolean) | setUserAgentFetchSync(Boolean)
getAutoInstance(Context) | Branch.initialize(Context, config)

<h3>Behavior changes</h3>
<ul>
<li><code>shutDown()</code> now clears the request-tracing callback, which previously leaked across re-initialization.</li>
<li><code>setAutomaticOpenEvents(false)</code> no longer registers the lifecycle observer at all. It previously registered then immediately unregistered, emitting an OPEN when the process was already foregrounded.</li>
</ul>
<h3>Tests</h3>
<p>94 across four files — <code>BranchConfigurationBuilderTest</code> (51), <code>BranchInitializeTest</code> (31), <code>DMAParametersTest</code> (5), <code>TraceLoggingTest</code> (7).</p></body></html>

- Correctness & Style


Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -392,38 +392,117 @@ synchronized private static Branch getAutoInstance(@NonNull Context context, Str
...
+            BranchLogger.w("Warning, attempted to reinitialize Branch SDK singleton!");
+            return;
+        }
+        branchReferral_ = new Branch(context.getApplicationContext());
+        applyBranchKey(context, branchReferral_, config.getBranchKey());
+
+        // Caller values applied first so branch.json/manifest serve only as fallbacks.
+        applyConfiguration(branchReferral_, config);
 
         BranchConfigurationManager.loadConfiguration(context, branchReferral_);
```
wpinho-branch: The caller's config is applied here and branch.json is loaded right after, so branch.json wins on anything that doesn't defend itself. The comment says the opposite.

I checked this by injecting a branch.json in a test and reading back what survived `initialize()`. A caller-supplied `apiUrl` loses to the one in the file, `fbAppId` loses, `setTestMode(false)` gets flipped back to true, and an attribution level of MINIMAL comes back as FULL.

That last one is why I'd fix this before it ships. An app that deliberately asks for minimal attribution ends up on full without anything telling it.

`setTestMode(true)` is the only value that survives, and only by accident: `checkTestMode` bails out early when the flag is already on, so it never reaches the file. `setAPIBaseUrlFromConfig`, `setFbAppIdFromConfig` and `setCPPLevelFromConfig` have no equivalent, so they just overwrite whatever `applyConfiguration` had set a moment earlier.

Two ways out that I can see: load branch.json before applying the caller's config, or give those three helpers the same already-set check that `checkTestMode` has. I didn't test either, so I don't want to hand you a diff blind.

Worth knowing that iOS #1614 landed on caller-wins for the same feature and wrote the reasoning into its `initialize:`, so whichever way this goes it's worth the two matching.

wpinho-branch (changes requested):
The Builder reads well and the coverage on the network settings is solid. What blocks is the ordering inside `initialize()`: branch.json is loaded after the caller's config, so it wins wherever there's nothing guarding against it, and that includes the attribution level. Detail is in the inline comment.

Small one: the reference section still has the `SDK-XXX -- <TITLE>` placeholder while the branch name carries the ticket.

I ran the existing `BranchInitializeTest` on this commit and it passes, plus a throwaway probe of my own for the ordering question. I didn't run the instrumented tests.

Branch-SDK/src/main/java/io/branch/referral/BranchConfigurationManager.java
```diff
@@ -65,7 +65,9 @@ public static void loadConfiguration(@NonNull Context context, @NonNull Branch b
      */
     private static void loadLoggingConfiguration(@NonNull Context context) {
         if (BranchUtil.getEnableLoggingConfig(context)) {
-            Branch.enableLogging();
+            BranchLogger.setLoggerCallback(null);
```
bboothe-branch, line 67: Following up on Willian's comment, it's 6 values, not 4. `loadLoggingConfiguration` also sets the logger callback to `null` and forces DEBUG, so a caller's `setLoggingCallback()` and `setLogLevel()` are both lost
   when branch.json sets `enableLogging`.

Worth noting the SDK already has the pattern this needs, in two places. `BranchPreinstall` only writes campaign/partner when the existing value is empty, and checkTestMode bails early when the flag is already on. That only protects **true** though. A caller-supplied **false** still falls through to the file. `setAPIBaseUrlFromConfig`, `setFbAppIdFromConfig`, `setCPPLevelFromConfig` and `loadLoggingConfiguration` are the ones that don't. 

We should decide to either give them the same type of check, or load branch.json before `config.applyTo` so precedence comes from the ordering instead of four separate guards.

Branch-SDK/src/main/java/io/branch/referral/BranchConfigurationManager.java
bboothe-branch, line 29: All seven loaders share one try/catch, and`Defines.BranchAttributionLevel.valueOf()` throws on an unrecognized string. So a typo like "MINMAL" in branch.json silently skips every loader after it including pre-install data.

No test in BranchInitializeTest.kt constructs a branch.json, so none of this is reachable by the suite. `initialize_testModeFalse_doesNotEnableTestMode` will fail as soon as one is added.

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
```diff
@@ -0,0 +1,223 @@
...
+        requestTracingCallback?.let { Branch._iBranchRequestTracingCallback = it }
+
+        // Identity & environment
+        BranchUtil.setTestMode(testMode)
+        apiUrl?.let { PrefHelper.setAPIUrl(it) }
+        cdnBaseUrl?.let { PrefHelper.setCDNBaseUrl(it) }
+        if (euEndpoint) PrefHelper.useEUEndpoint(true)
+
+        // Network
+        if (networkTimeout > 0) prefHelper.timeout = networkTimeout
```
bboothe-branch: `build()` validates networkTimeout, networkConnectTimeout and retryCount, which makes the matching guards in `applyTo` redundant. Meanwhile retryInterval and noConnectionRetryMax are guarded in `applyTo` but never validated in `build()`. I'd suggest consolidating in `build()` to keep the validation together.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -392,39 +392,62 @@ synchronized private static Branch getAutoInstance(@NonNull Context context, Str
...
+        branchReferral_ = new Branch(context.getApplicationContext());
+        applyBranchKey(context, branchReferral_, config.getBranchKey());
+
+        // Caller values applied first so branch.json/manifest serve only as fallbacks.
+        config.applyTo(branchReferral_);
 
         BranchConfigurationManager.loadConfiguration(context, branchReferral_);
 
-        // ProcessLifecycleOwner tracks the whole process, so no Application reference is required.
         branchReferral_.setupProcessLifecycleObserver();
```
bboothe-branch: When `automaticOpenEvents` is false, `initialize` calls `setupProcessLifecycleObserver()` then `unregister()` which seems unnecessary as it sets it up just to unregister it right afterwards. If this is intentional please ignore.

bboothe-branch (changes requested):
Hey @gdeluna-branch overall this looks great. Love the structure and design. Couple things I'd suggest before shipping.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
gdeluna-branch: Let's consider branch.json behavior out of scope for now as unsupported for native integrations. We will revisit for plugins.

Branch-SDK/src/main/java/io/branch/referral/BranchConfigurationManager.java
gdeluna-branch: enableLogging() already called enableLogging(null, DEBUG), it's inlined here since I removed the static. branch.json is out of scope for now.

Branch-SDK/src/main/java/io/branch/referral/BranchConfigurationManager.java
gdeluna-branch: This is the branch.json config manager, out of scope.

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
gdeluna-branch: Fixed.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
gdeluna-branch: Fixed.

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
```diff
@@ -0,0 +1,342 @@
...
+    val referringLinkAttributionForPreinstalledApps: Boolean,
+    val whitelistedSchemes: List<String>,
+    val uriHostsToSkip: List<String>,
+    val automaticOpenEvents: Boolean,
+    val userAgentFetchSync: Boolean
+) {
+
+    /**
+     * Writes every configured value through to the subsystem that owns it. Called once by
+     * [Branch.initialize]; this is the only place pre-init state is applied.
```
bboothe-branch:
```suggestion
     * [Branch.initialize];
```

Can delete "this is the only place pre-init state is applied." This is incorrect as branchKey goes through applyBranchKey, automaticOpenEvents is read at Branch.java, and BranchConfigurationManager.loadConfiguration runs at line 392. We can delete the words "this is the only place pre-init state is applied" from that comment.

Would also move it down to just above @JvmName("applyTo") so it documents right above that instead of also documenting for applyLogging.

Branch-SDK/src/test/java/io/branch/referral/BranchInitializeTest.kt
```diff
@@ -0,0 +1,515 @@
...
+
+    /**
+     * applyTo() installs the config's own logging callback, so a capture callback has to be
+     * supplied through the builder to see what initialize() emits.
+     */
+    private fun captureInitLogs(
+        level: BranchLogger.BranchLogLevel = BranchLogger.BranchLogLevel.DEBUG,
+        configure: BranchConfiguration.Builder.() -> Unit = {}
+    ): List<String> {
+        val captured = mutableListOf<String>()
```
bboothe-branch: captureInitLogs can run into a race condition. BranchInitializeTest.kt returns a mutable list that logging callback appends to. Init's open request logs requests logs from background coroutine, which can be a problem if the list is still being added to leading to a crash. Fix by implementing CopyOnWriteArrayList to make iteration safe by capturing snapshots.

`val captured = CopyOnWriteArrayList<String>()`

`return captured.toList()`

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -357,184 +357,132 @@ synchronized public static Branch getInstance() {
...
+     * {@code branch_configuration_applied} line above it was overridden after the configuration was
+     * applied.
      */
-    public void setBranchRemoteInterface(BranchRemoteInterface remoteInterface) {
-        if (remoteInterface == null) {
-            branchRemoteInterface_ = new BranchRemoteInterfaceUrlConnection(this);
-        } else {
-            branchRemoteInterface_ = remoteInterface;
+    private static void logInitializeComplete(@NonNull Branch branch, @NonNull BranchConfiguration config) {
+        if (!BranchLogger.getLoggingEnabled()) {
```
bboothe-branch:
```suggestion
        if (!BranchLogger.isLoggable(BranchLogger.BranchLogLevel.DEBUG))) {
```

logInitializeComplete has a dead guard. it early returns on BranchLogger.getLoggingEnabled() but applyLogging() unconditionally sets it true. if (BranchLogger.getLoggingLevel() == BranchLogLevel.NONE) return should be how it's setup so it doesn't build the whole JSON string just to discard it after.

Can set up using the fix below in BranchLogger

`@JvmStatic`
`fun isLoggable(level: BranchLogLevel): Boolean = loggingEnabled && shouldLog(level)`

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
```diff
@@ -0,0 +1,342 @@
...
+        BranchLogger.loggingEnabled = true
+    }
+
+    @JvmName("applyTo")
+    internal fun applyTo(branch: Branch) {
+        val context = branch.applicationContext
+        val prefHelper = branch.prefHelper
+
+        BranchLogger.logAlways(Branch.GOOGLE_VERSION_TAG)
+        BranchLogger.d(toJson())
```
bboothe-branch:
```suggestion
       if (BranchLogger.isLoggable(BranchLogger.BranchLogLevel.DEBUG)) BranchLogger.d(toJson())
```

Same as logInitializeComplete. Builds JSON just to potentially discard. Can set up to call it using the below code.

`@JvmStatic`
`fun isLoggable(level: BranchLogLevel): Boolean = loggingEnabled && shouldLog(level)`

bboothe-branch (changes requested):
Tested and all previous updates were fixed. Found some small areas that could use an update.

Branch-SDK/src/main/java/io/branch/referral/BranchConfigurationManager.java
wpinho-branch: This still reads the same way on the current head:

https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/113acdeee773d8b2afb5cb167c097a8b67b260b2/Branch-SDK/src/main/java/io/branch/referral/Branch.java#L390-L392

`applyTo` runs first, then `loadConfiguration` lands on top of it, and `loadLoggingConfiguration` clears the logger callback before raising the level.

It reproduces on device. With the builder set to one `apiUrl` and branch.json set to another, requests went to the branch.json host, so the file wins over a value the caller configured explicitly. Would it work to apply branch.json before the builder config, so that explicitly configured values win?

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
wpinho-branch: The line above this one is `logAlways(GOOGLE_VERSION_TAG)`, which the NONE default gates off as well, so the `!SDK-VERSION-STRING!` line stops reaching logcat. If that marker is what the Play SDK version reporting picks up, that reporting would go quiet for anyone who does not opt into logging. Worth exempting it while you are adding `isLoggable`?

Branch-SDK/src/main/java/io/branch/referral/BranchUtil.java
```diff
@@ -77,50 +64,16 @@ static boolean checkTestMode(Context context) {
...
-        // 3. Try string resources (lowest priority)
-        BranchLogger.v("Attempting to read branch key from string resources...");
-        branchKey = readBranchKeyFromStringResources(context);
-        if (branchKey != null) {
-            BranchLogger.v("Successfully read branch key from string resources");
         } else {
-            BranchLogger.w("Branch key not found in any source (branch.json, manifest, or string resources)");
+            BranchLogger.v("Branch key not present in branch.json");
         }
         return branchKey;
```
wpinho-branch, line 81: After this change the only references to `readBranchKey` left in the SDK are its own definition and the private helper it calls, so the builder becomes the only key source.

On device the drop is silent rather than loud. With both a `branchKey` and an `apiUrl` set in the TestBed's branch.json, requests went to the branch.json `apiUrl`, so the file is still being read, while the key on the wire stayed the builder's and nothing logged that the branch.json key had been dropped.

Worth deciding whether branch.json stays a supported key source, and either falling back to it here or removing the reader so the removal is explicit?

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
```diff
@@ -0,0 +1,342 @@
...
+        // Network — validated in Builder.build()
+        prefHelper.timeout = networkTimeout
+        prefHelper.connectTimeout = networkConnectTimeout
+        prefHelper.retryCount = retryCount
+        prefHelper.retryInterval = retryInterval
+        prefHelper.noConnectionRetryMax = noConnectionRetryMax
+        remoteInterface?.let { branch.setBranchRemoteInterface(it) }
+
+        // Privacy & attribution
+        attributionLevel?.let { branch.setConsumerProtectionAttributionLevel(it) }
```
wpinho-branch: The single-argument overload sends an OPEN, so applying an attribution level here emits one from inside `initialize`. On device that turns out to suppress the deep link resolution entirely.

Same TestBed, same launch (`branchtest://open?link_click_id=...`), the only difference being `.setAttributionLevel(FULL)` on the builder:

| build | requests at cold start |
| --- | --- |
| without the level | `v3/deeplink` then `v3/events/open` |
| with the level | `v3/events/open` only |

Five runs each, no variation. The two-argument overload sets the level without the OPEN, and with it applied the deeplink request comes back on every run:

```suggestion
        attributionLevel?.let { branch.setConsumerProtectionAttributionLevel(it, null) }
```

wpinho-branch (changes requested):
Re-reviewed against the current head. The builder is in good shape and initialize is far easier to follow than the old spread of setters. What I would like to settle is all downstream of the builder handing off.

Removing `setDMAParamsForEEA` leaves no way to update DMA consent after init. Consent is revocable by design, so an app cannot react when a user withdraws it mid-session. The follow-up PR adds `setDMAParameters`, so this may just be sequencing, but on its own this one ships the gap.

Two smaller things I am not blocking on. `BranchKeysValidatorCheck` reads the key from `PrefHelper` after `build()` has already required a non-blank one, so it can no longer fail. And `setRemoteInterface` does not affect link generation, since the generators capture the interface at construction.

Branch-SDK/src/main/java/io/branch/referral/BranchUtil.java
gdeluna-branch: branch.json is currently out of scope. We'll address when we design plugin support.

Branch-SDK/src/test/java/io/branch/referral/BranchInitializeTest.kt
gdeluna-branch: Fixed.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
gdeluna-branch: Fixed.

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
gdeluna-branch: Did some research, there is no dependency on this string.

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
gdeluna-branch: Fixed, though I think we'll need to revise this behavior in another PR.

gdeluna-branch:
@wpinho-branch Yes the runtime changing of DMAParameters is in the follow up pr #1397

bboothe-branch (approved):

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
```diff
@@ -0,0 +1,342 @@
...
+        // Network — validated in Builder.build()
+        prefHelper.timeout = networkTimeout
+        prefHelper.connectTimeout = networkConnectTimeout
+        prefHelper.retryCount = retryCount
+        prefHelper.retryInterval = retryInterval
+        prefHelper.noConnectionRetryMax = noConnectionRetryMax
+        remoteInterface?.let { branch.setBranchRemoteInterface(it) }
+
+        // Privacy & attribution
+        attributionLevel?.let { branch.setConsumerProtectionAttributionLevel(it, null) }
```
wpinho-branch, line 81: Nothing pins this. The two tests that touch `setAttributionLevel` only assert the value is stored on the config, so a revert to the one-arg overload would still pass.

wpinho-branch (approved):
Re-reviewed the delta. The attribution level fix is the right one: the two-arg overload is the one without the sendOpen, so the open no longer comes out of applyTo. Unit suite is green at this head, 529 tests. One comment below.

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
gdeluna-branch: Added tests

PR #1390, opened 2026-08-21
EMT-4077: run the L1 wire gate on 6.0.0-beta.0 (1 of 2)

wpinho-branch:
## The problem

The L1 wire gate never ran against `6.0.0-beta.0`. The branch was absent from the base filter and there was no `push` trigger at all, so the ticket's acceptance could not be met and the EMT-4079 badges have no branch status to read.

The validator also could not judge a beta capture: the path scope was `uri.startswith("/v1/")`, which skipped `/v3/deeplink` and `/v3/events/open` silently, and a global rule required `/v1/install`, which a correct beta capture never contains.

## What changed

The beta is on both triggers, each with a paths filter so a README change no longer spins up an emulator. The lists are duplicated rather than shared via a YAML anchor: Actions does not guarantee anchor support, and a silently dead trigger is the worst failure mode for a gate.

Path scope is now membership in `REQUIRED_PER_ENDPOINT`, so contracting an endpoint is a data change. The global mandatory-`/v1/install` rule is gone, matching what EMT-4048 did on iOS.

## One deviation from the ticket, deliberate

The ticket asks to drop `/v1/install` and `/v1/open` from the required-field lists. They stay, because this validator also gates master PRs and master still constructs `ServerRequestRegisterInstall`. Removing an entry is not "stop requiring it to appear", it is "stop looking inside it when it does", so keeping them costs the beta nothing. Two-line revert if you would rather follow the ticket literally.

## Evidence

```
5 self-tests           OK
real beta capture      VALIDATION PASSED (3/3 requests valid)
```

That green is honest but empty: the three requests are `/v3` and have no contract yet. The field lists follow in #1391.

The test that pinned the global rule is inverted rather than deleted, with a docstring recording why. Paths filters only prove themselves on a real push that matches and one that does not.

Ticket: https://branch.atlassian.net/browse/EMT-4077


gdeluna-branch (approved):
You need to rebase since this contains your other PR's changes.

PR #1396, opened 2026-08-27
Move v2/event to v3/event

gdeluna-branch:
## Reference
SDK-XXX -- <TITLE>.

- Correctness & Style


Branch-SDK-GPTDriver/mobileboost/suites/log-event-standard.json
```diff
@@ -29,7 +29,7 @@
     {
       "id": 5,
       "type": "ai",
-      "instruction": "Check if you see \"v2/event/standard\" and you see \"Status Code: 200\" right beneath it. Ignore all other \"Status Code: 200\" where preceded with \"v1/open\" or \"v1/install\""
+      "instruction": "Check if you see \"v3/events/standard\" and you see \"Status Code: 200\" right beneath it. Ignore all other \"Status Code: 200\" where preceded with \"v3/events/open\" or \"v1/install\""
```
wpinho-branch, line 32: These steps tap `Send Standard Event` and `See Requests`. Neither label exists in the TestBed, whose log button reads `Load Branch Logs`; both belong to the Link Simulator. So against the APK that this workflow builds, the run stops before it reaches this assertion. Pointing the suite at the Simulator, or adding the matching TestBed buttons, would actually exercise the endpoint change.

Branch-SDK/src/main/java/io/branch/referral/Defines.java
```diff
@@ -267,8 +267,8 @@ public enum RequestPath {
         RegisterInstall("v1/install"),
         RegisterOpen("v1/open"),
         ContentEvent("v1/content-events"),
-        TrackStandardEvent("v2/event/standard"),
-        TrackCustomEvent("v2/event/custom"),
+        TrackStandardEvent("v3/events/standard"),
+        TrackCustomEvent("v3/events/custom"),
```
wpinho-branch, line 271: Nothing pins these literals. The only assertion that touches them compares the enum against itself, so a singular `v3/event/...` would still go green:

https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/f754fa88b55107cb7389a46164b2721dc28f6fe1/Branch-SDK/src/androidTest/java/io/branch/referral/BranchEventTest.java#L84

A small assertion on the built URL would catch that.

wpinho-branch (commented):
The SDK side looks right. Version selection comes from the request override rather than the path, so the `user_data` shape is unaffected, and the plural form matches the `v3/events/open` route already in the enum.

The suite JSONs do not appear to be uploaded from this repo, so unless the live ones are updated on the MobileBoost side too, they would still be asserting the old paths after this merges. Worth a line in the description if that is already handled.

iOS moved open to v3 but still posts standard and custom events to the v2 paths on both its lines, so this would make Android the only side on v3 for events. Is there already a server-side decision behind that? I could not find one recorded in either repo.

Branch-SDK-GPTDriver/mobileboost/suites/log-event-standard.json
bboothe-branch: Verified this occurs. log-event-custom.json hits the same "See Requests" gap you found. set-session-id.json and set-customer-event-alias.json go further tap "Change App's Session ID" / "Set Customer Event Alias" and "Save", none of these exist in the TestBed at all. Even with working UI their final assertions (bls_session_id, customer_event_alias: aliasExample) could never pass — the SDK never emits bls_session_id, and the alias button hardcodes a different value.

Branch-SDK-TestBed/src/main/java/io/branch/branchandroidtestbed/MainActivity.java
```diff
@@ -640,6 +640,33 @@ public void onFailure(Exception e) {
...
+                            public void onSuccess(int responseCode) {
+                                BranchLogger.d("MODERNIZATION_DEBUG: Custom Event sent successfully: " + responseCode);
+                                showLongToast("Sent Branch Custom Event: " + responseCode);
+                            }
+
+                            @Override
+                            public void onFailure(Exception e) {
+                                BranchLogger.d("MODERNIZATION_DEBUG: Custom Event failed: " + e.getMessage());
+                                BranchLogger.d("BranchSDK_Tester: " + e.toString());
+                                Toast.makeText(getApplicationContext(), "Error sending Branch Custom Event: " + e, Toast.LENGTH_LONG).show();
```
bboothe-branch, line 664: The new custom event listener duplicates an existing error bug.  This bug exists already for Commerce/Content/Lifecycle listeners. To fix, we can reuse the logic in TestBedHelper.java to fix by wrapping the whole call in the main thread dispatch. This can be a separate pull request.

Example: 

```
new Handler(Looper.getMainLooper()).post(() ->
    TestBedHelper.showLongToast(getApplicationContext(), "Error sending Branch Custom Event: " + e));
```

bboothe-branch (changes requested):
PR looks good. A couple comments before approving.

ServerRequestLogEvent.java:87 still returns Branch_API_VERSION.v2. This needs to be updated to V3 for the 4 respective call sites below:

ServerRequestLogEvent.java:87:
```
    @Override
    public BRANCH_API_VERSION getBranchRemoteAPIVersion() {
        return BRANCH_API_VERSION.V2; //This is a v2 event
    }
```


ServerRequest.java:570
`if (version == BRANCH_API_VERSION.V2)`

ServerRequest.java:822:
`if (getBranchRemoteAPIVersion() == BRANCH_API_VERSION.V2)`


ServerRequest.java:62 (no V3 enum at all)

```
 public enum BRANCH_API_VERSION {
        V1,
        V1_LATD,
        V2
    }
```

Branch-SDK/src/main/java/io/branch/referral/Defines.java
wpinho-branch: Closed. `EventRequestPathTest` pins the exact literal that is concatenated into the request URL, so the singular form now fails.

Branch-SDK-GPTDriver/mobileboost/suites/log-event-standard.json
wpinho-branch: Correcting my earlier comment: the `Send Custom Event` button exists now, so log-event-custom clears its first tap. All four suites still stop before the assertion, on `See Requests` or on the buttons that set session ID and customer event alias.

Branch-SDK/src/test/java/io/branch/referral/EventRequestPathTest.kt
```diff
@@ -0,0 +1,103 @@
...
+            JSONObject(),
+            emptyList()
+        )
+
+    // --- Paths: pinned against literals, so a singular "v3/event/..." typo fails. -------------
+    // BranchEventTest only compares the enum to itself, which such a typo would survive.
+
+    @Test
+    fun standardEventPostsToV3() {
+        assertEquals("v3/events/standard", Defines.RequestPath.TrackStandardEvent.path)
```
wpinho-branch, line 49: The path selection itself stays unpinned: swapping the two branches at `util/BranchEvent.java:267` keeps all six new tests green.

wpinho-branch (commented):
The new test closes the path-literal thread: the pinned field is the exact string concatenated into the URL, and I confirmed the six tests pass at this head. Leaving the GPTDriver thread open, with a correction to my own earlier comment in it.

wpinho-branch (approved):
Upgrading my earlier COMMENTED to an approve. The change is correct and now pinned: the six new tests pass, the full suite is green at 444, and the literals match iOS exactly. Neither thread I left open is a defect in this change. The GPTDriver one is pre-existing infra debt and those suites do not run on PRs anyway, and the path selection pin is an additional ask, so treat both as non-blocking notes.

bboothe-branch (approved):

PR #1397, opened 2026-08-28
EMT-4057 — [Android] Replace setDMAParamsForEEA with a typed DMAParameters object

gdeluna-branch:
## Reference

EMT-4057 — [Android] Replace `setDMAParamsForEEA` with a typed `DMAParameters` object

Builds on the `BranchConfiguration` builder PR. Base: `gdeluna-branch/EMT-3912-2`.

### Changes

**`DMAParameters`** — now a class with a nested `Builder` and a private constructor. A typed object alone didn't fix the original problem: `new DMAParameters(true, false, true)` was as transposable as `setDMAParamsForEEA(true, false, true)`. The builder makes a Java call site name every field.

```java
DMAParameters dma = new DMAParameters.Builder()
        .setEeaRegion(true)
        .setAdPersonalizationConsent(false)
        .setAdUserDataUsageConsent(true)
        .build();
```

Also adds `toBuilder()` for changing one field without restating the others.

**`Branch.setDMAParameters(DMAParameters)`** — new instance method. Consent is mutable at runtime; removing `setDMAParamsForEEA` had left no post-init path.

**Inconsistency warning** — consent granted while `eeaRegion` is `false` produces a warning. It's determined in `build()` but carried on the instance and emitted by whichever consumer takes the object (`applyTo()` at init, `setDMAParameters()` at runtime). Logging isn't configured yet at `build()` time, so a warning logged there would be dropped. It does not throw — with `toBuilder()` and the runtime setter, a partially-coherent instance is a legitimate intermediate.

**`BranchConfiguration.Builder.build()`** — reports every validation failure in one exception instead of throwing on the first, so a caller with three bad values fixes them in one pass.

```
Invalid BranchConfiguration:
  - Branch key cannot be empty. Get your key from dashboard.branch.io/settings.
  - Network timeout must be a positive number of milliseconds (got -1).
  - Retry interval must be a positive number of milliseconds (got 0).
```

### Verification

Unit: 546 tests, 0 failures, two clean runs. `DMAParameters(true, true, true)` confirmed non-compiling via a throwaway probe.

On device (Pixel_9a, Android 16): TestBed sets `eeaRegion=true, adPersonalizationConsent=false, adUserDataUsageConsent=true`; all three launch requests carried `dma_eea:true, dma_ad_personalization:false, dma_ad_user_data:true` and returned `200`. Capture shared separately.

### Breaking

`DMAParameters`' positional constructor is now private — affects anyone who picked it up from the beta branch.

- Correctness & Style


Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -2377,26 +2106,19 @@ public void logEventWithPurchase(@NonNull Context context, @NonNull Purchase pur
...
-            BranchLogger.v("setFBAppID to " + fbAppID);
-        } else {
-            BranchLogger.w("setFBAppID: fbAppID cannot be empty or null");
-        }
+    public void setDMAParameters(@NonNull DMAParameters params) {
+        params.logWarnings();
+        prefHelper_.setEEARegion(params.getEeaRegion());
+        prefHelper_.setAdPersonalizationConsent(params.getAdPersonalizationConsent());
+        prefHelper_.setAdUserDataUsageConsent(params.getAdUserDataUsageConsent());
     }
```
wpinho-branch, line 2121: The three consent values land as separate preference writes, since each setter commits on its own:

https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/0b85568e507301067e20996cf5811272128c8435/Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java#L1348-L1350

The request path reads them independently just before send, so an update delivered off the main thread can be observed mid-write. Batching them into one editor commit would carry the typed object's atomicity through to the wire, and a test asserting the three keys reach the request body would cover the part that only the device run checks today.

Branch-SDK/src/main/java/io/branch/referral/DMAParameters.kt
```diff
@@ -0,0 +1,88 @@
...
+    val adUserDataUsageConsent: Boolean,
+    private val warnings: List<String>
+) {
+
+    /**
+     * Emitted by whichever consumer takes this object, rather than from [Builder.build], because at
+     * build time the logger is still disabled and the lines would be lost.
+     */
+    @JvmName("logWarnings")
+    internal fun logWarnings() {
```
wpinho-branch, line 33: This compiles to a public, unmangled method on a public class, so consumer code can reach it. `toJson` in `BranchConfiguration` is `internal` too but carries no `@JvmName`, and it comes out with the module suffix, which looks like the difference here. Is the explicit JVM name needed for the call from `Branch`, or could it come off, keeping this off the published surface?

Branch-SDK-TestBed/src/main/java/io/branch/branchandroidtestbed/CustomBranchApp.java
```diff
@@ -14,25 +15,59 @@
...
+                        .setLoggingCallback((message, severity) -> {
+                            Log.d("BranchTestbed", message);
+                            saveLogToFile(message);
+                        })
+                        .setRequestTracingCallback(tracingCallback())
+                        .setDMAParameters(new DMAParameters.Builder()
+                                .setEeaRegion(true)
+                                .setAdPersonalizationConsent(false)
+                                .setAdUserDataUsageConsent(true)
+                                .build());
```
wpinho-branch: This leaves the sample app declaring EEA with fixed consent on every launch, which could mislead someone reading a TestBed capture later. Routing it through the existing settings override would keep it opt-in.

wpinho-branch (changes requested):
The builder shape reads well, and the runtime setter fills the gap the previous PR left, so the two together make sense to me.

The base looks off though. This branch appears to be built on the configuration builder branch rather than the one it targets, which is why the diff here replays the whole builder change. As it stands, merging would carry that work into the v3 events branch and from there onto the beta, landing it without its own review. Retargeting would shrink this to the DMA change and keep the two separable. Am I reading the intent right?

Branch-SDK/src/main/java/io/branch/referral/Branch.java
gdeluna-branch: Good suggestion. Done.

Branch-SDK-TestBed/src/main/java/io/branch/branchandroidtestbed/CustomBranchApp.java
gdeluna-branch: Fixed, added a toggle in Settings ("Enable DMA Params"), off by default. The app only sets DMA values now if you turn it on, so a normal run doesn't declare EEA consent anymore.

Branch-SDK/src/main/java/io/branch/referral/DMAParameters.kt
gdeluna-branch: Tested this directly. Removing @JvmName breaks the build. Branch.java calls params.logWarnings(), and Kotlin renames internal methods at the JVM level unless @JvmName is set. Without it, Java code in the same module can't find the method. Same reason applyTo() already has it.

You're right that this makes the method reachable from outside the SDK, same as applyTo() already is. Looked at a few ways around it and none hold up while Branch.java needs to call it directly. Leaving this for the pre-GA sealing pass rather than solving it here.

bboothe-branch (approved):
Approved.

Branch-SDK/src/main/java/io/branch/referral/PrefHelper.java
```diff
@@ -1037,6 +1037,18 @@ boolean getEEARegion() {
...
+    /**
+     * Writes all three DMA consent values in a single editor commit, so a concurrent read (the
+     * request builder reads them independently just before send) cannot observe a torn combination
+     * of old and new values.
+     */
+    void setDMAParameters(boolean isEEARegion, boolean hasAdPersonalizationConsent, boolean hasAdUserDataUsageConsent) {
+        prefsEditor_.putBoolean(KEY_DMA_EEA, isEEARegion)
+                .putBoolean(KEY_DMA_AD_PERSONALIZATION, hasAdPersonalizationConsent)
+                .putBoolean(KEY_DMA_AD_USER_DATA, hasAdUserDataUsageConsent)
+                .apply();
```
wpinho-branch: The shared `prefsEditor_` means another caller's `apply()` can commit this batch half written. Since `isDMAParamsInitialized()` only checks the EEA key, that partial state sends `dma_eea=true` with both consents false, which declares the wrong consent rather than none.

```suggestion
        appSharedPrefs_.edit()
                .putBoolean(KEY_DMA_EEA, isEEARegion)
                .putBoolean(KEY_DMA_AD_PERSONALIZATION, hasAdPersonalizationConsent)
                .putBoolean(KEY_DMA_AD_USER_DATA, hasAdUserDataUsageConsent)
                .apply();
```

Branch-SDK/src/main/java/io/branch/referral/BranchConfiguration.kt
```diff
@@ -80,9 +80,8 @@ class BranchConfiguration private constructor(
         // Privacy & attribution
         attributionLevel?.let { branch.setConsumerProtectionAttributionLevel(it) }
```
wpinho-branch: This branch is behind its base, and line 81 still has the one-arg overload while the base now has `(it, null)`. Resolving toward this branch on rebase would revert that fix.

wpinho-branch (changes requested):
Two things before this goes in: the batched setter does not deliver the atomicity its javadoc claims, and the pending rebase can revert the attribution fix on the base. Unit suite is green at this head, 547 tests, and the TestBed toggle and the @JvmName question both check out.

PR #1405, opened 2026-09-08
EMT-4083: contract the warm launch scenarios, warm_https_onNewIntent and warm_uriScheme

wpinho-branch:
Two warm scenarios, both entered while the app is alive and backgrounded: `warm_https_onNewIntent` arrives as an https link, `warm_uriScheme` as a `branchtest://` one. Drivers, captures, contracts, and both running in the L1 gate.

## What the captures say

The contracts are written from the measurement, not from the ticket. The ticket predicted one `/v3/deeplink` and exactly one `/v3/events/open`; both warm scenarios measure two and three, which is two opens more than `cold_https`. The only thing a warm launch does that a cold one does not is background and foreground the app, and that is a coincidence worth stating rather than a mapping this contract proves. What the ticket asked for and the capture confirms is the absence of an install, asserted at zero.

The two scenarios share counts and order deliberately. What `warm_uriScheme` adds is not a different wire shape but proof that the manifest's `branchtest` filter matches and that the OS hands a scheme intent to a backgrounded app, neither of which is reachable from a JVM test. The entry point is what separates them, asserted in both directions through `external_intent_uri`; without it the second capture says nothing the first does not. The mapping itself is tested on the JVM in `RequestDeepLinkUriMappingTest`, not here.

## A defect the merge closes

This branch produced `wire-w1.txt` and `wire-w2.txt` that no consumer read. Neither appeared in the artifact path list, the summary grep or the failure dump, so both warm scenarios were missing from the public endpoint table and from the failure output. #1401 globs all three consumers to `wire-*.txt`, which picks the renamed files up without naming them.

## The rename

`W1` becomes `warm_https_onNewIntent` and `W2` becomes `warm_uriScheme`, with their fixtures, contract comments, test references, driver KDoc, check pairs and workflow capture filenames. Both fixtures move as pure renames, zero content lines changed. Neither warm capture carries a stamped `l1_scenario`, because the warm drivers are `TEST_CLASS` driven rather than `COLD_SCENARIO` driven, so the marker leg that fails only on a live emulator does not reach this branch.

The two warm workflow lines got shorter, which is the opposite of what a rename diff usually does. One fell from 136 to 130 characters and the other from 140 to 126, where the rename alone would have taken it to 152. Dropping the `io.branch.gptdriver.tests.` prefix pays for the longer name twice over, and each bare name resolves to the same fully qualified class the line passed before.

Suite 49 to 50 tests, the addition being the base's marker guard, re-verified after the merge commit rather than before.


wpinho-branch:
#1417 adds an N3 fixture carrying `cpp_level`; whichever of the two merges second adds it to the keep set.

PR #1406, opened 2026-09-09
EMT-4058: add Kotlin coroutine suspend variants

gdeluna-branch:
# Add Kotlin coroutine suspend variants

## What this does

Adds Kotlin coroutine versions of three Branch APIs. You can now await a result
instead of passing a callback.

Nothing existing changes. Every callback method still behaves the same way.

## New functions

```kotlin
suspend fun Branch.requestDeepLinkData(uri: Uri): JSONObject
suspend fun BranchUniversalObject.createLink(context, linkProperties, defaultToLongUrl = true): String
suspend fun BranchShortLinkBuilder.createLink(): String
suspend fun BranchEvent.awaitLogEvent(context: Context)
```

Each one returns a value or throws `BranchException`, which holds the same
`BranchError` the callback would have handed back.

They are extension functions because `Branch`, `BranchUniversalObject` and
`BranchEvent` are Java classes and cannot have suspend methods.

`awaitLogEvent` is named differently because `BranchEvent` already has a
`logEvent(Context)` method. An extension with the same name would never be called.

## Two supporting changes

`setBranchRemoteInterface` now rebuilds the link generators. Before this, swapping
the network layer had no effect on link creation, because the generators kept the
old one. The method exists so tests can swap in a fake, so this was a bug.

`BranchUrlBuilder.createUrlRequest` is pulled out of two methods that were building
the same object. `createLink` needs it to read the long URL.

## Commits

1. Adds the empty Kotlin sample app module.
2. Adds the suspend functions and wires the sample app to call them.
3. Adds the tests.

## Testing

12 new unit tests. Full suite is 557 tests, all passing. The tests run the real
request queue against a fake network layer rather than mocking the new functions.

Also ran the sample app on an emulator against the live test app. All seven
buttons work:

- created a short link from a BranchUniversalObject
- created a short link from a builder
- got a `BranchException` when creation failed and `defaultToLongUrl` was false
- got a long URL back when it was true
- logged an event
- resolved a link and got back the same channel and feature the builder set
- cancelled a request and confirmed it left the queue

There is no cancellation test for `createLink`. Link creation never enters the
queue, so there is nothing to observe, and both versions I tried still passed
when I broke the code on purpose. Please do not add one back without checking it
can fail.


Branch-SDK/src/main/java/io/branch/referral/BranchCoroutines.kt
```diff
@@ -0,0 +1,56 @@
...
+suspend fun BranchEvent.awaitLogEvent(context: Context): Unit =
+    suspendCancellableCoroutine { continuation ->
+        logEvent(context, object : BranchEvent.BranchLogEventCallback {
+            override fun onSuccess(responseCode: Int) {
+                continuation.resume(Unit)
+            }
+
+            override fun onFailure(e: Exception) {
+                continuation.resumeWithException(e)
+            }
```
wpinho-branch, line 27: This one throws a plain `Exception` while the other three throw `BranchException`, so the only way to know why it failed is to parse the message. Since this becomes public API, could `BranchEvent.handleFailure` pass a `BranchException` to `onFailure` instead? It's still an `Exception`, so existing callbacks keep working, though the message text would change.

Branch-SDK/src/test/java/io/branch/referral/BranchDeepLinkCoroutinesTest.kt
```diff
@@ -0,0 +1,140 @@
...
+        )
+
+        try {
+            Branch.getInstance().requestDeepLinkData(uri)
+            fail("expected BranchException when the deep link cannot be resolved")
+        } catch (e: BranchException) {
+            assertTrue(
+                "expected the server failure to be carried through, got: ${e.message}",
+                e.branchError.errorCode != 0
+            )
```
wpinho-branch, line 96: `errorCode` is never 0, so this passes no matter which error comes back. A 500 arrives as 500 here, so it can be pinned:

```suggestion
            assertEquals(
                "the BranchError the callback path would have delivered must survive intact",
                500,
                e.branchError.errorCode
            )
```

wpinho-branch (approved):
Looks good overall. It needs a rebase: the base branch already has the same `Branch.initialize` fix for `EventRequestPathTest` and `QueueContainsInstallOrOpenTest`, which is where the conflict comes from, so the "Also fixed" part of the description can go. The unit test workflow runs on PRs now, so the rebase will also give this one a CI run.

Two small things on the testbed: nothing in CI builds it, so it can drift from the SDK without anyone noticing, and `ExampleUnitTest` and `ExampleInstrumentedTest` are leftovers from the Android Studio template.

Unrelated to this change but close to it: on this branch the callback `generateShortUrl` with `defaultToLongUrl = true` returns null on a server failure instead of the long URL, because the async generator never gets that flag. `createLink` handles it on its own, so the two now behave differently. Probably worth a separate ticket.

Branch-SDK/src/main/java/io/branch/referral/LinkCoroutines.kt
```diff
@@ -0,0 +1,44 @@
...
+                )
+            )
+
+        suspendCancellableCoroutine { continuation ->
+            lateinit var request: ServerRequestCreateUrl
+
+            val listener = Branch.BranchLinkCreateListener { url, error ->
+                when {
+                    url != null -> continuation.resume(url)
+                    request.isDefaultToLongUrl -> continuation.resume(request.longUrl)
```
bboothe-branch: `request.longURL` is called from inside the `BranchLinkCreateListener`. On any exception, `generateLongUrlWithParams` catch block calls `callback_.onLinkCreate(null, error)`, thus calling this listener again, and `request.longUrl again`. It effectively continues to call this forever. Fixable by moving `request.longURL` outside of the listener as it currently only recurses when `generateLongUrlWithParams` throws.

Branch-SDK/src/main/java/io/branch/referral/LinkCoroutines.kt
```diff
@@ -0,0 +1,44 @@
...
+    withContext(Dispatchers.IO) {
+        val branch = Branch.getInstance()
+            ?: throw BranchException(
+                BranchError(
+                    "Trouble creating a URL.",
+                    BranchError.ERR_BRANCH_NOT_INSTANTIATED
+                )
+            )
+
+        suspendCancellableCoroutine { continuation ->
```
bboothe-branch, line 26: Worth adding a timeout wrapper to createLink()'s suspendCancellableCoroutine(). It covers every callback so that setBranchRemoteInterface's rebuilding doesn't prevent drop of the callback.

Branch-SDK/src/main/java/io/branch/referral/Branch.java
```diff
@@ -462,6 +462,14 @@ void setBranchRemoteInterface(BranchRemoteInterface remoteInterface) {
...
         }
+
+        // Generators capture the interface at construction; rebuild or link creation keeps
+        // posting through the replaced one.
+        if (modernLinkGenerator_ != null) {
+            modernLinkGenerator_.shutdown();
+        }
+        modernLinkGenerator_ = new ModernLinkGenerator(context_, branchRemoteInterface_, prefHelper_);
+        legacyLinkGenerator_ = new BranchLegacyLinkGenerator(prefHelper_, branchRemoteInterface_);
     }
```
bboothe-branch, line 476: shutdown() doesn't  truly clear the linkcache because it only clears the generator's cache, not the linkCache itself. Add linkCache_.clear() here in setBranchRemoteInterface() next to the rebuild.

bboothe-branch (changes requested):
Overall good, but some things are needed before merging.

Comments added below.

Make the resume one-shot: one AtomicBoolean per continuation, CAS checked at every resume arm.

A second callback throws IllegalStateException: Already resumed on the queue's thread, and the consumer's catch sits outside its for (trigger in processingTrigger) loop, so that exception ends the consumer coroutine and the queue stops processing for the rest of the process. All arms need it because the second callback can land on a different branch than the first.

Branch-SDK/src/main/java/io/branch/referral/LinkCoroutines.kt
wpinho-branch: Confirmed: when `getLongUrl` throws (tracking disabled plus a user URL with no query is enough), the catch in [`generateLongUrlWithParams`](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/b79df4a6c3885906e7408d4c4d5e1e978cf1a346/Branch-SDK/src/main/java/io/branch/referral/ServerRequestCreateUrl.java#L255-L257) calls `callback_`, which is this listener, so it re-enters until a `StackOverflowError` escapes and `createLink` never resumes. Reading the long URL from a request with its own no-op listener avoids it:

```suggestion
                    request.isDefaultToLongUrl -> continuation.resume(createUrlRequest(Branch.BranchLinkCreateListener { _, _ -> }, true).longUrl)
```

Branch-SDK/src/main/java/io/branch/referral/LinkCoroutines.kt
wpinho-branch: The network call is already bounded by [`withTimeout`](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/b79df4a6c3885906e7408d4c4d5e1e978cf1a346/Branch-SDK/src/main/java/io/branch/referral/ModernLinkGenerator.kt#L103); what drops the callback is `shutdown()` cancelling mid-flight, because the `withContext(Dispatchers.Main)` in [the generator's catch](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/b79df4a6c3885906e7408d4c4d5e1e978cf1a346/Branch-SDK/src/main/java/io/branch/referral/ModernLinkGenerator.kt#L285-L294) throws on the cancelled job. Making that `withContext(NonCancellable + Dispatchers.Main)` delivers the cancellation as an error instead, so I'd fix it there rather than add a second timeout here.

Branch-SDK/src/main/java/io/branch/referral/ModernLinkGenerator.kt
bboothe-branch: `NonCancellable` at **ModernLinkGenerator.kt:288** closes the case where `shutdown()` cancels the scope mid-flight, but it doesn't close the case where the scope is already cancelled when `scope.launch` is called. Can fix by adding `scope.launch(start = CoroutineStart.ATOMIC)`

bboothe-branch: Referenced only by ModernLinkGenerator test. Can be removed or commented out as production goes through lower scope.launch via Branch.java:1006.

bboothe-branch:
```suggestion
        scope.launch(start = CoroutineStart.ATOMIC) {
```

`NonCancellable` at **ModernLinkGenerator.kt:288** closes the case where `shutdown()` cancels the scope mid-flight, but it doesn't close the case where the scope is already cancelled when `scope.launch` is called. Can fix by adding `scope.launch(start = CoroutineStart.ATOMIC)`

bboothe-branch (commented):
Fixes look great. One additional comment I wanted to note.

Non Blocking: BUO.createLink and BUO.generateShortURL(context,lp,callback) behave differently when tracking is disabled. Create. The callback passes null as the error regardless of outcome. The suspend version takes the async path and surfases failures as BranchException. I'd say we go with the suspend version.

bboothe-branch (approved):

PR #1409, opened 2026-09-11
EMT-4288: read the remote interface at call time in both link generators

wpinho-branch:
`setBranchRemoteInterface` is documented as the seam that substitutes the HTTP layer for tests. It did not reach link generation.

Both link generators captured the `BranchRemoteInterface` instance at construction, and construction happens inside Branch's own setup, before an integrator can substitute anything. So a mock installed through `BranchConfiguration.Builder.setRemoteInterface` was never consulted by short link creation, and the documented seam silently did not work for the one thing people use it for.

## What changed

Both generators now take a `() -> BranchRemoteInterface` provider and read it at the call site instead of holding an instance. The two construction sites in `Branch.java` pass live-reading lambdas.

`ModernLinkGenerator`'s primary constructor is `@JvmOverloads` so the Java call site can pass a three-argument lambda without the scope and timeout defaults, and the existing instance-typed constructor is retained so callers substituting a fixed instance keep compiling.

## The failing test came first

Both methods failed against the unmodified code, deterministically across four attempts:

```
Wanted but not invoked:
remoteInterface.make_restful_post(<any>, <any>, "v1/url", <any>);

Actually, there were zero interactions with this mock.
```

Zero interactions is the strong form. The interface was not receiving a malformed call, it was not being consulted at all, which is exactly the defect the ticket describes.

## One thing in the test worth knowing

The new test calls `Branch.shutDown()` before initializing. `Branch.initialize` returns before applying its configuration when the singleton already exists (`Branch.java:378-382`), so a test class that leaks the singleton makes a later `initialize` a silent no-op and the mock is never wired.

One class in the JVM suite does leak, which made this test flake at full-suite scale while passing in isolation. That is EMT-4339 and is not fixed here. The guard makes this test independent of what ran before it rather than dependent on every other class's teardown.

## Proof

`./gradlew :Branch-SDK:testDebugUnitTest`: 545 tests, 0 failures, 0 errors, across three clean runs with no retry fired.

Scope is `Branch-SDK/src/main` and its unit tests. The instrumented harness is EMT-4206 and is untouched here.


wpinho-branch:
Moved to draft: #1406 (merged) already rebuilds the link generators in setBranchRemoteInterface, and this PR's RemoteInterfaceSubstitutionTest passes on 6.0.0-beta.0 at 05d9f706 without this change. The current diff is not mergeable as is.

wpinho-branch:
Closing: #1406 (merged) already rebuilds the link generators in setBranchRemoteInterface, and this PR's RemoteInterfaceSubstitutionTest passes on 6.0.0-beta.0 at 05d9f706 without this change.

PR #1415, opened 2026-09-16
EMT-4373: run the unit tests and the L1 wire gate for 6.0.0-beta.2

wpinho-branch:
Active beta work now targets `6.0.0-beta.2`, but `unit-tests.yml` and `sdk-l1-validation.yml` only list `6.0.0-beta.0`, so pull requests to the new line, and merges into it, run neither `verify` nor `validate-wire-logs`.

This adds `6.0.0-beta.2` to the `push` and `pull_request` branch lists of both workflows. Every existing entry stays, and no job, path filter or step changes.

Config only, so there is no device measurement. The evidence is this pull request itself: it targets `6.0.0-beta.2`, and `pull_request` runs are read from the head branch, so `verify` and `validate-wire-logs` appearing below is the proof that the triggers work. Both files parse, and the parsed branch lists were checked before the push.


.github/workflows/sdk-l1-validation.yml
```diff
@@ -5,7 +5,7 @@ on:
   # read a branch status, which only a push event produces. A pull_request-only trigger leaves
   # both unmet.
   push:
-    branches: ["6.0.0-beta.0"]
+    branches: ["6.0.0-beta.0", "6.0.0-beta.2"]
```
gdeluna-branch: Please wildcard the revision number since we'll possibly be creating several of these.

gdeluna-branch (approved):

.github/workflows/sdk-l1-validation.yml
wpinho-branch: Wildcarded in 2b933621: both workflows now trigger on `6.0.0-beta.*`.

